PR #98, opened 2014-07-07
Created strongly typed configuration class

kant2002:
Created strongly typed configuration for the following purposes:
- Remove configuration parsing logic and defaults handling from the
CDVViewController.m
- Improved testability of the Cordova configuration and other parts of
code which related on configuration could be much clearly mocked.
- Reduction for code duplication during configuration parsing.
- Ability to verify configuration issues during parsing, For example
passing not a number to the value which expected to be a number should
result to warning, so application developer could easily find error on
their side.
- Single place for storing configuration defaults.

Drawbacks in current implementation
- Not sure about memory leaks during such approach. Not very familiar
with iOS. This is need review
- No tests so far, could be added later if whole idea get traction.
- Currently no support for settings with limited set of values allowed.
Like «PaginationMode». Again could be added if whole idea would be
supported.

### DIFF
--- a/CordovaLib/Classes/CDVConfiguration.h
+++ b/CordovaLib/Classes/CDVConfiguration.h
@@ -1,0 +1,37 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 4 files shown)


shazron:
Thanks but due to the pluggable webview feature, each WebView plugin has its own configuration now (4.0.x branch). This PR will be auto-closed.

PR #104, opened 2014-09-16
generates a workspace instead of subproject

edewit:
Initial part of [CB-5921](https://issues.apache.org/jira/browse/CB-5921)


edewit:
now that swift options are added is this still relevant?

yjukaku:
@edewit Can you please clarify what you mean? To be able to use CocoaPods, we still need a workspace to be generated instead of a project with a sub-project.

edewit:
The initial idea was to be able to support swift plugins. This is now fixed by changing build.xconfig. For cocoapods this pr could still be relevant

purplecabbage:
cordova-ios has supported workspaces+cocoapods for awhile now.  Somehow this was never closed.  Thanks for contributing.

PR #119, opened 2014-11-21
proof of concept for proposal discussed in CB-8032

tony--:
This PR is meant to demonstrate the proposal I made in a comment on CB-8032.
As such, I did not test it - it is only meant for discussion.

This adds a plugin result delegate that can munge the plugin result js before it is returned to the user's application.

The Cordova Local Web Server plugin would set itself as the delegate:
```
self.viewController.commandDelegate.pluginResultDelegate = self;
```

and implement the mungePluginResult method something like the following:
```
-(NSString *)mungePluginResult:(NSString *)result {
	return [result stringByReplacingOccurrencesOfString:@"file://" withString:[NSString stringWithFormat:@"http://localhost:%lu/", self.server.port]];
}
```

### DIFF
--- a/CordovaLib/Classes/CDVCommandDelegate.h
+++ b/CordovaLib/Classes/CDVCommandDelegate.h
@@ -19,6 +19,7 @@
 
 #import "CDVAvailability.h"
 #import "CDVInvokedUrlCommand.h"
+#import "CDVPluginResultDelegate.h"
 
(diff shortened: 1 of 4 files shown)


shazron:
Auto-closing since 4.0.x method has a transformURL method in the CommandDelegate that plugins check

PR #123, opened 2015-01-07
Fixed callbackFromNative

mdudek:
NO_RESULT status was resolved as failure. This bug is there from Oct 14, 2014.


purplecabbage:
You cannot fix this here.  cordova.js is the output of https://github.com/apache/cordova-js

mdudek:
Thanks for info. I will fix it there...

PR #126, opened 2015-01-21
CB-8197 Switch to nodejs for ios platform scripts

vladimir-kotikov:
Implementation for [CB-8197](https://issues.apache.org/jira/browse/CB-8197)
Part of this work is already done for create/update scripts in https://issues.apache.org/jira/browse/CB-5706 but platform scripts placed in cordova folder of platform is still being unported.


bin/lib/versions.js
```diff
@@ -85,17 +85,59 @@ exports.get_apple_xcode_version = function() {
     child_process.exec('xcodebuild -version', function(error, stdout, stderr) {
         if (error) {
             d.reject(stderr);
+        } else {
+            var version = stdout.split('\n')[0].slice(6);
```
agrieve: Would be better to use a regexp here I think:
    var m = /Xcode (.*)/.exec(stdout);
    if (m) {
        d.resolve(m[1]);
    ...

bin/lib/check_reqs.js
```diff
@@ -0,0 +1,94 @@
...
+        return Q.reject('Cordova tooling for iOS requires Apple OS X');
+    }
+    // Check whether tool command is available at all
+    var tool_command = shell.which(tool);
+    if (!tool_command) {
+        return Q.reject(optMessage || (tool + 'command is unavailable.'));
+    }
+    // check if tool version is greater than specified one
+    return versions.get_tool_version(tool).then(function (version) {
+        return minVersion <= parseFloat(version) ?
```
agrieve: I don't think apple restricts to two-number versions (e.g. Xcode 5.1.1 was a thing). Maybe best approach (rather than checking in all of semver module), is to add a helper to versions.js for comparing versions via splitting on '.'

bin/lib/versions.js
```diff
@@ -85,17 +85,59 @@ exports.get_apple_xcode_version = function() {
...
+/**
+ * Gets specific tool version
+ * @param  {String} toolName Tool name to check. Known tools are 'xcodebuild', 'ios-sim' and 'ios-deploy'
+ * @return {Promise}         Promise that either resolved with tool version
+ *                                   or rejected in case of error
+ */
+exports.get_tool_version = function (toolName) {
+    switch (toolName) {
+        case 'xcodebuild': return exports.get_apple_xcode_version();
+        case 'ios-sim': return exports.get_apple_xcode_version();
```
agrieve: these two are pointing to the wrong function.

bin/node_modules/nopt/.npmignore
```diff
@@ -0,0 +1 @@
+node_modules
```
agrieve: Why exclude nopt's dependencies when npm pack'ing, and not just exclude nopt itself?

agrieve: Also - do you not need to check in the "abbrev" module, since nopt depends on it?

agrieve: NVM - I see it now :S

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -0,0 +1,184 @@
...
+    // Valid values for "--target" (case sensitive):
+    var validTargets = ["iPhone-4s", "iPhone-5", "iPhone-5s", "iPhone-6-Plus", "iPhone-6",
+        "iPad-2", "iPad-Retina", "iPad-Air", "Resizable-iPhone", "Resizable-iPad"];
+    if (args.target && validTargets.indexOf(args.target) < 0 ) {
+        return Q.reject(args.target + " is not a valid target for emulator");
+    }
+
+    // check for either ios-sim or ios-deploy is available
+    // depending on arguments provided
+    var checkTools = args.device ? check_reqs.check_ios_deploy() : check_reqs.check_ios_sim();
```
agrieve: This should come after args.list, no?

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -0,0 +1,184 @@
...
+        if (args.emulator) return listEmulators();
+        return listDevices().then(function () {
+            return listEmulators();
+        });
+    }
+
+    return checkTools.then(function () {
+        if (args.nobuild) {
+            // --nobuild option is passed. Skipping build...
+            return Q();
```
agrieve: So you did this in another place as well, but you don't need to return a promise here. It would be the same to just "return", and even shorter to:

    if (!args.nobuild) {
        return build.run(argv);
    }
    // by default, returns undefined, which is the same as returning Q()

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -0,0 +1,184 @@
...
+        }
+        // if --nobuild isn't specified then build app first
+        return build.run(argv);
+    }).then(function () {
+
+        var xcodeProjFiles = shell.ls(projectPath).filter(function (name) {
```
agrieve: Seen this in a few spots now. Probably would be good to make a helper function for it. Maybe stick it in build.js?

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -0,0 +1,184 @@
...
+
+        var projectName = path.basename(xcodeProjFiles[0], '.xcodeproj');
+
+        var appPath = path.join(projectPath, 'build', (args.device ? 'device' : 'emulator'), projectName + '.app');
+        // select command to run and arguments depending whether
+        // we're running on device/emulator
+        if (args.device) {
+            return checkDeviceConnected().then(function () {
+                return deployToDevice(appPath);
+            }).fail(function () {
```
agrieve: This will deploy to simulator if deployToDevice() fails. I think we want it to only do this if checkDeviceConnected() fails. So, you should:

    return checkDeviceConnected().then(function () {
        return deployToDevice(appPath);
    }, function () {
        return deployToSim(appPath, args.target);
    });

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -0,0 +1,184 @@
...
+            console.warn('Found multiple .xcodeproj directories in \n' +
+                projectPath + '\nUsing first one');
+        }
+
+        var projectName = path.basename(xcodeProjFiles[0], '.xcodeproj');
+
+        var appPath = path.join(projectPath, 'build', (args.device ? 'device' : 'emulator'), projectName + '.app');
+        // select command to run and arguments depending whether
+        // we're running on device/emulator
+        if (args.device) {
```
agrieve: I don't think this is quite right. If `--device`, then we want to only attempt `deployToDevice()`. `checkDeviceConnected()` should run only if neither flag is passed.

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -0,0 +1,184 @@
...
+        // TODO: Is there any other way to get emulator's output to use in log command?
+        '--stderr', logPath, '--stdout', logPath,
+        '--exit'];
+    return spawn('ios-sim', simArgs);
+}
+
+function listDevices() {
+    var listDevicesScript = path.join(cordovaPath, 'lib', 'list-devices');
+    console.log("Available iOS Devices:");
+    return spawn(listDevicesScript);
```
agrieve: why spawn these rather than require() them?

bin/lib/versions.js
vladimir-kotikov: Updated in https://github.com/MSOpenTech/cordova-ios/commit/079fcf3c32aa8fce98ad68ee062aaff00cb0dab3

bin/lib/check_reqs.js
vladimir-kotikov: Updated in https://github.com/MSOpenTech/cordova-ios/commit/079fcf3c32aa8fce98ad68ee062aaff00cb0dab3

bin/lib/versions.js
vladimir-kotikov: Updated in https://github.com/MSOpenTech/cordova-ios/commit/079fcf3c32aa8fce98ad68ee062aaff00cb0dab3

bin/node_modules/nopt/.npmignore
vladimir-kotikov: This file introduced by nopt, not by me. Anyway we're shipping node dependencies bundled in this package so this won't affect en-user.
@agrieve Your thoughts?

bin/templates/scripts/cordova/lib/run.js
vladimir-kotikov: Updated in https://github.com/MSOpenTech/cordova-ios/commit/079fcf3c32aa8fce98ad68ee062aaff00cb0dab3

bin/templates/scripts/cordova/lib/run.js
vladimir-kotikov: Don't knew that this will work. Thanks for advice!

bin/templates/scripts/cordova/lib/run.js
vladimir-kotikov: Updated in https://github.com/MSOpenTech/cordova-ios/commit/079fcf3c32aa8fce98ad68ee062aaff00cb0dab3

bin/templates/scripts/cordova/lib/run.js
vladimir-kotikov: Well, i just ported an original flow. Take a look:
```bash
    DEVICE_CONNECTED=$(ios-deploy -c)
    if [ $? != 0 ]; then
    	echo "No device is connected, trying Simulator." 1>&2;
        USE_SIMULATOR=true
    else
        # if we got here, we can deploy the app, then exit success
        if ! [ "$NO_BUILD" = true ]; then
            "$CORDOVA_PATH/build" --device || exit $?
        fi
        ios-deploy -d -b "$DEVICE_APP_PATH"
    	exit 0
    fi
```

But your point makes sense. Updated in https://github.com/MSOpenTech/cordova-ios/commit/079fcf3c32aa8fce98ad68ee062aaff00cb0dab3

bin/templates/scripts/cordova/lib/run.js
vladimir-kotikov: This because these scripts hadn't been ported. Converted them to Node in https://github.com/MSOpenTech/cordova-ios/commit/e4c200b79a9bf5903c07eb70d50841ed328a71cc

PR #136, opened 2015-04-09
[CB-8838]Moved commandQueue push into non-WK_WEBVIEW_BINDING branch.

AlexMouton:
Hi Shazron (et al.)
I am working on an html5 video game that is packaged for ios using cordova.
We were having memory issues with UIWebView and have integrated your work on WKWebView recently.
Its working great for the performance! thanks! but, we are of course feeling alpha growing pains.

One thing we have noticed is that when we invoke the cordova Barcode plugin, it is always called twice!

I have only looked at the code a short time, but it appears to me that:
* the first call to Barcode happens, and in iosExec, it gets placed in commandQueue.
* the first call passes to Objective C world by WKScriptMessageHandler - (void)userContentController:(WKUserContentController*)userContentController didReceiveScriptMessage:(WKScriptMessage*)message
* this call to Barcode does its thing, completes, and it's javascript response handler code winds through:
    CDVCommandDelegateImpl::sendPluginResult
    CDVCommandDelegateImpl::evalJSHelper
    CDVCommandDelegateImpl::evalJSHelper2
* in Helper2, the callback is finally passed to a webViewEngine, which processes the Javascript, and returns, on completion, a list of commands remaining in the javascript commandQueue.
* these commands are queued on the Obj-C side and run in CDVCommandQueue fashion.

AHA!

* since the command was persisted in commandQueue before execution, and not removed, it is called again!

SO

I moved the enqueue behavior into the branch that excludes bridgeMode === jsToNativeModes.WK_WEBVIEW_BINDING
This fixes it, but I am not confident its the right solution.

Do you have any insight into this problem?
Feedback for my solution?
A broader or more appropriate venue for the conversation?

Thanks!
Alex Mouton

### DIFF
--- a/CordovaLib/cordova.js
+++ b/CordovaLib/cordova.js
@@ -970,14 +970,14 @@ function iOSExec() {
 
     var command = [callbackId, service, action, actionArgs];
 
-    // Stringify and queue the command. We stringify to command now to
-    // effectively clone the command arguments in case they are mutated before
-    // the command is executed.
-    commandQueue.push(JSON.stringify(command));
-    
     if (bridgeMode === jsToNativeModes.WK_WEBVIEW_BINDING) {
         window.webkit.messageHandlers.cordova.postMessage(command);
     } else {
+        // Stringify and queue the command. We stringify to command now to
+        // effectively clone the command arguments in case they are mutated before
+        // the command is executed.
+        commandQueue.push(JSON.stringify(command));
+
         // If we're in the context of a stringByEvaluatingJavaScriptFromString call,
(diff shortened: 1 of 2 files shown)


shazron:
Hi, thanks for the PR. Please also send a PR to the cordova-js repo. 

The exec bridge code is undergoing a transition of repos, but for now we need to also update cordova-js. See: https://github.com/apache/cordova-js/blob/b4af1c5ec477dd98cd651932ea6df6d46705d7f9/src/ios/exec.js#L177

shazron:
FYI this is in 4.0.x, will be auto-closed.

PR #138, opened 2015-04-21
CB-8485 Support for creating signed archive for iOS

nikhilkh:
Allows the user to specify the provisioning profile and the signing identity using the command line and a build.json file.

### DIFF
--- a/bin/templates/scripts/cordova/build-extras.xcconfig
+++ b/bin/templates/scripts/cordova/build-extras.xcconfig
@@ -18,8 +18,5 @@
 //
 
 //
-// XCode Build settings for "Debug" Build Configuration.
+// Auto-generated config file to override configuration files (build-release/build-debug).
 //
-
-#include "build.xcconfig"
-#include "build-extras.xcconfig"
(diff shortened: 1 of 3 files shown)


nikhilkh:
@shazron @agrieve Wonder if you can take a look?

shazron:
Looks sound to me. I'll give it a spin later today before I pull it in.

mattdsteele:
Hi folks, I'm trying this out - looks really slick!

I had one question: this produces a signed `.app` file. What's the best way to convert it to an `.ipa` we can install on a device? I found [this StackOverflow](http://stackoverflow.com/a/25006630) post, is running `xcrun` the recommended approach?

shazron:
Yes xcrun

nikhilkh:
This change creates an ipa in addition to .app file by invoking xcrun: https://github.com/MSOpenTech/cordova-ios/commit/ebcc29ef9a2ca4ebe49fd319868a514c0a558499#diff-6a830db18b2f183f537e0d0f98275067R114

PR #139, opened 2015-05-04
[CB-8948] Clipboard fix for iOS safari copy

markharding:
Using the iOS helper copy widget, its not possible to paste into any HTML5 inputs. A possible workaround is to grab from the clipboard and re-copy when the app resumes.


shazron:
Hi Mark, not sure I understand what scenario this code helps fix? Any examples online?

markharding:
Checkout http://voxpelli.com/2015/03/ios-safari-url-copy-paste-bug/. It's actually a safari bug but this patch does get around the issue.

PR #140, opened 2015-05-11
CB-9013: Fix listing of multiple devices in list-devices for iOS

gpetrov:
when multiple of same type devices are connected, like two iPhones,
only the first was returned.

Now all connected devices will be returned.

Note it might be even better if there was a way to get the devices
names as descriptions because now they are all called ‘iPhone’


shazron:
Is this similar?
https://issues.apache.org/jira/browse/CB-8907

If not, can you file a separate issue at https://issues.apache.org/jira/browse/CB
and add a CB-XXXX prefix to the issue title (where XXXX is your issue number).

gpetrov:
@shazron is different issue indeed, the other is about the emulators. Mine is about the connected devices list.

Created: https://issues.apache.org/jira/browse/CB-9013

PR #144, opened 2015-06-04
CB-9116 Ensure platform scripts are executable as the platform could be added on non-OSX environment

daserge:
Corresponding [Jira issue](https://issues.apache.org/jira/browse/CB-9116)

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -87,6 +87,13 @@ module.exports.run = function (argv) {
         }
         return Q.nfcall(fs.writeFile, path.join(__dirname, '..', 'build-extras.xcconfig'), extraConfig, 'utf-8');
     }).then(function () {
+        // Ensure platform scripts are executable as the platform could be added on non-OSX environment
+        shell.find(projectPath).filter(function (file) {
+            return file.match(/\.sh$/i);
+        }).forEach(function (file) {
+            fs.chmodSync(file, '755');
+        });
+    }).then(function () {
         var configuration = args.release ? 'Release' : 'Debug';
 
         console.log('Building project  : ' + path.join(projectPath, projectName + '.xcodeproj'));


bin/templates/scripts/cordova/lib/build.js
```diff
@@ -87,6 +87,13 @@ module.exports.run = function (argv) {
         }
         return Q.nfcall(fs.writeFile, path.join(__dirname, '..', 'build-extras.xcconfig'), extraConfig, 'utf-8');
     }).then(function () {
+        // Ensure platform scripts are executable as the platform could be added on non-OSX environment
+        shell.find(projectPath).filter(function (file) {
```
sgrebnov: Should we check `cordova` folder only? (projectPath/cordova). My concern is that there could be too many files in other folders like `www` so this could be slow.

daserge: Makes sense; although there is another script there: 
`ios/build/HelloCordova.build/Debug-iphonesimulator/HelloCordova.build/Script-304B58A110DAC018002A0835.sh`
So we can probably check all folders except `www` - what do you think?

omefire:
LGTM !
- what causes file permissions to change when the platform is added on a non-OSX system ?

daserge:
@omefire I think that it is because git does not include exec permission for these files.

omefire:
@daserge , does this behavior show up if for example, I add android from windows, then check it into git ?

daserge:
@omefire I don't have Android environment on Mac now, but as far as I can see there are no shell scripts in Android platform.

bin/templates/scripts/cordova/lib/build.js
sgrebnov: sounds good

nikhilkh:
Does it make sense to convert any of these scripts to node instead of 'shell'?

omefire:
It'd be a great idea to convert the scripts from 'shell' to 'node'.
bash is hard to follow and make sense of.

daserge:
Closing this - decided to rewrite the scripts in node.

PR #145, opened 2015-06-08
* is a Right-to-left associative operator in C and Objective-C

Coeur:
See:
http://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B#Operator_precedence
http://www.techotopia.com/index.php/Objective-C_2.0_Operator_Precedence#Objective-C_Operator_Precedence_and_Associativity

### DIFF
--- a/bin/uncrustify.cfg
+++ b/bin/uncrustify.cfg
@@ -277,7 +277,7 @@ sp_balance_nested_parens                 = false    # false/true
 sp_paren_brace                           = force    # ignore/add/remove/force
 
 # Add or remove space before pointer star '*'
-sp_before_ptr_star                       = remove   # ignore/add/remove/force
+sp_before_ptr_star                       = force   # ignore/add/remove/force
 
 # Add or remove space before pointer star '*' that isn't followed by a variable name
 # If set to 'ignore', sp_before_ptr_star is used instead.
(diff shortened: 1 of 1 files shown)


shazron:
Not sure I see the entire point of this, we've decided Objective-C object pointer declaration should be "NSString* foo" instead of "NSString *foo". This change would potentially change everything in the existing codebase (so git blame will be less useful).

purplecabbage:
I agree with Shazron. It is a semantic discussion without benefit, and ultimately would obscure other's contributions.

Coeur:
But the original decision was semantically incorrect. It teaches a bad habit. And `blame` can also be configured to hide specific contributions, like the ones related to uncrustify.

Coeur:
Also, the day you move to, let say, Swift, you won't avoid changing latest contribution to someone else.

cjpearson:
The current style is not incorrect or a bad habit. It may not follow Apple's style, but both styles are valid and readable.

Many C programmers choose to attach the asterisk to the variable, emphasizing syntax, while most C++ programmers prefer to emphasize the type by attaching the asterisk to the type.

Since Objective-C, like C++, is an Object Oriented language, I think it makes much more sense to emphasize type by keeping the asterisk attached to the type.

Also, there are several common situations in Objective-C where there is no variable for you to attach the asterisk to even if you wanted.

Casting

`(NSString*)someObject`

Method Parameters

`- (void)someMethod:(NSObject*)param;`

Generics

`NSArray<NSString*>* someArray;`

The common defense against grouping the asterisk with the type is that one might think these two lines behave the same.
    
    int* a, b, c;
    int *a, *b, *c;

I think that's a pretty easy situation to avoid. Just declare your variables on different lines.

For whatever it's worth, I vote keeping the style the way it is.

shazron:
Stroustrup's take on this (who I think we can agree is an authority): 
http://www.stroustrup.com/bs_faq2.html#whitespace

```
The critical confusion comes (only) when people try to declare several pointers with a single declaration:

	int* p, p1;	// probable error: p1 is not an int*
```

Other valid points in his FAQ item as well.

PR #146, opened 2015-06-11
CB-8197 Switch to nodejs for ios platform scripts

daserge:
Convert `copy-www-build-step.sh` to nodejs
Adds `glob` module and `bundledDependencies` to `package.json`

[Jira issue](https://issues.apache.org/jira/browse/CB-8197)


bin/templates/project/__CLI__.xcodeproj/project.pbxproj
```diff
@@ -374,7 +374,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "cordova/lib/copy-www-build-step.sh";
+			shellScript = "node cordova/lib/copy-www-build-step.js";
```
nikhilkh: This change assumes `node` to be in the $PATH for this to work. I'm not familiar with OS X and I'm assuming this assumption does not cause a problem.

nikhilkh:
LGTM. Adding `glob` does increase the size of ios platform considerably. I'm assuming there is no other cheaper way to do this.

bin/templates/project/__CLI__.xcodeproj/project.pbxproj
shazron: I thought about this earlier when the PR came in and I think its a good assumption since it's a Cordova project and the user is using the CLI. The scenario where a user would take the project from platforms/ios and run it stand-alone is miniscule. 

Not sure if this is supported fully but Xcode.app does include node.js under `/Applications/Xcode.app/Contents/Developer/usr/share/xcs/Node/bin/node`

You would dynamically get this path by using `xcode-select --print-path` then appending `/usr/share/xcs/Node/bin/node`

daserge: Hello @shazron, @nikhilkh,
Thanks for your review.
Regarding node' path - I propose to leave it as is taking into account that stand-alone case is uncommon.
We can address it later in case of issues.

purplecabbage: +1 The cli dependent on node, so it makes sense to require it in the path.
Does this affect users who are not using the cli? Probably still not a big deal either way.

PR #151, opened 2015-08-17
CB-9500 iOS build no sign arg

njtman:
As of Cordova iOS 3.9.0, Xcode signing is automatically added to the build step when compiling for device.  There are use cases where we would not want this to automatically happen (see CB-9500).  An argument has been added to skip the code signing when building for an iOS device.  Simply supply the   

--noSign

argument when running the compile / build command for iOS to skip code signing.


njtman:
Sorry, I messed up this pull request.  I will be opening up another shortly.

PR #152, opened 2015-08-17
CB-9500 Added no sign argument to iOS build

njtman:
As of Cordova iOS 3.9.0, Xcode signing is automatically added to the build
step when compiling for device.  An argument has been added to skip
the signing step (xcrun command) when building for an iOS device. 

Simply supply:

--noSign

argument when running the compile / build command to skip iOS app signing.

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -199,6 +199,7 @@ module.exports.help = function help() {
     console.log('    --codeSignIdentity      : Type of signing identity used for code signing.');
     console.log('    --codeSignResourceRules : Path to ResourceRules.plist.');
     console.log('    --provisioningProfile   : UUID of the profile.');
+    console.log('    --device --noSign       : Builds project without application signing.');
     console.log('');
     console.log('examples:');
     console.log('    build ');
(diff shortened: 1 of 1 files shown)


nikhilkh:
Another option is to use 'xcode' node module to parse the `productname` which corresponds with the .app name and find it .app file to sign. Alternative, shell.ls can be used to find the .app file and create an ipa. 

I want to question why would you have a different name for the .xcodeproj and the app name? I noticed, we allow changing the name in config.xml and thereafter being able to update .xcodeproj name and the product name - does that not work for you?

njtman:
@nikhilkh 
I am aware of the name attribute in the config.xml.
We are setting the productname in the xcode build file because we have run into issues where setting a name with spaces or special characters, doesn't play nicely with xcode project names.

In our use case, we set the cordova project name to a very simple name - one with no spaces or special characters. We can then change the app's display name for ios or android (sometimes slightly different between the platforms due to name wrapping on Android) to almost anything we want, without having to worry about xcode or gradle project incompatibilities.  

I like the idea that you suggested about using the xcode node module to parse the project name.  The no wrap option is not ideal, but it was the simplest and easiest to implement, as our custom build environment already signs our applications for us.

shazron:
@nikhilkh any objections to this getting pulled in?

nikhilkh:
This looks good. Let's get it in!

shazron:
Hi @njtman I couldn't find your name (from the patch's From: field) in the Apache CLA list: http://people.apache.org/committer-index.html

Once you have filed one, I can proceed with merging this in. Instructions here: https://www.apache.org/licenses/icla.txt

njtman:
@shazron I have signed and submitted my icla to secretary@apache.org.
Thanks!

PR #153, opened 2015-08-18
CB-9462 iOS 3.9.0 breaks npm link modules

daserge:
Using `shell.exec` as `shelljs` does not support symlinks dereferencing [yet](https://github.com/arturadib/shelljs/pull/164).

[Jira issue](https://issues.apache.org/jira/browse/CB-9462)


rayshan:
Thanks!

PR #154, opened 2015-08-18
CB-9512 Cordova-ios build/run should fail on Windows with a more mean…

daserge:
…ingful message

[Jira issue](https://issues.apache.org/jira/browse/CB-9512)

### DIFF
--- a/bin/templates/scripts/cordova/build.bat
+++ b/bin/templates/scripts/cordova/build.bat
@@ -1,19 +1,19 @@
 :: Licensed to the Apache Software Foundation (ASF) under one
-:: or more contributor license agreements.  See the NOTICE file
+:: or more contributor license agreements. See the NOTICE file
 :: distributed with this work for additional information
-:: regarding copyright ownership.  The ASF licenses this file
+:: regarding copyright ownership. The ASF licenses this file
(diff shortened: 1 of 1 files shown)


bin/templates/scripts/cordova/run.bat
```diff
@@ -15,11 +15,5 @@
...
-SET script_path="%~dp0run"
-IF EXIST %script_path% (
-        node %script_path% %*
-) ELSE (
-    ECHO.
-    ECHO ERROR: Could not find 'run' script in 'cordova' folder, aborting...>&2
-    EXIT /B 1
-)
+ECHO ERROR: Running 'ios' platform on Windows is not supported>&2
+EXIT /B 0
```
robpaveza: Why are you exiting with a success status?  The previous version exited with status code 1.  

Probably should also change build.bat to return status code 1.

daserge: It allows to build&run supported platforms without interruption in case we have several platforms added ([Jira issue with example](https://issues.apache.org/jira/browse/CB-9512)).

purplecabbage:
I have essentially done the same thing in #158, but for build/log/run/clean

daserge:
Great, thanks, closing this in that case :)

PR #160, opened 2015-09-09
CB-9328 Use ios-sim as a node module, not a CLI utility

macdonst:


shazron:
`require('ios-sim')` is called here, but how is the module installed? 
ideally it should be installed when you create the project, inside the project, or part of cordova-lib.
It should be part of the platform, if the CLI could run "npm install" on the platform.
Trying to avoid making the user run "npm install ios-sim"

macdonst:
@shazron thanks for the feedback, I meant to ask about including npm modules along with the cordova-ios. This way folks would not need to `npm install ios-sim` right?

shazron:
yup. But not sure where we would add this dependency... we could just check-in the `node_modules` in the `bin/templates/scripts/cordova/lib` folder (after installing ios-sim there). There is already a `node_modules` in `bin`.

macdonst:
@shazron okay, I think I've got it. I don't want to merge until I get a second set of eyes though.

shazron:
thanks @macdonst I suppose this needs to get the patched 5.0.1?

shazron:
Just tested with `cordova emulate` -- verified I didn't have any ios-sim installed anywhere before I did so. I suppose 5.0.1 is not needed since that was a patch for the CLI way anyways..

shazron:
We also can get rid of all checks for the ios-sim requirement: https://github.com/apache/cordova-ios/blob/f2ade2282b56a560199f493a5feb32481a16b6d5/bin/lib/check_reqs.js#L58-L60

macdonst:
@shazron Yeah, I didn't bother to update to 5.0.1 as it only affects the project when used as a CLI but I can do that if you'd like.

I actually did get rid of the checks before but forgot to stage the file so I just updated the commit.

shazron:
Nah no need to update to 5.0.1, thanks Simon! I'll give it a spin

PR #163, opened 2015-09-15
CB-9558 Blob schemes won't load in iframes

purplecabbage:


shazron:
+1. Just a note about the removal of the about: scheme in the commit itself

PR #164, opened 2015-09-22
CB-9685 A fix for the magnifying glass popping up on iOS9 longpress

EddyVerbruggen:
I thought `CDVLongPressFix` would not be the best name. Especially if we ever need to fix any other gestures Apple breaks for us, so I went for `CDVGestureHandler`.

Btw, a check for `IsAtLeastVersion(@"9")` was omitted on purpose because it doesn't hurt on lower versions (it even suppresses other longtap glitches we've been seeing for a long while now).

I hope you like it!

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVGestureHandler/CDVGestureHandler.h
+++ b/CordovaLib/Classes/Private/Plugins/CDVGestureHandler/CDVGestureHandler.h
@@ -1,0 +1,26 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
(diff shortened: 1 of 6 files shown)


chaffeqa:
:+1:

zhangjianying:
good ,3q 4 u time

Xhamps:
:+1:

EddyVerbruggen:
I think I'll be adding a configuration for this to config.xml as not all folks may like this feature (new PR).

The plugin that inspired this PR has been around for a few weeks now: https://github.com/EddyVerbruggen/cordova-plugin-ios-longpress-fix and while most like it, it may break text selection. If your app needs that this feature will cause trouble.

I propose enabling this private plugin only when a certain config.xml property has been enabled.

And I'd also like to add a feature to suppress the magnifying glass for devices with 3D Touch support. In the plugin mentioned above I've already done so, by specifying a preference: `<preference name="Suppress3DTouch" value="true" />`.

Any thoughts others want to share perhaps?

zhangjianying:
good idea. thinks

jcesarmobile:
+1 for preference

BTW, I'm not able to reproduce the magnifier glass issue on iOS 9.1, maybe apple fixed it?

EddyVerbruggen:
BTW I am seeing it on 9.1. iPhone 6S, longpress and 3D Touch.

shazron:
Sounds good, can you get the code in a new pull request and reference the same CB-9685? I'll re-open it once you do so.

EddyVerbruggen:
Thanks for your feedback guys, I've opened a new PR #174.

jcesarmobile:
I've tried on iPad 4, iPod touch 5 and iPhone 6 and the bug is gone on 9.1. On the iPad and iPod tested the same app before updating (magnifying glass appeared) and after updating (magnifying glass didn't appear).
Maybe they forgot to fix it on devices with 3D Touch.
Can anybody test on more devices with 9.1?

PR #167, opened 2015-09-23
CB-9690 Can't submit iPad apps to the App Store for iOS 9

macdonst:
Closes something


devgeeks:
Doesn't that completely opt out of "Slide Over and Split View" ?

tripodsan:
I think this should be revisited, once the multi-tasking support is enabled (CB-9161)

macdonst:
Yes, that is the intention. Comments on the JIRA ticket don't show up in this PR but that is our plan.

ptarjan:
this fixes half of https://issues.apache.org/jira/browse/CB-9729 (can you release a point release with this in it please)?

Also, is the other pList problem a similar fix?

PR #170, opened 2015-10-02
CB-9719 set enable_modules build setting to no

eymorale:
In 3.x, this setting was set to no. It currently breaks frameworks that have CordovaLib as a dependency since it is not a module.

### DIFF
--- a/bin/lib/check_reqs.js
+++ b/bin/lib/check_reqs.js
@@ -25,7 +25,7 @@ var XCODEBUILD_MIN_VERSION = '6.0.0';
 var XCODEBUILD_NOT_FOUND_MESSAGE =
     'Please install version ' + XCODEBUILD_MIN_VERSION + ' or greater from App Store';
 
-var IOS_DEPLOY_MIN_VERSION = '1.7.0';
+var IOS_DEPLOY_MIN_VERSION = '1.8.0';
 var IOS_DEPLOY_NOT_FOUND_MESSAGE =
     'Please download, build and install version ' + IOS_DEPLOY_MIN_VERSION + ' or greater' +
     ' from https://github.com/phonegap/ios-deploy into your path, or do \'npm install -g ios-deploy\'';


shazron:
LGTM. 
However, the side effect of this is, if any modules use the module import syntax, ie `@import UIKit;` this might break I think, but the workaround there is to use the old import method, so its fine.

The final fix would be to make CordovaLib as a module, but that's a lot more work potentially.

csantanapr:
I did some investigation today, I found the reason why putting 
CLANG_ENABLE_MODULES = NO
in build.xcconfig doesn't get reflected, is because build.xcconfig is for Project, not the Target, the Project file is set to Yes, and Project file has higher precedence than Project xcconfig

For the variable assingment to be able to work it needs to be set in Target's release/build or extras xcconfig

There is another value that also fixes our problem with a framework referring to a non-module header/class from CordovaLib
CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES

From looking at all levels in XCode, it looks like the iOS Platform default is:
CLANG_ENABLE_MODULES = NO and CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = NO
But XCode 7 when it creates a new projects, overwrites the default of one of them to CLANG_ENABLE_MODULES = YES

It might be safer to set 
CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES
which I think is more closer to the compile error that we are getting, and maybe has less side effects.

By Google Power I discover this page, explains variable assignment precedence:
http://pewpewthespells.com/blog/xcconfig_guide.html

What you all think?
And what would be the best to assign it? Project (build.xcconfig) or Target (build-debug.xcconfig and build-release.xcconfig)
We could use extras but I would like to leave extra empty for the developer/user to overwrite values from Target and Project xcconfig programmatically

shazron:
Regarding the xcconfig - for command line builds, anything set by the xcconfig flag should override everything. This is what the setting from the man page for xcodebuild says:
```
-xcconfig filename
           Load the build settings defined in filename when building all targets.  These set-
           tings will override all other settings, including settings passed individually on
           the command line.
```

So I'm confused, unless Xcode 7 has a bug in it. Couldn't verify unless we have a test project to get the failure case, and to test a fix.

Thanks for investigating this. Like you said, try setting CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES in build.xcconfig to see if it fixes your problem?. build.xcconfig is the base xcconfig for all Build Configurations.

PR #171, opened 2015-10-07
CB-9719 set allow_non_modular_includes to yes

eymorale:
Ran into some issues with git so did a new pull request. @shazron @csantanapr I added that setting change in build.xcconfig and confirmed that it picks up the setting both in xcode and in the cli


shazron:
I fixed the test failure above and ran the tests myself, they pass.

PR #172, opened 2015-10-16
Allow about:blank to fix twitter embed

ilicgo:
Twitter widget don't work, because the twitter javascript create an iframe without src attribute. This made a request with the scheme about (about:blank). Like android platform we allow about scheme. We don't know if it is a good idea, but it solve our issue with twitter embed.


shazron:
This PR  is unusable since it has conflicts. Please rebase against master, not the 3.9.x branch
Note that it should be "clean" in that it should only reflect your proposed commits, not others (see the PR commit graph for your PR in Github)

shazron:
For cordova-ios-4.0 (to be released next week), use the `<allow-navigation href="about:*" />` in config.xml to allow this now.

PR #174, opened 2015-10-24
CB-9685 Enhanced the fix for the iOS9 longpress magnifying glass issue

EddyVerbruggen:
This feature is now disabled by default and can be enabled for longpress in `config.xml`:
`<preference name="SuppressesLongPressGesture" value="true" />`

When 3D Touching the webview (apply pressure during your touch) the magnification glass will also pop up. It won't even wait for a longpress in this case. To disable this issue as well, add this to `config.xml`:
`<preference name="Suppresses3DTouchGesture" value="true" />`

Note that if the former is true, the latter is ignored since they essentially use the same gesture `UILongPressGestureRecognizer` (which is a shame but that's how Apple implemented it). The other way around works fine, so to have the behavior of the previous version of this plugin, you'd do:

```xml
    <preference name="SuppressesLongPressGesture" value="true" />
    <preference name="Suppresses3DTouchGesture" value="false" />
```

Omitting the properties is the same as setting both to`false`.

The reason for having the developer explictly enabling these features is that disabling longpress may break text selection (most relevantly in input fields).
The reason there's a sep property for 3D Touch is it will break onclick handlers, but ontouchend works just fine, so it really depends on your app/framework if it's a workable fix.

Please merge this so I can sleep better at night.


chaffeqa:
:+1:

shazron:
The only thing new is 371c136315b0c4f481ee479242dd769c4280b932 correct?

shazron:
Pulled in. Can you add docs too, please? Thanks :) 
Here: https://github.com/apache/cordova-docs/blob/master/www/docs/en/edge/guide/platforms/ios/config.md

EddyVerbruggen:
@shazron Of course, overlooked that! See [404](https://github.com/apache/cordova-docs/pull/404).

manfield:
The problem was solved, but not everywhere.
In the InAppBrowser view, the problem still occurs (cordova-ios: 4.4.0).

shazron:
@manfield this only applies to the main webview at startup. Since InAppBrowser is launched later, it is not applied. https://github.com/EddyVerbruggen/cordova-ios/blob/371c136315b0c4f481ee479242dd769c4280b932/CordovaLib/Classes/Private/Plugins/CDVGestureHandler/CDVGestureHandler.m#L56-L61

manfield:
Is it a behavior that should be fixed or I should apply some configuration in my code as the webview is called?

shazron:
don't know - it's definitely not a cordova-ios issue but a plugin issue.

manfield:
Alright. Thanks.
I posted the issue on Apache Foundation Jira, mentioning the cordova-plugin-inappbrowser plugin.

CVeniamin:
I can confirm @manfield issue. 
I also use InAppBrowser and this commit solves the problem for CordovaWebView but then when opening a InAppBrowser instance magnification glass comes back and shows content that resides behind InAppBrowser. 
By completely ignoring the InAppBrowser instance and without ability to be disabled when using cordova-plugin-inappbrowser it is a bit annoying.

CVeniamin:
@manfield, I think that I found the problem with InAppBrowser it's showing wrong content (from behing InAppBrowser) on magnifying glass .

PR #175, opened 2015-10-26
Prepare cordova-ios 3.9.2 release

vladimir-kotikov:


shazron:
This should make it in as well: https://issues.apache.org/jira/browse/CB-9721

vladimir-kotikov:
@shazron, done, added https://github.com/MSOpenTech/cordova-ios/commit/51bf20cea66813e188c57127c5fdc32f9a625cdd.

Please also review deprecation warnings added in https://github.com/MSOpenTech/cordova-ios/commit/3d69d8c37e025eea8ac479b1f2396931c60f4da4. Not sure if i done this correctly, as i see now a lot of deprecation warnings in XCode, since those deprecated APIs still used internally.

shazron:
I think that's all that's needed. Thanks @vladimir-kotikov !

vladimir-kotikov:
@shazron, updated

shazron:
@vladimir-kotikov Sorry I had to update this, it was wrong: https://github.com/apache/cordova-ios/commit/29f10b475414e84121e2e77f79095b6bccdc6b35

jtomaszewski:
Agree with @akofman , After update to xcode, I need to update pbxproj as well to make the effect work.

jtomaszewski:
You can use this hook to fix it https://gist.github.com/jtomaszewski/3229031e352730b0bab0 .

akofman:
@jtomaszewski : or just this plugin https://github.com/akofman/cordova-plugin-disable-bitcode ;)

PR #177, opened 2015-11-03
CB-9836 Add .gitattributes to prevent CRLF line endings in repos

daserge:
Updates gitattributes to common web project template to help git recognize binary files

[Jira issue](https://issues.apache.org/jira/browse/CB-9836)


shazron:
looks good to me

PR #178, opened 2015-11-03
CB-9883 - Remove unused iOS bridges

shazron:
Please review.


csantanapr:
With this removal, is there a minimum ios version that this will work only using the IFRAME_NAV bridge?

shazron:
IFRAME_NAV has always worked from the beginning and has been the default, the other bridge modes are experimental to try to gain speed, but flakey -- like the comments in the .js say.

csantanapr:
@shazron is this a lesson learned that certain experimental things should be kept in experimental branches and not in master :+1:

PR #179, opened 2015-11-09
CB-9883

purplecabbage:
Also remove the wk-webview bridge as the plugin itself simply clobbers cordova.exec now.
1 Bridge! You shall not pass!

### DIFF
--- a/CordovaLib/cordova.js
+++ b/CordovaLib/cordova.js
@@ -1,5 +1,5 @@
 // Platform: ios
-// 7703d38498920bdcfadd574e475553a26fc490e5
+// 8e9610fe33fc743fcaf5d920064f0deb2cad1715
 /*
  Licensed to the Apache Software Foundation (ASF) under one
  or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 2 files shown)


shazron:
It's the Bridge of IFrame-Dûm! Good work Olórin of the Cabbages Purple

chaffeqa:
I don't know if this was the path intended, but the new changes break the falling back to UIWebView since simply including the WKWebview clobbers `cordova.exec` (and the browser proceeds to barf on a non-existent object on `window.webkit.messageHandler` when running via the UIWebView)

Understandable if this is simply a move away from falling back to UIWebView, but my understanding was that you would still gracefully degrade support to UIWebView (for instance if iOS 8 and not using local webserver)

shazron:
Good point, didn't think of that scenario. Ideally each WebView engine, when activated, should inject its own bridge. This would solve this scenario. 

We could do this for the WKWebViewEngine plugin -- when it is activated, read the .js file and inject it. The UIWebViewEngine (private) plugin could do it as well in the same manner but we would have to remove the .js from cordova-js-src folder (which is aggregated by cordova-js to create cordova.js) and leave this as a concern for the platform itself.

purplecabbage:
I'll fix that.

purplecabbage:
Is this sufficient in the wkwebview plugin?
```
if(window.webkit && window.webkit.messageHandler) {
    module.exports = iOSExec;
}
else {
    // this will end up clobbering cordova.exec with cordova.exec
    // a special case for falling back to UIWebView
    module.exports = cordova.exec; 
}
```
It will prevent the clobber in this case.

chaffeqa:
That looks great to me! honestly i dont know how i felt about the idea of dynamically injecting a script... call me old school ;)

Thanks guys a ton for all the hard work on this platform! super appreciated!

purplecabbage:
Are you able to verify this works @chaffeqa?

PR #185, opened 2015-12-21
Add additional valid targets for simulation

busterc:
Targets Added:
- iPhone-6s
- iPhone-6s-Plus
- iPad-Air-2
- iPad-Pro


shazron:
Thank you! Manually testing now, and will merge.

PR #186, opened 2015-12-22
CB-10233: Support different config.xml file per CDVViewController instance

mirkoluchi:
- Add a configFile property to CDVViewController (path to the config
file, absolute or relative to main bundle)
- Move loading of settings from -init to -viewDidLoad


shazron:
Thanks! Since this is a non-trivial change and is a critical path -- I would like to see some tests. See the `tests` folder and load the .xcworkspace to see the existing unit tests.

mirkoluchi:
Hi, I added some tests to assert that the correct config file is loaded.
If you this other regression tests are needed, please tell me and I'll try to figure out how to do that.

shazron:
Looks good to me, I'll merge this in after testing.

PR #187, opened 2015-12-24
Use "cdv_" prefixed scrollView for UIView to avoid conflicts

tualatrix:
After upgrading cordova-ios from 3.9.x to 4.0.x, I found there's a conflict for my custom UIView and cordova. It is caused by the category in Cordova, CDVPlugin.h:

```
@interface UIView (org_apache_cordova_UIView_Extension)

@property (nonatomic, weak) UIScrollView* srollView;

@end
```

As an extension that used internally in Cordova, it should not use a common name of "scrollView". If developer's project have a custom UIView that has a scrollView, the project will be broken.

So I did this pull request that rename this "scrollView" to "cdv_srollView".

Please take a review, thanks.

### DIFF
--- a/CordovaLib/Classes/Public/CDVPlugin.m
+++ b/CordovaLib/Classes/Public/CDVPlugin.m
@@ -25,11 +25,11 @@
 
 @implementation UIView (org_apache_cordova_UIView_Extension)
 
-@dynamic scrollView;
+@dynamic cdv_srollView;
 
-- (UIScrollView*)scrollView
+- (UIScrollView*)cdv_srollView
 {
-    SEL scrollViewSelector = NSSelectorFromString(@"scrollView");
+    SEL scrollViewSelector = NSSelectorFromString(@"cdv_srollView");
 
     if ([self respondsToSelector:scrollViewSelector]) {
         return ((id (*)(id, SEL))objc_msgSend)(self, scrollViewSelector);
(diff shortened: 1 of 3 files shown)


shazron:
I understand why this change should be made. However, since it is part of the public API (it is part of CDVPlugin), any changes to this will have to go in a major version (the next will be 5.0).

The original reason this was done (the Category) was as stated in the 4.0 Release Notes:
https://github.com/apache/cordova-ios/blob/6cc86efbb17f59f2a9626b0867406be697fd47ab/RELEASENOTES.md#400-dec-04-2015
"Make webView property dynamic in CDVViewController and CDVPlugin (from CDVWebViewEngineProtocol reference). Added scrollView category to UIView for backwards compatibility reasons."

I believe you also have a typo "cdv_srollView" should be "cdv_scrollView" (missing a 'c' in scroll)

CordovaLib/Classes/Public/CDVPlugin.m
```diff
@@ -25,11 +25,11 @@ Licensed to the Apache Software Foundation (ASF) under one
...
 @implementation UIView (org_apache_cordova_UIView_Extension)
 
-@dynamic scrollView;
+@dynamic cdv_srollView;
 
-- (UIScrollView*)scrollView
+- (UIScrollView*)cdv_srollView
 {
-    SEL scrollViewSelector = NSSelectorFromString(@"scrollView");
+    SEL scrollViewSelector = NSSelectorFromString(@"cdv_srollView");
```
purplecabbage: typo: @"cdv_scrollView"

purplecabbage:
Closing this because it is stale.
If you want to change cdv_srollView to cdv_scrollView, and rebase, then please reopen and resend.
Thanks for contributing!

PR #188, opened 2015-12-30
Stay attached to LLDB when running app

saghul:
https://github.com/apache/cordova-ios/commit/ba7bafdbd8d4dab29899a7a804a6bf8ad6893e51 added the `justlaunch` flag to `ios-deploy`, which conflicts with `-d`. The previous behavior was to stay attached to the debugger when the application launched, and by reading the issue in Jira I found no traces of why this behavior was overturned.

Ideally I'd like to continue to deploy to the device and stay attached to the debugger without using Xcode.

### DIFF
--- a/bin/templates/scripts/cordova/lib/run.js
+++ b/bin/templates/scripts/cordova/lib/run.js
@@ -133,9 +133,9 @@ function checkDeviceConnected() {
 function deployToDevice(appPath, target, extraArgs) {
     // Deploying to device...
     if (target) {
-        return spawn('ios-deploy', ['--justlaunch', '-d', '-b', appPath, '-i', target].concat(extraArgs));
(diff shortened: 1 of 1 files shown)


shazron:
I apologize this wasn't made clear in a JIRA issue. This is to make the behaviour consistent with all the other platforms, where after deploying they always detach. We didn't change this in cordova-ios-3.x, waiting until the major version bump of 4.0.

saghul:
Oh, I see. Is there a way to attach to the debugger after the app has launched without using Xcode? I found that quite useful.

Thanks!

shazron:
It's possible, but non-trivial.
ios-deploy has to write out a script that is read by lldb. Here's the tokenized version: 
https://github.com/phonegap/ios-deploy/blob/master/src/ios-deploy.c#L30-L41

Then you launch the script using `lldb -s [script_path]`

PR #190, opened 2016-01-21
Disable ios-deploy wifi mode when deploying to a device

mlegenhausen:
Currently when running `cordova run ios --device` the app gets deployed two times via usb and wifi. But when disconnecting from USB it is not possible to deploy at all, so the deployment fails. This issue is related to phonegap/ios-deploy#170 and disables this behavior to get the deployment working again, when using usb.


shazron:
Note that this has to be bumped up to 1.8.3 as well:
https://github.com/apache/cordova-ios/blob/842517eb7b5b0fdf8aefd3102372d182b155f7c4/bin/lib/check_reqs.js#L28

... since the --no-wifi option is only in 1.8.3.

mlegenhausen:
Thanks for the hint. Commit updated.

bitjson:
I was just looking for a solution to this, thanks! :+1:

PR #194, opened 2016-02-17
CB-10138  Adds missing plugin metadata to plugin_list module

imaffett:
This is the same fix as https://issues.apache.org/jira/browse/CB-10138 , but for Cordova iOS


vladimir-kotikov:
LGTM

PR #196, opened 2016-02-22
CB-10668 Moved node_modules to root

stevengill:
* updated package.json to include dependencies and bundledDependencies
* updated create.js to copy root node_modules instead of bin/node_modules
* removed bin/node_modules directory


nikhilkh:
I'm curious - Is this required for the new npm fetch work? Does it mean older versions of cordova-ios will stop working with the new npm fetch approach?

stevengill:
Nothing to do with fetch. We noticed `npm pack` was not properly adding in nested node_modules for the latest `cordova-osx` release. I took this opportunity to quickly update both osx and ios to use `package.json` to manage dependencies like our other platforms.

PR #199, opened 2016-02-28
CB-10673 fixed conflicting plugin install issue with overlapped <sour…

bso-intel:
…ce-file> tag
The problem is that copyNewFile is too strict for <source-file> tag.
You will never know which two plugins will write the library to the same target-dir of the source-file.
We should let it copy the library to the same location.


bso-intel:
@nikhilkh,
Could you review this?
Thanks.

PR #203, opened 2016-03-11
CB-11238 Expose supportedOrientations method

purplecabbage:
Expose supportedOrientations getter/setter so native code can override the current behavior.
I think this would require a minor version bump only, as we are exposing a new (native) API, but not breaking any existing functionality.


purplecabbage:
This pull-request is somewhat less important now that I have found a way to do this with the existing code.
This can still be added to a future minor release so that in the future this accessor is exposed publicly.

shazron:
Filed: https://issues.apache.org/jira/browse/CB-11238

Can you prefix the PR title with "CB-11238"? Thanks.

PR #207, opened 2016-03-17
- fix generate .ipa ios

ramonornela:


shazron:
Please file an issue on issues.cordova.io stating the exact problem, and how to reproduce it. This is not only for searchability, but also verifying the fix.

ramonornela:
Ok @shazron

shazron:
Any update on this? If not I will probably close this.

PR #211, opened 2016-03-24
CB-10942 - Cannot <allow-navigation href="https://foo.bar" /> for links in that site, if you have <allow-intent href="https://*" />

shazron:


jcesarmobile:
Tested and working fine

I think the first if can be removed
```
if (allowNavigationsPass && allowIntentPass) {
    return CDVIntentAndNavigationFilterValueNavigationAllowed;
}``` 

This one is enough
```
if (allowNavigationsPass) {
    return CDVIntentAndNavigationFilterValueNavigationAllowed;
}```

shazron:
I put that conditional in on purpose, to be explicit in the code what the rules are (I know the comment is in there as well) "unoptimized", it should be fine since algorithmically it doesn't add any complexity imo

jcesarmobile:
Then LGTM

barryvdh:
I can confirm this issue happening since v4.0.1 at least. v3.9.2 works fine.
All my app links now open in a separate browser window, which isn't supposed to.

Would be nice if this patch could be integrated :)

purplecabbage:
lgtm!

PR #212, opened 2016-04-03
CB-10962 CB-10945 removes justlaunch flag waiting for a fix in ios-deploy

akofman:


omefire:
This works, but it brings us back to when we used to have to exit lldb manually.

omefire:
I'd prefer this PR be merged into ios-deploy: https://github.com/phonegap/ios-deploy/pull/218
That way, we wouldn't have to fix this issue here.

akofman:
Sure I 100% agree with that. That's why I asked to @shazron what he thinks. If we have to wait 6 months more, I think it's better to keep something working instead of the actual situation.

omefire:
The thing is that this used to work just fine. Some recent changes might have broken it.

akofman:
Maybe it's related to the last XCode updates. What version are you using ?

omefire:
I'm using XCode Version 7.3 (7D175)

omefire:
I tested the following PR: https://github.com/phonegap/ios-deploy/pull/218
and unfortunately, it didn't fix the issue for me.

akofman:
hmmmm indeed.

omefire:
Could it be XCode update(from 7.2 to 7.3) that is causing this issue ?

akofman:
I think it's more than possible.

akofman:
XCode 7.2 is using lldb-340.4.119.1 while XCode 7.3 is using lldb-350.0.21.3.

omefire:
Hmmm, interesting

omefire:
I think we should pull this in, while waiting for a fix for the XCode 7.3 issue. What do you think @shazron ?

shazron:
This change will modify expected behaviour if someone is scripting this, since the run call will not terminate. That is, however, a lower priority consequence versus fixing this bug right now. I'll investigate a fix for ios-deploy for today, if there is no progress, we should proceed with this change.

omefire:
OK, sounds good. I'm running some investigations as well.

shazron:
Update (good news?): https://github.com/phonegap/ios-deploy/pull/218#issuecomment-206597622

shazron:
Everyone, try the new ios-deploy version 1.8.6.
This PR can be closed.

PR #213, opened 2016-04-06
CB-11019 Handle app name change more accurately

vladimir-kotikov:
This PR fixes potential bug in 'prepare' functionality, when changing application name (which assumes renaming of some folders inside of platform directory) does not update project's directories locations properly. Also, the `projectFile` cache needs to be invalidated because it could contain instance that refers the old paths.

See [CB-11019](https://issues.apache.org/jira/browse/CB-11019) for more details


sgrebnov:
:+1: lgtm

PR #214, opened 2016-04-07
CB-10957 Remove build-*.xcconfig files from build output

omefire:


shazron:
This PR removes crucial functionality, namely the inheritance of build settings from the .xcconfig files:
screen shot 2016-04-07 at 4 13 33 pm

It should only remove the .xcconfig files from here in Build Phases for the Target:
screen shot 2016-04-07 at 4 14 25 pm

omefire:
Thanks @shazron, will recheck it and make appropriate changes.

omefire:
can't find a good reference for the .pbxproj file, do you have any link ?

shazron:
Not sure what you mean to do, but I would just load the .xcodeproj file in Xcode and delete those references I mentioned in Copy Bundle Resources? (and only stage those changes, it will try to make other changes that are not needed, like removing quotes).

That might be more work than probably just removing these references:
screen shot 2016-04-08 at 10 08 14 am

as you can see they correspond to the screenshot I used in the last comment.

omefire:
Yes, that's what I just realized. I just loaded it in xcode and deleted those from the 'Build Phases' tab -> 'Copy Bundle Resources'. 
Much easier.

omefire:
@shazron could you please take another look ? This time, I did it via Xcode. I noticed minor diffs between before & after and I think it's due to my version of Xcode(diffs are mainly double quotes vs no quotes)

shazron:
The changes are not good - Xcode has a habit of removing those double quotes, which we need. I mentioned in my last comment " (and only stage those changes, it will try to make other changes that are not needed, like removing quotes)."

So probably manually remove the right changes I mentioned only.

shazron:
See staging patches -> https://git-scm.com/book/en/v2/Git-Tools-Interactive-Staging or use a gui tool like SourceTree to only stage those hunks you need

shazron:
Once you do the changes, you should run "npm test" (as you can see your patch has already failed the create tests in Travis)

omefire:
@shazron I'm merging this in. I've addressed your CR notes.

PR #219, opened 2016-04-28
CB-11161 Reuse PluginManager from common to add/rm plugins

vladimir-kotikov:
A change to use `PluginManager` class to manage plugin installation/uninstallation instead of custom logic

JIRA: [CB-11161](https://issues.apache.org/jira/browse/CB-11161)


nikhilkh:
@jasongin Can you please help review? Related to https://github.com/apache/cordova-android/pull/301

jasongin:
:+1:

PR #220, opened 2016-05-03
CB-11117: Optimize prepare for ios platform, clean prepared files

jasongin:
This uses the `FileUpdater` module added in apache/cordova-lib#429 to skip copying files that didn't change. Some refactoring was required because previously the target directories would just be wiped before copying; now we need to map out the source and target directories so the `FileUpdater` has the necessary information to determine the optimal set of file operations.

And I'm making it so that `cordova clean ios` also cleans the prepared files (but the `clean` shell script does not).


shazron:
This currently fails tests (because the cordova-common required is 1.3.0, and the one published on npm is still 1.2.0).

jasongin:
@shazron I know - a cordova-common 1.3 release is in progress.

shazron:
Squashed the commits, and ran the tests -- all green.

PR #224, opened 2016-05-18
CB-11265: Remove Target checking for cordova-ios

sarangan12:
This check does not serve any real purpose. But, if you already have a simulator which is not in the array, this will just stop the run. If you dont have this check and provide an invalid simulator, the only effect is that the build will continue and then the error message. 

@omefire @riknoll @nikhilkh Can you please review and merge this PR?


nikhilkh:
+ @vladimir-kotikov who originally added this code.

LGTM.

vladimir-kotikov:
LGTM. This code has been ported from bash script without any changes. I guess the purpose was to fail fast if user mispelled the simulator name.

PR #225, opened 2016-06-01
CB-11352: Call setUserAgent before creating a view

ankon:


shazron:
2 unit-test failures. (run the workspace in the tests subfolder).

1. https://github.com/apache/cordova-ios/blob/3d3a746cf86f63908da513a7c12001ee6dac7ebb/tests/CordovaLibTests/CDVUserAgentTest.m#L108 
2. https://github.com/apache/cordova-ios/blob/3d3a746cf86f63908da513a7c12001ee6dac7ebb/tests/CordovaLibTests/CDVUserAgentTest.m#L132

ankon:
Hmm. I'm quite new to objective c, so need to understand more how to fix that.

Thinking along the lines of giving the engine plugin the chance to report what the user agent would be, which WKWebView then would do by creating a temporary view. Alternatively: is there a point where the engine gets called and could safely recreate the view, some form of "yes, we're good, finalize your setup?"

Does that sound reasonable?

shazron:
The lines I quoted show the descriptions of what failed - you don't need to touch the tests, just fix your code. The tests run on the standard UIWebView, not WKWebView, and must pass before any PR is accepted.

In any case, in iOS 9, there is a "customUserAgent" property for WKWebView that should be set instead: https://developer.apple.com/library/ios/documentation/WebKit/Reference/WKWebView_Ref/#//apple_ref/occ/instp/WKWebView/customUserAgent and not the cordova workaround way.

oliviertassinari:
I have just tested this PR. It's working fine on iOS 9.3 with WKWebView.

PR #228, opened 2016-06-07
[CB-10916] Support display name for iOS

dpogue:
This is the iOS Platform API support side of https://github.com/apache/cordova-lib/pull/453.

### DIFF
--- a/bin/templates/scripts/cordova/lib/prepare.js
+++ b/bin/templates/scripts/cordova/lib/prepare.js
@@ -193,6 +193,7 @@ function updateProject(platformConfig, locations) {
     var name = unorm.nfd(platformConfig.name());
     var pkg = platformConfig.getAttribute('ios-CFBundleIdentifier') || platformConfig.packageName();
     var version = platformConfig.version();
+    var displayName = platformConfig.shortName && platformConfig.shortName();
 
     var originalName = path.basename(locations.xcodeCordovaProj);
 
@@ -208,6 +209,10 @@ function updateProject(platformConfig, locations) {
 
     if (platformConfig.getAttribute('defaultlocale')) {
         infoPlist['CFBundleDevelopmentRegion'] = platformConfig.getAttribute('defaultlocale');
+    }
+
+    if (displayName) {
+        infoPlist['CFBundleDisplayName'] = displayName;
     }
 
     // replace Info.plist ATS entries according to <access> and <allow-navigation> config.xml entries
--- a/tests/spec/unit/prepare.spec.js
+++ b/tests/spec/unit/prepare.spec.js
@@ -1485,7 +1485,13 @@ describe('prepare', function() {
                 expect(exceptionDomains['null']).toBeUndefined();
                 expect(exceptionDomains['undefined']).toBeUndefined();
             });
-        });   
+        });
+        it('Test#020 : <name> - should write out the display name to info plist as CFBundleDisplayName', function(done) {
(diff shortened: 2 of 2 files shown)


shazron:
Pull in when we bundle update to the released cordova-common that includes https://github.com/apache/cordova-lib/pull/453

shazron:
This PR needs to be re-based.

dpogue:
Rebased, but still blocked on the cordova-lib PR

shazron:
I would like this in 4.3.0. Hopefully we can resolve https://github.com/apache/cordova-lib/pull/453 for cordova-cli 6.4.0

fredgalvao:
Is there any reason why this PR doesn't also export `CFBundleName` as the result of the short name on config.xml? Follow me for a while:

According to [this page on the developer portal](https://developer.apple.com/library/content/qa/qa1892/_index.html) it seems like these two properties go along and, unless they are too big, should most probably have the same value.

Now, with the recent findings and issues related to renaming projects (which causes malformed pbxproj files and unsolvable string search&replace operations), it seems to me like we could simply leave the project structure and project name (not to be confused with app name) a generic static value, and deal with the `<name>` and `<short-name>` outputing them to `CFBundleDisplayName` and `CFBundleName` respectively.

That's what I did when I evolved my cordova project into a whitelabel, so that I could generate different apps with the same code base. However, in my case I needed to use a string replacement on `build.gradle` to set the `applicationId` on android (the app name works fine, it's updated in the string.xml correctly and my main activity class doesn't need to be renamed). There's an [issue](https://issues.apache.org/jira/browse/CB-11374) for the `applicationId` thing too, btw. On iOS, I also had to use [cordova-plugin-app-name](https://github.com/blakgeek/cordova-plugin-app-name/blob/master/plugin.xml) to deal with the "renaming breaks the project" issue (leaving the `<name>` untouched, otherwise I'd be screwed).

## Summary of my _opinion_

To evolve into a more whitelabel-friendly structure with respect to naming apps:

- Evolve `<name>` and `<short-name>` handling to affect **ONLY** {`strings.xml`@`app_name`} on android, and {`CFBundleName` + `CFBundleDisplayName`} on iOS, leaving the following items with a generic static name [activity class name on android, xcode project files (plist, pbxproj)].

Please @shazron and @dpogue tell me what you think of all of this.

dpogue:
@shazron This can go in to 4.4.1 now that cordova-common changes are merged

fredgalvao:
Should my previous comment go into an issue dedicated to discussion? There was no feedback on it, and it seems like there is a consensus into doing the simple version, which imo doesn't actually solve the root problem (and could solve many other problems if overhauled).

@shazron @dpogue  I'm honestly interested in at least knowing your reasoning and opinion regarding my previously stated PoV on the topic. Who knows, maybe I'm even using an overly complex solution on my side.

dpogue:
@fredgalvao Personally, I think your idea is the right way to go in terms of project structure, and should definitely be brought up for discussion on the dev list (because it probably has implications beyond just the display name of the app and because we'd probably want all supported platforms to agree).  The scope for this particular PR is just to update the name in the Info.plist.

shazron:
@dpogue looks like a Simulator timeout. Can you close and re-open the PR to trigger the CI again
@fredgalvao ditto what @dpogue said

shazron:
@dpogue bah, same timeout error. Please do the close and re-open PR dance again

dpogue:
@shazron All good now :)

PR #229, opened 2016-06-15
Incorrect reference to UIWebView

eladmoshe:
In `CDVUIWebViewNavigationDelegate.m` there is a call to `self.enginePlugin.webView` which returns nil.
This bug breaks the Cordova mechanism for posting notifications to the default notification center, so the application cannot listen to notifications.
The webview is already part of the functions signature, so we can simply use `theWebView` instead of `self.enginePlugin.webView`.


eladmoshe:
The CI failed, but I'm pretty sure it has nothing to do with this change.

shazron:
I agree about the CI failure.
Check out this commit: https://github.com/apache/cordova-ios/commit/2e41e221d50dd00b832879cd3b4ced372f8be428

I think it fixes this problem -- although the description seems unrelated, if you go to the JIRA issue there are more details: https://issues.apache.org/jira/browse/CB-11375

shazron:
With the release of cordova-ios-4.2.0 today which includes the commit above, I believe this already fixes your issue. Please close this PR, thanks.

eladmoshe:
Thanks.

PR #230, opened 2016-06-23
CB-11475 Ignore unsupported 60x60 icon

vladimir-kotikov:
The 60x60 icon is not used in `AppIcon` catalog and has been removed in 68af7ea. However if you have `<icon width="60" height="60" ...>` in `config.xml` it'll be added to catalog as `icon-60.png`. This will result in Xcode warning `The app icon set "AppIcon" has an unassigned child`

See also [CB-11475](https://issues.apache.org/jira/browse/CB-11475)


vladimir-kotikov:
@shazron, could you please take a look?

shazron:
LGTM 👍

PR #231, opened 2016-06-24
Fix Macos Japanese Filename Problem. (Change NFD projectName to NFC t…

knight9999:
This request fixes the NFC / NFD problem. (Filename String for Mac)
In Mac, The filename is treated as NFD.
However, the target name in xcode is treated as NFC.
Therefore, cordova build fails when we use the specific Japanese name for project title.
(For example, 

    cordova create サンプル com.example.sample サンプル
    cordova platform add ios
    cordova build

==> Fail.
Because the サンプル in NFC is different from and サンプル in NFD.)

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -146,7 +146,7 @@ function getXcodeArgs(projectName, projectPath, configuration, isDevice) {
         xcodebuildArgs = [
             '-xcconfig', path.join(__dirname, '..', 'build-' + configuration.toLowerCase() + '.xcconfig'),
             '-project', projectName + '.xcodeproj',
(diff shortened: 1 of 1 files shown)


masahirotanaka:
@shazron @stevengill Let me know if you need more explanation. It's the locale issue, but this causes some apps having specific commonly used Japanese characters to fail building.

shazron:
I believe this was only a problem with 4.1.1 and below. Please try with 4.2.0, just released today. 
I arrived at a similar fix: https://issues.apache.org/jira/browse/CB-11401 but it was already fixed in 4.2.0 using other means.

knight9999:
As you say, 4.2.0 works successfully. I only tested with 4.1.1.
Sorry, I withdraw this request. Thanks!

PR #232, opened 2016-06-28
CB-11270 - Handle JavaScript onclick handler navigation

shazron:


CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m
```diff
@@ -95,23 +95,33 @@ - (CDVIntentAndNavigationFilterValue) filterUrl:(NSURL*)url
...
     switch (filterValue) {
         case CDVIntentAndNavigationFilterValueNavigationAllowed:
             return YES;
         case CDVIntentAndNavigationFilterValueIntentAllowed:
-            // only allow-intent if it's a UIWebViewNavigationTypeLinkClicked (anchor tag)
-            if (UIWebViewNavigationTypeLinkClicked == navigationType) {
+            // only allow-intent if it's a UIWebViewNavigationTypeLinkClicked (anchor tag) OR
+            // it's a UIWebViewNavigationTypeOther, and it's an internal link
+            if ([[self class] shouldOpenURLRequest:request navigationType:navigationType]){
                 [[UIApplication sharedApplication] openURL:url];
```
tony--: Currently if we make it in to openURL, we'll still fall through to return NO.
Since we handled it (by sending it to openURL), should we return YES from inside the if clause?

tests/CordovaLibTests/CDVWhitelistTests.m
```diff
@@ -295,6 +295,37 @@ - (void)testAllowIntentsAndNavigations
...
+    
+    // mainDocumentURL and URL are the same in the NSURLRequest
+    // Only UIWebViewNavigationTypeLinkClicked and UIWebViewNavigationTypeOther should return YES
+    XCTAssertTrue([CDVIntentAndNavigationFilter shouldOpenURLRequest:telRequest navigationType:UIWebViewNavigationTypeLinkClicked]);
+    XCTAssertTrue([CDVIntentAndNavigationFilter shouldOpenURLRequest:telRequest navigationType:UIWebViewNavigationTypeOther]);
+    XCTAssertFalse([CDVIntentAndNavigationFilter shouldOpenURLRequest:telRequest navigationType:UIWebViewNavigationTypeReload]);
+    XCTAssertFalse([CDVIntentAndNavigationFilter shouldOpenURLRequest:telRequest navigationType:UIWebViewNavigationTypeBackForward]);
+    XCTAssertFalse([CDVIntentAndNavigationFilter shouldOpenURLRequest:telRequest navigationType:UIWebViewNavigationTypeFormSubmitted]);
+    XCTAssertFalse([CDVIntentAndNavigationFilter shouldOpenURLRequest:telRequest navigationType:UIWebViewNavigationTypeFormResubmitted]);
+
```
tony--: Would it make sense to test the intent-allowed case?  This currently fails - see comment in comments section.
```    
    // Only CDVIntentAndNavigationFilterValueIntentAllowed and CDVIntentAndNavigationFilterValueNavigationAllowed should return YES
    XCTAssertTrue([CDVIntentAndNavigationFilter shouldOverrideLoadWithRequest:telRequest navigationType:UIWebViewNavigationTypeOther filterValue:CDVIntentAndNavigationFilterValueNavigationAllowed]);
    XCTAssertTrue([CDVIntentAndNavigationFilter shouldOverrideLoadWithRequest:telRequest navigationType:UIWebViewNavigationTypeOther filterValue:CDVIntentAndNavigationFilterValueIntentAllowed]);
    XCTAssertFalse([CDVIntentAndNavigationFilter shouldOverrideLoadWithRequest:telRequest navigationType:UIWebViewNavigationTypeOther filterValue:CDVIntentAndNavigationFilterValueNoneAllowed]);

```

CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m
```diff
@@ -95,23 +95,33 @@ - (CDVIntentAndNavigationFilterValue) filterUrl:(NSURL*)url
...
             }
-            // consume the request (i.e. no error) if it wasn't a UIWebViewNavigationTypeLinkClicked
+
```
tony--: Currently if we make it in to openURL, we'll still fall through to return NO.
Since we handled it (by sending it to openURL), should we return YES from inside the if clause?

tony--:
I ran the tests - they passed.

I wrote a manual test with an index.html that has a javascript tel: link and an iframe with the same javascript tel: link.  The index.html tel:link works, the iframe tel:link does not work - this is what we want, I think.

Please see the inline comments on the "Files changed" tab.

CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m
shazron: returning NO here signals to the WebView that it should not load the URL inside itself, so it is correct in this context since we open the url externally.

tests/CordovaLibTests/CDVWhitelistTests.m
shazron: See my comment on that note. The operative test here is for "shouldOverrideLoadWithRequest" -- if the intent is allowed, it should return NO, as in it shouldn't override the load with the request (since it is opened externally).

CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m
tony--: Got it (especially now that I reread [the doc for the outer method](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIWebViewDelegate_Protocol/#//apple_ref/occ/intfm/UIWebViewDelegate/webView:shouldStartLoadWithRequest:navigationType:))!  Thanks for explaining.

tests/CordovaLibTests/CDVWhitelistTests.m
tony--: Yep, I see now that this test wouldn't make sense.

tony--:
LGTM!

PR #233, opened 2016-07-01
CB-9371, CB-9703: Fix how prepare handles orientation on ios

cjpearson:
Relates to https://github.com/apache/cordova-lib/pull/260

This changes the behavior of the default orientation preference to match the default orientation settings for new native iOS projects.

It also no longer modifies the orientation settings if the user does not specify an orientation preference.


shazron:
Confirmed that Xcode uses the new defaults. Pulling this in and testing.

PR #236, opened 2016-07-19
Adding WebViewEngine plugin that uses WKWebView

bhariharan:
This pull request adds a full implementation of the ```Cordova``` engine web view using ```WKWebView```. Currently, there's only a plugin that uses ```UIWebView```.

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWKWebViewEngine/CDVWKWebViewDelegate.h
+++ b/CordovaLib/Classes/Private/Plugins/CDVWKWebViewEngine/CDVWKWebViewDelegate.h
@@ -1,0 +1,43 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ 
+ Created by Bharath Hariharan on 7/15/16.
+ */
+
+#import <WebKit/WebKit.h>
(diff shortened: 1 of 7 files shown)


bhariharan:
@shazron Could you review please? I'm using this in my app and figured I'd contribute back.

dpogue:
The WKWebView engine is provided as a plugin: https://github.com/apache/cordova-plugin-wkwebview-engine

bhariharan:
@dpogue Why does that plugin have an ```iOS 9``` minimum? All the required ```WKWebView``` APIs are available in ```iOS 8```.

dpogue:
@bhariharan WKWebView on iOS 8 can't load pages from the filesystem (via `file:///` URLs), which makes it largely useless for Cordova apps

bhariharan:
@dpogue Oh right. I was aware of that issue. Thanks! Are there any plans to bring the plugin in and make ```WKWebView``` the default going forward?

dpogue:
Probably not while there are [so many known issues](https://issues.apache.org/jira/browse/CB-11469?jql=project%20%3D%20CB%20AND%20labels%20%3D%20wkwebview-known-issues) with WKWebView

bhariharan:
Thanks @dpogue! I'll close this then.

bhariharan:
@dpogue Shouldn't [this](https://github.com/apache/cordova-plugin-wkwebview-engine/blob/master/src/ios/CDVWKWebViewEngine.m#L356) be ```decidePolicyForNavigationAction``` selector for ```WKNavigationDelegate```, instead of what it is there? I'm having some trouble getting my app working. It's looping on ```gap://ready``` and never seems to get out that loop. Any ideas?

shazron:
@bhariharan what cordova-ios version are you using? cordova-ios-4.1.0 fixes that loop problem. no, what you [linked to](https://github.com/apache/cordova-plugin-wkwebview-engine/blob/1b31d4d9fafc3b06a8a6f8d958737d2312b415ab/src/ios/CDVWKWebViewEngine.m#L356) is correct -- it pushes this decision off to an internal plugin that parses and handles `<allow-intent>` and `<allow-navigation>` tags in config.xml

bhariharan:
@shazron I figured out the looping problem. In your default implementation of `WebViewEngine`, you now call delegates of the navigation delegate only if the view controller itself doesn't handle delegate callbacks. This is problematic in my case, since my subclass of `CDVViewController` handles delegate callbacks, but I want to call the `super` after doing my part, which in this case, is the navigation delegate part of the plugin.

shazron:
@bhariharan if you could file a new issue at issues.cordova.io we'll try to fix this

PR #238, opened 2016-07-22
CB-11535 CB-10361 ios: fix bug with remove frameworks

matrosovN:


shazron:
hi @matrosovN can you add "CB-10361" to the PR title prefix as well? So it will now be:
`CB-11535 CB-10361 ios: fix bug with remove frameworks`

This is for tracking in JIRA.

matrosovN:
@shazron, updated.

PR #241, opened 2016-08-15
CB-11705 - Adding CordovaDefaultWebViewEngine configuration

hnilsen:
Adding CordovaDefaultWebViewEngine configuration option to be able to use a different WebView as default and/or fallback.

This way you can extract the UIWebViewEngine plugin from the private space. You can also force the WKWebView to be default, or some other webview if needed.

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -148,13 +148,13 @@
         }
         path = absolutePath;
     }
-    
+
     // Assert file exists
     if (![[NSFileManager defaultManager] fileExistsAtPath:path]) {
         NSAssert(NO, @"ERROR: %@ does not exist. Please run cordova-ios/bin/cordova_plist_to_config_xml path/to/project.", path);
         return nil;
     }
-    
+
     return path;
 }
 
@@ -162,7 +162,7 @@
 {
     // read from config.xml in the app bundle
     NSString* path = [self configFilePath];
-    
+
     NSURL* url = [NSURL fileURLWithPath:path];
 
     self.configParser = [[NSXMLParser alloc] initWithContentsOfURL:url];
@@ -273,7 +273,7 @@
 - (void)viewDidLoad
 {
     [super viewDidLoad];
-    
+
     // Load settings
     [self loadSettings];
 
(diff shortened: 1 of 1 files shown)


hnilsen:
This is from a discussion in #cordova-ios on Slack. cc @shazron

hnilsen:
JIRA task: https://issues.apache.org/jira/browse/CB-11705

shazron:
I think you misunderstand that part of the code. You can already swap the default webview engine. It will use the setting "CordovaWebViewEngine" as you can see below the code you are changing. 

Also, see https://github.com/apache/cordova-plugin-wkwebview-engine
In its plugin.xml, it installs itself as:
https://github.com/apache/cordova-plugin-wkwebview-engine/blob/42c847b2680a80e52ffd3617119ddd6634751453/plugin.xml#L48

The value is the class name.

hnilsen:
That doesn't swap out the private CDVUIWebViewEngine used by iOS8, does it? That's why I'm adding the option to extract and use my own version of CDVUIWebViewEngine. This is what I'm able to do now:

```
<preference name="CordovaDefaultWebViewEngine" value="CDVSpvUIWebViewEngine" />
<preference name="CordovaWebViewEngine" value="CDVSpvWKWebViewEngine" />
```

```
    <feature name="CDVSpvUIWebViewEngine">
        <param name="ios-package" value="CDVSpvUIWebViewEngine" />
    </feature>

    <feature name="CDVSpvWKWebViewEngine">
        <param name="ios-package" value="CDVSpvWKWebViewEngine" />
    </feature>
```

This enables me to create my own UIWebView plugin for iOS8.

shazron:
I get what you are saying now, you want the "fallback" webview-engine to be your plugin. I'll review this PR soon with that in mind.

hnilsen:
Yes, the current situation is that I can supply a plugin for my own WKWebView, but not for the UIWebView, which is a private plugin.

I'd like to be able to control them both. My PR is a suggestion on how that can be accomplished.

shazron:
Tested locally (passes - Travis CI had a hiccup with the simulator, above) and did a code review, looks good to me.

Can you create a new docs PR for the new setting?
It should be in the same location as `CordovaWebViewEngine`:
http://cordova.apache.org/docs/en/latest/config_ref/index.html
The file location is:
https://github.com/apache/cordova-docs/blob/master/www/docs/en/dev/config_ref/index.md 
(just click on the Edit button at that GH location to create a PR)

hnilsen:
Sure thing - I'll edit the docs accordingly.

PR #244, opened 2016-08-26
CB-11771 Deep symlink directories to target project instead of linking the directory itself

mbektchiev:
When installing a plugin with custom library using the --link option the whole directory is symlinked and temporary
files leak into the original plugin directory on build. This leads to broken builds if the same plugin is linked in
2 projects targeting different Cordova versions.

fixes [CB-11771 Installing plugin with symlink option pollutes original directory with intermediate files](https://issues.apache.org/jira/browse/CB-11771)

### DIFF
--- a/bin/templates/scripts/cordova/lib/plugman/pluginHandlers.js
+++ b/bin/templates/scripts/cordova/lib/plugman/pluginHandlers.js
@@ -53,7 +53,8 @@ var handlers = {
             if (!fs.existsSync(srcFile)) throw new CordovaError('Cannot find resource file "' + srcFile + '" for plugin ' + plugin.id + ' in iOS platform');
             if (fs.existsSync(destFile)) throw new CordovaError('File already exists at detination "' + destFile + '" for resource file specified by plugin ' + plugin.id + ' in iOS platform');
             project.xcode.addResourceFile(path.join('Resources', path.basename(src)));
-            shell.cp('-R', srcFile, project.resources_dir);
+            var link = !!(options && options.link);
+            copyFile(plugin.dir, src, project.projectDir, destFile, link);
         },
         uninstall:function(obj, plugin, project, options) {
             var src = obj.src,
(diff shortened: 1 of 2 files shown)


shazron:
Thanks! LGTM. What "temporary files" leak, do you have more info on that?

mbektchiev:
I'm not sure if the same behavior can be reproduced with `cordova-ios`. But since the code is exactly the same as the one in `cordova-android` I thought it might be good idea to keep it consistent and fix it here as well. 

Basically the scenario is to install a plugin which contains a custom Android library code with `--link`. All intermediate build files are created inside a `build` sub-directory and they leak into the plugin's directory.

mbektchiev:
See https://github.com/Wizcorp/phonegap-facebook-plugin/tree/master/platforms/android/FacebookLib and https://github.com/Telerik-Verified-Plugins/Telerik-AppFeedback/tree/master/src/android/FeedbackLib as two examples of such libraries.

shazron:
I tested this on the `cordova-plugin-console` plugin and `phonegap-facebook-plugin`. The results before and after applying the patch are the same.

In the `PROJECTNAME/plugins` folder, the folder for the plugin itself is linked, and not its contents individually (same as before the patch).

mbektchiev:
You're right, it seems that currently the `--link` option is disregarded in the `resource-file` and `framework` handlers. I've changed them and now you can see the difference in the linking style with the `phonegap-facebook-plugin`.

shazron:
Merged into master: https://github.com/apache/cordova-ios/commit/c6587cffe7afded445500c2f30addfdf65290797  -- you can close this PR. Thanks!

mbektchiev:
Thanks, @shazron 
I noticed that you've marked the Jira Issue as resolved. In order to completely fix it we should also merge https://github.com/apache/cordova-android/pull/326 and https://github.com/apache/cordova-lib/pull/485.

shazron:
Oops sorry, I'll re-open the issue

PR #245, opened 2016-08-30
CB-11731 Re-read ios.json on every prepare

vladimir-kotikov:
This fixes empty 'ios.json' file after calling a sequence of `'prepare' -> 'addPlugin' -> 'prepare'` within one node process. A typical scenario when this might happen is installing ios platform into project with other platforms (and plugins) already installed. Then logic in ios' `prepare` caches `PlatformJson` instance and overrides all changes to ios.json, made during plugin installation in second `prepare`

Related JIRA: [CB-11731](https://issues.apache.org/jira/browse/CB-11731)
This also might be a real reason of [CB-11589](https://issues.apache.org/jira/browse/CB-11589) and https://github.com/apache/cordova-lib/pull/464

@stevengill, @dpogue, please take a look


dpogue:
:+1: This is definitely the root cause of CB-11589

stevengill:
LGTM

PR #246, opened 2016-08-31
CB-11712 - <name> changes in config.xml does a 'search and replace all' for occurrences of the old name with the new name in the pbxproj

shazron:
The error message looks like this:
```
Error: The product name change (<name> tag) in config.xml is not supported dynamically.
To change your product name, you have to remove, then add your ios platform again.
Make sure you save your plugins beforehand using `cordova plugin save`.
       	cordova plugin save
       	cordova platform rm ios
       	cordova platform add ios
```

Not sure if I'm supposed to be hardcoding `cordova` for downstream distributions.

### DIFF
--- a/bin/templates/scripts/cordova/lib/prepare.js
+++ b/bin/templates/scripts/cordova/lib/prepare.js
@@ -32,7 +32,6 @@ var CordovaError = require('cordova-common').CordovaError;
 var PlatformJson = require('cordova-common').PlatformJson;
 var PlatformMunger = require('cordova-common').ConfigChanges.PlatformMunger;
 var PluginInfoProvider = require('cordova-common').PluginInfoProvider;
-var projectFile = require('./projectFile');
 var FileUpdater = require('cordova-common').FileUpdater;
 
 /*jshint sub:true*/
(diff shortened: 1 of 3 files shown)


shazron:
I am fixing the tests.

shazron:
Tests pass, squashing the commits.

bromeostasis:
@shazron just updated cordova and started running into this error. We use cordova to apply one codebase to several different white labeled apps with different data and styles. 

Until now we've been dynamically changing the name by removing/re-adding the platform we're working on and then using different sets of config.xml/stylesheet/js config. Is there any way to continue this type of behavior, or will we have to create separate cordova projects for each of our White Labels? This would be a serious pain considering we have over 20!!

Thanks!

jcesarmobile:
What is your error exactly?

bromeostasis:
@jcesarmobile well I'm running into the expected error:

```
Error: The product name change (<name> tag) in config.xml is not supported dynamically.
To change your product name, you have to remove, then add your ios platform again.
Make sure you save your plugins beforehand using `cordova plugin save`.
        cordova plugin save
        cordova platform rm ios
        cordova platform add ios
```

which I understand the purpose of. We definitely were switching out the \<name\> tag to change the name of the project. I'm just wondering if there's any way to switch the name of the project dynamically with the latest version of cordova, since that's one of our main reasons for using it.

jcesarmobile:
But you said that you were removing and re adding the platforms, so it shouldn't complain in that case.
This might be related to Cordova CLI 7 also using the package.json for storing the platforms. Can you create a new issue on http://issues.cordova.io/?

bromeostasis:
@jcesarmobile sure, I'll confirm the specific steps and then post a new issue. Any thoughts on a workaround if this a problem?

jcesarmobile:
only downgrade to previous CLI

bromeostasis:
@jcesarmobile I found a few bugs on our end that prevented things from working right; looks like the upgrade just exposed those, so no problems here!

Thanks!

PR #250, opened 2016-09-01
CB-9762 (iOS) Add launch storyboard support

kerrishotts:
- [X] Signed ICLA
- [X] Jira Issue: [CB-9762](https://issues.apache.org/jira/browse/CB-9762)
- [X] Tests:
    - [X] Automated (npm test; all passing)  Note: no automated tests for this functionality, though
    - [X] Manual (with and without launch storyboard images):
        - iPhone 6s (iOS 10b8)
        - iPad Pro (iOS 10b8)

Note: The splashscreen plugin will require updating to use the new images; but I think that should be a separate (but related) issue. I don't think it will be difficult to add though (crossing my fingers)!

Note: Documentation will need to be updated to reflect the changes in this PR

Note: No default splash images are provided; they would be overwritten anyway upon
adding the platform. I've added the images I used below:

default 2x universal anyany
default 2x universal comany
default 2x universal comcom
default 3x universal anycom
default 3x universal comany


shazron:
re: your last commit -> https://github.com/apache/cordova-ios/pull/258

kerrishotts:
@shazron: Thanks! Removed my last commit and waiting on CI results -- should all be green now with #258. :-)

PR #257, opened 2016-09-16
CB-11860 - Update packaging strategy for Xcode 8

dpogue:
Background info on the mailing list: https://lists.apache.org/thread.html/85a715681471c774821d76a90ec288943eca1a4bb0a267ba299d2eb3@%3Cdev.cordova.apache.org%3E

Mostly just opened for review now, and getting any feedback or comments. Tests will probably fail, and will be updated after review.

A note that using xcodebuild for exporting the archive requires using the system installation of Ruby, so if you're an RVM user then you'll need to `rvm use system` first. I'm looking to see if there's a way we can work around this.

/cc @shazron

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -116,8 +116,34 @@ module.exports.run = function (buildOpts) {
           return spawn('xcodebuild', xcodearchiveArgs, projectPath);
         }
 
+        function unpackIPA() {
+            var ipafile = path.join(buildOutputDir, projectName + '.ipa');
+
+            // unpack the existing platform/ios/build/device/appname.ipa (zipfile), will create a Payload folder 
+            return spawn('unzip', [ '-o', '-qq', ipafile ], buildOutputDir);
+        }
+
+        function moveApp() {
+            var appFileInflated = path.join(buildOutputDir, 'Payload', projectName + '.app');
+            var appFile = path.join(buildOutputDir, projectName + '.app');
+            var payloadFolder = path.join(buildOutputDir, 'Payload');
+
+            // delete the existing platform/ios/build/device/appname.app 
+            return spawn('rm', [ '-rf', appFile ], buildOutputDir)
+                .then(function() {
+                    // move the platform/ios/build/device/Payload/appname.app to parent 
+                    return spawn('mv', [ '-f', appFileInflated, buildOutputDir ], buildOutputDir);
+                })
+                .then(function() {
+                    // delete the platform/ios/build/device/Payload folder
+                    return spawn('rm', [ '-rf', payloadFolder ], buildOutputDir);
+                });
+        }
+
         return Q.nfcall(fs.writeFile, exportOptionsPath, exportOptionsPlist, 'utf-8')
-                .then(packageArchive);
+                .then(packageArchive)
+                .then(unpackIPA)
+                .then(moveApp);
     });
 };
 
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -24,7 +24,8 @@ var Q     = require('q'),
     shell = require('shelljs'),
     spawn = require('./spawn'),
     check_reqs = require('./check_reqs'),
-    fs = require('fs');
+    fs = require('fs'),
+    plist = require('plist');
 
 var events = require('cordova-common').events;
 
@@ -54,7 +55,7 @@ module.exports.run = function (buildOpts) {
             var buildType = buildOpts.release ? 'release' : 'debug';
             var config = buildConfig.ios[buildType];
             if(config) {
-                ['codeSignIdentity', 'codeSignResourceRules', 'provisioningProfile', 'developmentTeam'].forEach(
+                ['codeSignIdentity', 'codeSignResourceRules', 'provisioningProfile', 'developmentTeam', 'packageType'].forEach(
                     function(key) {
                         buildOpts[key] = buildOpts[key] || config[key];
                     });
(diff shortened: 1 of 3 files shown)


chancezeus:
Just a note, I used this variant to successfully build a couple of IPA's for iOS. There is one downside (at the moment) I could find, it breaks the run option, so at the moment I have to manually run ios-deploy to debug the application.

dpogue:
Running on a device, or running on the simulator? In theory, running is supposed to be a separate step from packaging, but I'll try this out today and see what I find.

chancezeus:
Both, it complains it cannot find the [/_PACKAGE_NAME/_].app bundle, 
which is correct, since only the xcarchive and ipa are created. You can 
use ios-deploy (which I do manually for now) to install an IPA, but 
ios-deploy itself has an error where it cannot find 
"[/_PACKAGE_NAME_/].ipa/Info.plist" if you try to use an ipa as a bundle 
instead of a ".app" folder

shazron:
The easiest way is to create a post-archive step to decompress the .ipa (which is just a zip) to extract the .app bundle, in the same location.

shazron:
Regarding the .ipa Info.plist -- ios-deploy absolutely needs the bundle identifier of the app to install the app using lldb.

ekrapfl:
Out of curiosity, is there any reason to do the archive step at all when building in order to run?  Perhaps there could be an option to skip archiving, which would be called when doing a run.

dpogue:
The only reasons to go through the archive step is to ensure that the build is signed with the right certificates. Building and deploying the .app directly would be signed with a development certificate instead of a distribution one.

shazron:
Re: unzipping the .ipa. node.js has built-in support through the zlib module: https://nodejs.org/docs/latest-v0.10.x/api/zlib.html

ekrapfl:
dpogue: I see.  I was only thinking of the case of trying to run a debug build on a device.  I that case, I think skipping the archiving step would potentially be fine.  However, if you were attempting to do a run --release --device, then it would need to do the archive step for proper signing.

shazron: That approach seems like it would work great as long as the .ipa is also kept in its original state as well.

shazron:
@dpogue Just tested this. We need to add a warning if users are not using the system ruby, like the  pre-reqs check for xcodebuild (it wouldn't work like you said without the system ruby). I didn't even need to use the `--developmentTeam` flag for it to work. `cordova build --device` worked like it did before.

@chancezeus in `platforms\ios\device` there is already a symlink to the .app bundle. Unfortunately `ios-deploy` doesn't traverse the symlink for some reason, gives an error : `Can't access app path '/Users/shaz/Desktop/f/platforms/ios/build/device/my.app' : No such file or directory` . I'll have to make a patch release to ios-deploy for this, and update the minimum requirements to cordova-ios that has to go with this pull request.

dpogue:
Yeah, I was hoping to find a convenient way to override Ruby gem paths with environment variables or something, but for now it's probably worth just printing a warning for now.

I'm curious if the `--developmentTeam` is needed if you had multiple teams, but it's nice that for most people it will just work without issues.

shazron:
I filed https://github.com/phonegap/ios-deploy/issues/258 to solve the symlink issue.

shazron:
Sorry spoke too soon, turns out you still need `--developmentTeam` since I didn't realize I was using Xcode 7 command line tools :/ 

Also, the .app symlink in `platforms/ios/build/device` turns out to be a dud, it doesn't point to a file that exists, strange -- so we will need to rely on extracting the .ipa

shazron:
I'm going to work on the unzip part and send a patch to @dpogue -- this should be the last PR before we lock it down for cordova-ios 4.3.0 release

shazron:
@dpogue Attached the patch for your xcodebuild branch. I tried to send a PR to your branch but Github's interface just didn't allow me to.
 
[CB-11860.xcodebuild.post-archive.patch.zip](https://github.com/apache/cordova-ios/files/503730/CB-11860.xcodebuild.post-archive.patch.zip)

dpogue:
Thanks @shazron! I've pushed that

shazron:
One flaw with this export archive business -- the .app in the .ipa that is packaged does *not* contain the `www` folder thus the cordova app will crash when run using `cordova run --device`. I'm not sure why it wasn't packaged.

dpogue:
The www files aren't anywhere in the .ipa? I've been making .ipa builds with an enterprise cert for a few weeks now and uploading them for our QA team, and nobody has reported any issues 😕

shazron:
@dpogue not for a development build using a dev cert, at least for me

shazron:
If I add the `www` folder into `Copy Bundle Resources` it is fine:
screen shot 2016-09-30 at 5 14 39 pm

shazron:
Solved it. I had to add `--packageType=development`. If the `packageType` flag was omitted, there was no `www` folder. This is strange since the exportOptions method should default to `development` according to the docs.

bin/templates/scripts/cordova/lib/build.js
```diff
@@ -88,25 +89,61 @@ module.exports.run = function (buildOpts) {
...
         return spawn('xcodebuild', xcodebuildArgs, projectPath);
     }).then(function () {
         if (!buildOpts.device || buildOpts.noSign) {
             return;
         }
+
+        var exportOptions = {'compileBitcode': false};
+
+        if (buildOpts.packageType) {
+            exportOptions.method = buildOpts.packageType;
```
dpogue: @shazron We could default to "development" here...

shazron: Yeah we should. I just tested it, and it works with that default. Must be an xcodebuild bug.

dpogue:
Updated, thanks!

shazron:
Hmm. On multiple runs, sometimes the `www` folder is not there, so I'm not sure what's going on -- doesn't seem to be reliable. I'll dig in later.

shazron:
Looks like if the `build/device/test.app` already exists (not a symlink) and we go through the process, it is always replaced by one without a `www`. If the .app package is deleted, and the process re-runs, it creates it properly. Will try to tweak the script.

shazron:
The only way I can get a consistent .ipa that has the `www` folder in it, is if we wipe the `platforms/ios/build/device` folder, will try that now.

shazron:
@dpogue Ok, this should do it. Patch attached.

[CB-11860.pre-build.clean.patch.zip](https://github.com/apache/cordova-ios/files/507552/CB-11860.pre-build.clean.patch.zip)

dpogue:
@shazron Fixed an issue with the `rm` that was causing ENOENT on initial builds. Tests should pass now.

shazron:
Thanks @dpogue, I'll do final tests and pull this in.

PR #260, opened 2016-09-28
CB-10078 fixed by refreshing cached userAgent on version bump

kerrishotts:
Adds app version to the list of items that invalidates the cached user agent.


shazron:
LGTM 👍

PR #261, opened 2016-10-05
Print a warning if system ruby is not in use

dpogue:
This should give people some indication as to why xcodebuild export sometimes fails with errors that make no sense.

/cc @shazron


shazron:
Strange Travis CI failure. I can't see what's wrong. Can you close and re-open this PR to trigger a new build, perhaps its a one-off issue.

shazron:
Pulling this in.

PR #269, opened 2016-10-12
CB-12002 - Support LSApplicationQueriesSchemes

dpogue:
⚠️ Depends on https://github.com/apache/cordova-lib/pull/502
### Platforms affected

iOS
### What does this PR do?

Adds support for parsing `<allow-intent>` tags in config.xml and creating a list of allowed schemes to be written as `LSApplicationQueriesSchemes` in the Info.plist.

For more info about LSApplicationQueriesSchemes, see https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/LaunchServicesKeys.html#//apple_ref/doc/uid/TP40009250-SW14

Added spec test.
### Checklist
- [x] Reported an issue in the JIRA database: [CB-12002](https://issues.apache.org/jira/browse/CB-12002)
- [x] Commit message follows the format.


shazron:
The dependent changes will be in the cordova-common 1.5.1 release, and will be integrated as soon as it is released.

shazron:
@dpogue Can you rebase? cordova-common 1.5.1 is in the repo now

jcesarmobile:
I commented this on the issue

> LSApplicationQueriesSchemes is only needed for canOpenURL, but we use openURL without querying if it can be opened first. That opened the apps if they were on the allow-intent (worked on iOS 9 and earlier, but didn't test on iOS 10)

shazron:
Yeah looks like openURL does not need the key: https://developer.apple.com/reference/uikit/uiapplication/1622952-canopenurl
(see last paragraph -- it has a typo, UIDefaultLaunchStoryboard should be LSApplicationQueriesSchemes

dpogue:
hmm, I think in our case we were trying to use canOpenURL to see if a partner app was able to launch and redirect to the store otherwise.  This also came up in the Slack channel recently, and was mentioned in some Github threads on notification plugins trying to launch mailto links.

shazron:
So if they use `canOpenURL` they need this, but if they just use `openURL`, like our whitelist system, we don't. Therefore I believe our whitelist system with `<allow-intent>` does not need these keys.

jcesarmobile:
We where using canOpenURL on inAppBrowser plugin, but I removed it to fix CB-11178 as we weren't really sending any response to the user if canOpenURL was false.

I don't think it's a good idea to add all the allow-intent schemes into the LSApplicationQueriesSchemes, as not all are needed to be there (as http or https), also, the PR doesn't handle the case of a duplicate scheme, as you can put multiple urls starting by http in allow intent tags and this will add http protocol multiple times.

As we are going to allow to write into the info.plist from the config.xml soon, and we don't use canOpenURL on cordova-ios, I don't think this should be merged.

shazron:
eslint errors.

shazron:
Please check the "Allow Edits from Maintainers" checkbox on this pull request.

dpogue:
It is checked already :\ I can try toggling it off and back on?

shazron:
oh? i don't see it, maybe I don't have karma -- weird, I thought we get those already.

shazron:
Oh I see "Add more commits by pushing to the allow-intent branch on dpogue/cordova-ios" maybe thats it

shazron:
Reviewing PRs for new cordova-ios release. @dpogue @jcesarmobile -- not sure what the status of this is.

dpogue:
I believe (but have not confirmed) that it is possible to use `<edit-config>` to do this now, so this PR would not be strictly necessary

jcesarmobile:
I confirm the edit-config tag allows to add the LSApplicationQueriesSchemes, so I'm -1 on merging this

PR #270, opened 2016-10-28
input file occurs error:[Creating an image format with an unknown type is an error]

puny-d:
I have found the input[type='file'] can't get image on ios 10 with cordova.

there is an error on Xcode: Creating an image format with an unknown type is an error.


jcesarmobile:
Hi.
You have created a Pull Request, can you close it?
This is not the way of reporting issues, you can report issues on http://issues.cordova.io/
I have seen that message too, but it looks like a warning, the image is selected anyway (at least on iOS 10.1), but the image is upside down.
Anyway, the input type file is managed by the webview itself, not by Cordova, we can't change it's behaviour

PR #276, opened 2016-11-10
CB-12130 - Launch storyboard images are not updated or cleaned

NielsLeenheer:
Launch storyboard images were not updated or cleaned in
`updateLaunchStoryboardImages()` or `cleanLaunchStoryboardImages()`
because the check for `Images.xcassets` in
`getLaunchStoryboardImagesDir()` failed.

The reason was that the `platformProjDir` and `projectRoot` were
joined, while both were relative to the root directory. So the
resulting directory was non-sensical. This changes makes
`platformProjDir` relative to `cordovaProject.root`, which matches the
same check in `updateSplashScreens()` and `updateIcons()`.

### DIFF
--- a/bin/templates/scripts/cordova/lib/prepare.js
+++ b/bin/templates/scripts/cordova/lib/prepare.js
@@ -685,7 +685,7 @@ function getLaunchStoryboardImagesDir(projectRoot, platformProjDir) {
  */
 function updateLaunchStoryboardImages(cordovaProject, locations) {
     var splashScreens = cordovaProject.projectConfig.getSplashScreens('ios');
-    var platformProjDir = locations.xcodeCordovaProj;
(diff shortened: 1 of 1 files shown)


kerrishotts:
Hi Niels -- great talking with you on Slack!

Just so everyone else knows, I'll be working on the feature next week. I had a massive influx of client work the last couple weeks, so haven't got to it yet, but next week looks clear. :-) Feature should be fully implemented in the next release.

Part of the fix I need to push is exactly what you've included here, so I'll pull your PR in next week so you can get credit.

Thanks for your help!

kerrishotts:
@NielsLeenheer -- would you mind squashing these commits down to a single commit? I'd like to pull this in.

NielsLeenheer:
Squashed!

kerrishotts:
Awesome; thanks! :-)

kerrishotts:
Pulled in; thanks for your contribution!

PR #277, opened 2016-11-11
CB-11243: Fix target-device and deployment-target preferences

cjpearson:
### Platforms affected
iOS

### What does this PR do?

Fixes target-device and deployment-target preferences. There are several bugs reporting this issue. [CB-11243](https://issues.apache.org/jira/browse/CB-11243) [CB-11399](https://issues.apache.org/jira/browse/CB-11399) [CB-11772](https://issues.apache.org/jira/browse/CB-11772) [CB-11732](https://issues.apache.org/jira/browse/CB-11732)

This change adds `IPHONEOS_DEPLOYMENT_TARGET` and `TARGETED_DEVICE_FAMILY` back to the project file. This is necessary for the preferences to work until [CB-11468](https://issues.apache.org/jira/browse/CB-11468) is completed.

### Checklist

### DIFF
--- a/bin/templates/project/__TEMP__.xcodeproj/project.pbxproj
+++ b/bin/templates/project/__TEMP__.xcodeproj/project.pbxproj
@@ -303,7 +303,9 @@
 				GCC_THUMB_SUPPORT = NO;
 				GCC_VERSION = "";
 				INFOPLIST_FILE = "__PROJECT_NAME__/__PROJECT_NAME__-Info.plist";
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				LD_RUNPATH_SEARCH_PATHS = "@executable_path/Frameworks";
+				TARGETED_DEVICE_FAMILY = "1,2";
 				PRODUCT_NAME = "__PROJECT_NAME__";
 			};
 			name = Debug;
@@ -323,7 +325,9 @@
 				GCC_THUMB_SUPPORT = NO;
 				GCC_VERSION = "";
 				INFOPLIST_FILE = "__PROJECT_NAME__/__PROJECT_NAME__-Info.plist";
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				LD_RUNPATH_SEARCH_PATHS = "@executable_path/Frameworks";
+				TARGETED_DEVICE_FAMILY = "1,2";
 				PRODUCT_NAME = "__PROJECT_NAME__";
 			};
 			name = Release;
--- a/bin/templates/scripts/cordova/build.xcconfig
+++ b/bin/templates/scripts/cordova/build.xcconfig
@@ -23,9 +23,7 @@
 //
 
 HEADER_SEARCH_PATHS = "$(TARGET_BUILD_DIR)/usr/local/lib/include" "$(OBJROOT)/UninstalledProducts/include" "$(OBJROOT)/UninstalledProducts/$(PLATFORM_NAME)/include" "$(BUILT_PRODUCTS_DIR)"
-IPHONEOS_DEPLOYMENT_TARGET = 8.0
 OTHER_LDFLAGS = -ObjC
-TARGETED_DEVICE_FAMILY = 1,2
 
 // Type of signing identity used for codesigning, resolves to first match of given type.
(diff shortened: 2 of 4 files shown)


shazron:
CB-11648 will be obsolete and may not be needed with https://issues.apache.org/jira/browse/CB-12127 already implemented. Example use: https://issues.apache.org/jira/browse/CB-11245?focusedCommentId=15652604&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15652604

shazron:
Reviewing this to understand where the bug

shazron:
Evaluated this. Turns out this is not the problem - see #278 failing test for this issue which looks like a failure in node-xcode updateBuildProperty. Note that the test project pbxproj already contains the two build settings, so its not a matter of the settings not being there.

cjpearson:
It seems like the xcconfig settings override the buildFlag flag. When I run this command on a starter cordova-ios 4.3.0 project, the MinimumOSVersion in the App's Info.plist is still iOS 8.0.

    cordova compile ios --buildFlag="IPHONEOS_DEPLOYMENT_TARGET=9.0"

I get a similar result with the `TARGETED_DEVICE_FAMILY` setting.

I added the settings back into the pbxproj to handle the case where the user doesn't use one of these config options. Otherwise the default would be a minimum version of iOS 10 and iPhone only.

cjpearson:
Regarding the tests, based on what I'm seeing they should not be failing. The settings are properly added to the project, but they are ignored in the final product.

shazron:
@cjpearson we need the unit tests to work first before this can be pulled in. I added those tests (which should have been there in the first place) because this PR didn't come with one but this is Cordova's fault not having one in the first place. I'll try to get that sorted.

Regarding xcconfig, yeah you are right xcconfig overrides the command line build settings (XCODE_XCCONFIG_FILE env var to a xcconfig file overrides them all, but that's not relevant here).

cjpearson:
It looks like the tests were failing because of the spy on writeFileSync. I've added a fix and pulled them into this PR. I also tried to add a test to verify the final .app is correct, but I couldn't find any examples to work with.

shazron:
Ah yes, thanks for catching that! The unit test should be adequate. I'll get this pulled in.

zetura:
Still having the issue with Cordova 6.5.0 (and cordova-ios 4.3.1).
I have `target-device='handset'` defined, I removed completely the ios platform and add it again, but I always get this code generated in my info.plist : 
```
<key>UISupportedInterfaceOrientations~ipad</key>
<array>
   <string>UIInterfaceOrientationPortrait</string
   <string>UIInterfaceOrientationPortraitUpsideDown</string>
</array>
```

My app is then refused by Apple : 
```
Performance - 2.4.1

  We noticed that your app did not run at iPhone resolution when reviewed on iPad running iOS 10.2.1.
  Please see attached screenshots for details.  

Next Steps  
Please revise your app to ensure it runs and displays properly at iPhone resolution on iPad.

  Resources  
For information on iOS device screen sizes and resolutions, please see the iOS Human Interface Guidelines as well as the Points versus Pixels in the View Programming Guide for iOS. 
```

Thanks

shazron:
@zetura that setting is actually correct, it modifies the TARGETED_DEVICE_FAMILY setting in the Xcode project and I verified that it works (sets it to 1 for Handset). Those keys in Info.plist are related to the 'Orientation' preference in config.xml. Something else must be going on for the rejection, not sure yet.

@kerrishotts do you have any insight? not sure if related to Launch Storyboards

kerrishotts:
@shazron @zetura I was able to run a couple apps (both cordova-ios@4.3.1) at iPhone resolutions on an iPad simulator just now, with and without launch storyboards. @zetura, it might help to see the screenshots from Apple to see what they saw, if can share them (redact as necessary).

zetura:
Hi, thanks for your quick feedback !
It seems that the issue is because the app is rendered at scale x2 on iPad (and you can also deactivate the zoom at the bottom right). And we thought it was related to this issue, because we just don't want the app to be available for iPad, at iPhone resolution or not. So we thought it was because we had the iPad orientation in the info.plist file.
Maybe it has nothing to do with Cordova (sorry if it's the case).
You can see the iPad screenshot here : 
screen-shot-2017-03-02-at-3 58 31-pm

PR #279, opened 2016-11-16
CB-12084: Update project build settings & plist

kerrishotts:
### Platforms affected

 - iOS

### What does this PR do?

Addresses the bug discovered in CB-12084 where cleaning would corrupt the Xcode project. Also addresses the particular use case where the user was eliminating legacy launch images from the Xcode project -- this code will now take care of that automatically.

Here's the logic:

- If the user only specifies images for the storyboard, then the Xcode project should not attempt to use an asset catalog for launch images as well. This PR will remove that particular build property from the project (`ASSETCATALOG_COMPILER_LAUNCHIMAGE_NAME`)
- If the user specifies both storyboard images and legacy launch images, then nothing changes; build settings will continue to have `ASSETCATALOG_COMPILER_LAUNCHIMAGE_NAME` set to `LaunchImage`.
- If the user specifies ONLY legacy launch images, `ASSETCATALOG_COMPILER_LAUNCHIMAGE_NAME` is set to `LaunchImage`; this ensures that if the user has gone back to legacy images after having tried the storyboard images that the Xcode project still works as expected.

- Manual testing in each of the above scenarios
- `npm test` completed successfully with no errors (but it would as no tests currently trigger any of the above behavior)
- Travis CI is green.

### Checklist
    - Note that automated test coverage is coming in a separate PR since that test coverage will test the feature as a whole (not just this change).

### DIFF
--- a/bin/templates/scripts/cordova/lib/prepare.js
+++ b/bin/templates/scripts/cordova/lib/prepare.js
@@ -34,6 +34,14 @@ var PlatformMunger = require('cordova-common').ConfigChanges.PlatformMunger;
 var PluginInfoProvider = require('cordova-common').PluginInfoProvider;
 var FileUpdater = require('cordova-common').FileUpdater;
 
+// launch storyboard and related constants
+var LAUNCHIMAGE_BUILD_SETTING  = 'ASSETCATALOG_COMPILER_LAUNCHIMAGE_NAME';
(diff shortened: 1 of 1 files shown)


kerrishotts:
let there be tests!

kerrishotts:
Let there be /more/ tests!

PR #281, opened 2016-12-14
CB-12247: Hard link instead of soft linking

Mitko-Kerezov:
Softlinking resource files (.wav files [for example](https://github.com/apache/cordova-plugin-dialogs/tree/master/src/ios/CDVNotification.bundle)) leads to some unexpected behavior. Said resource files are stored as symbolic links inside the .ipa and the app cannot be installed on iOS 10.
Instead of creating symbolic link create regular links - it is better than copying the files performance-wise and the solution is more robust.

Part of the fix for [CB-12247: Symlinking resource files leads to inability to install app on iOS 10](https://issues.apache.org/jira/browse/CB-12247)

### DIFF
--- a/bin/templates/scripts/cordova/lib/plugman/pluginHandlers.js
+++ b/bin/templates/scripts/cordova/lib/plugman/pluginHandlers.js
@@ -304,7 +304,7 @@ function copyFile (plugin_dir, src, project_dir, dest, link) {
     shell.mkdir('-p', path.dirname(dest));
 
     if (link) {
-        symlinkFileOrDirTree(src, dest);
+        linkFileOrDirTree(src, dest);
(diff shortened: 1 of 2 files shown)


Mitko-Kerezov:
Note: I've tested this with `cordova-android` and it doesn't seem to reproduce - for one reason or another `gradle` seems to follow symlinks when zipping up the `.apk` file, so this fix may be enough to mitigate the whole problem. E.g. there's no need to propagate it to all the other platforms and to plugman as well

kerrishotts:
Can you update the title so that it includes the issue number with the following form:

issue#: description

That way the issue tracker will get all the updates here automatically.

Mitko-Kerezov:
Sure thing @kerrishotts

PR #283, opened 2016-12-30
CB-6274 Added support for BackgroundColor preference

mdoelker:
Supports #RGB #RRGGBB #AARRGGBB 0xRRGGBB and 0xAARRGGBB as noted in CB-6273. Fails gracefully for invalid color values (bg remains white).

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.h
+++ b/CordovaLib/Classes/Public/CDVViewController.h
@@ -79,6 +79,7 @@
 - (NSString*)appURLScheme;
 - (NSURL*)errorURL;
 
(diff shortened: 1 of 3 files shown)


shazron:
Thanks for the PR! Unit tests still need to be added.

mdoelker:
Right. Any hints what the correct place for this would be? It seems like there are no tests for any other preference yet. Would adding a test for the method to the CDVViewControllerTest.m be sufficient?

shazron:
That would be where I would put it, since you are testing the `colorFromColorString` from CDVViewController here.

PR #286, opened 2017-01-10
CB-12287: Remove hardcoded simulator build destination

kerrishotts:
### Platforms affected

iOS 4.3.1 and below

### What does this PR do?

In `cordova-ios@4.3.1` and below, the Xcode build destination when building for a simulator is hardcoded as "iPhone 5s". This works as long as the user has an iPhone 5s simulator installed, but if they remove it, the build process breaks.

This PR follows the same logic as `cordova emulate ios` in choosing a default destination, but also respects `--target` if specified. For example:

```
    $ cordova emulate ios       # will currently select iPhone SE simulator to build against and emulate
    $ cordova emulate ios --target="iPhone-7-Plus"  # will build against and emulate on the iPhone 7+
```

If there are no simulators available, the behavior is undefined, but Xcode wouldn't be able to build either.

Manual tests and `npm test`

### Checklist
    - Note: a couple of tests were modified so that they could pass, but they don't actually test this issue. Tests covering the inner logic will come in a separate PR

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -53,7 +53,33 @@ var buildFlagMatchers = {
     'shared_precomps_dir' : /^(SHARED_PRECOMPS_DIR=.*)/
 };
 
+/**
+ * Returns a promise that resolves to the default simulator target; the logic here
+ * matches what `cordova emulate ios` does. 
+ * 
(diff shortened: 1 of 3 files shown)


purplecabbage:
LGTM!
Not a breaking change in my opinion, no major bump required.

shazron:
LGTM

PR #293, opened 2017-03-02
CB-12523 - Remove iOS 8 support

shazron:
### Platforms affected

iOS

### What does this PR do?

Bump up minimum deployment target to iOS 9.

`npm test` and also created a new project with the cordova-ios platform that includes this change, and built for both device and simulator.

### Checklist


jcesarmobile:
A few tests are set to iOS 8 like https://github.com/apache/cordova-ios/blob/master/tests/spec/unit/fixtures/test-config-2.xml#L15

Should them also be updated?

shazron:
Travis CI: Looks like the iOS Simulator was never even run, and it failed. Tested it locally again and it works. Flaky CI (esp since it took over 24 hours before this was run on CI)

@jcesarmobile looks like the tests even pass with that 8.0 value in. Which tells me that value was never tested in the first place? (bad test)

PR #297, opened 2017-03-15
CB-12050 ios: Create shared scheme for framework target, for Carthage support

khawkins:
### Platforms affected
iOS

### What does this PR do?
Adds a shared build scheme for the Cordova framework target of the iOS project, which is a prerequisite for Carthage support.

1. Manually ran Carthage against this change, to ensure that it was able to download and build the framework target.
2. Imported the resultant framework into a test iOS app (Objective-C), instantiated a `CDVViewController` instance, added the default configuration to the app (config.xml, www/index.xml), and verified that the expected view would load at app startup.

### Checklist

**NB:** There are possibly scenarios where automated testing of Carthage support could be considered, but I'm not sure what that looks like at this point.


shazron:
Good stuff, tested it, it works as you described.
Two issues related to using module import using `@import Cordova;` which I will address in a separate PR though:
screen shot 2017-03-16 at 3 53 02 pm

shazron:
Posted #298

PR #301, opened 2017-04-05
CB-12617 : added engine strict for users with older node versions

audreyso:
### What does this PR do?
Added engine strict for users with older node versions

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,10 @@
     "tmp": "^0.0.26",
     "uncrustify": "^0.6.1"
   },
+  "engineStrict": true,
+  "engines" : { 
+    "node" : ">=4" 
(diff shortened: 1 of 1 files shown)


shazron:
`enginestrict` is useless without the corresponding `engines` key.
We should add this to the PR as well:
```
"engines" : { 
    "node" : ">=4" 
} 
```

audreyso:
Thank you @shazron . Updated!

shazron:
@audreyso I got this warning now:
```
npm WARN engineStrict Per-package engineStrict (found in this package's package.json)
npm WARN engineStrict won't be used in npm 3+. Use the config setting `engine-strict` instead.
```

PR #309, opened 2017-04-21
CB-8980: Ensure copied resource-files are cleaned

dpogue:
Requires https://github.com/apache/cordova-lib/pull/547 for the intended functionality, but is safe to merge even without it.

Addresses the point 1 of https://github.com/apache/cordova-android/pull/321#issuecomment-294985346 where `resource-file` targets were not getting cleaned.


shazron:
This will go in cordova-ios@4.4.1

PR #335, opened 2017-08-31
CB-12940: [IOS] Use deployment-target from config.xml for platform version in Podfile

shazron:
### Platforms affected

self

### What does this PR do?

Generate Podfile using deployment-target option from config.xml

Unit tests

### Checklist


shazron:
unfortunately in my tests, setting:
```
<platform name="ios">
        <preference name="deployment-target" value="8.0" />
    </platform>
```
in the root config.xml does not set the min deployment target in the Podfile. need to do more debugging and/or add more unit tests.

shazron:
Punting this to cordova-ios@4.5.1

shazron:
Punting this to cordova-ios@4.5.2

surajpindoria:
After some testing I can confirm that this is working. Setting the deployment-target in `config.xml` does properly set the platform version in the Podfile, but only on plugin add.

Follow up questions:
1. Should we also be checking deployment-target when we remove plugins?
1. Are there any other commands we should be checking for this? Or only plugin add/remove?

shazron:
Hmm. I think at minimum it should restore the default deployment target on plugin remove. What do you think? This can get hairy however, if somehow other plugins touched that value as well...

fssrepository:
the target is fine, but why the use_frameworks! has been removed, it simply breaks Swift based plugins, i had to add manually it above the target.

shazron:
use_frameworks has never been supported in Cordova iOS. See https://issues.apache.org/jira/browse/CB-11893

fssrepository:
So you meant to say the only way to use FCM library with that GoogleToolboxForMac products, if i download the 100MB and put inside tha plugin what has been done previously?

My main problem is leave it cordova for a while,just use the pure xcodebuild, if i install with pod install, why the hack he generates products folders, if he cannot build in mixed mode?

cthos:
Sorry to necrothread this, but I'm working on a Plugin with a pod file which requires the 9.0 to be included in the Podfile, and I don't _really_ want anyone to have to install an unreleased version of the cordova-ios platform in order to use it.

So I was wondering if there's any way for the plugin or project to hook into when the Podfile's generated? I poked around and it seems like it occurs between any of the hooks being called, so on the `before_install` it doesn't exist yet, and it'll never make it to the `after_install` because the `pod install` will fail.

Is there anything I can hack around while waiting on a `4.5.5` release?

shazron:
@cthos you will just have to patch (search+replace?) https://github.com/apache/cordova-ios/blob/f8a817bc8e7c1925c0c4ec7b474ec2ce48f1643d/bin/templates/scripts/cordova/lib/Podfile.js#L36 after the platform is added through the `after_platform_add` hook.

cthos:
@shazron - That doesn't cover the case where they just do a `cordova plugin add`, right?

It'd have to go something like this (I could totally be missing something key here though):
1. `cordova plugin add foo-plugin`
2. Said plugin fails to install due to the Podfile having the wrong version
3. `cordova platform rm ios`
4. `cordova platform add ios` <--- Hook fires after this is done?

Or are we saying the implementing project would have to implement such a hook _before_ they add the plugin?

shazron:
Yeah your last statement was what my scenario was attempting to do, sorry.

Try the plugin hook `before_plugin_install`, to patch the file.

cthos:
So in the `before_plugin_install` of the plugin, go patch `Podfile.js` to replace the version number, if the current platform version is at 4.5.4 or lower?

🤔 - That could work, yeah. I'll give that a shot.

zdennis:
Any ideas when this is going to be released? It doesn't appear to be in the latest 4.5.5.

dpogue:
The next major version of Cordova-iOS is [changing how we handle CocoaPods](https://github.com/apache/cordova-discuss/pull/108/files?short_path=9ca3847#diff-9ca38470fd0c4640b232cf20ded9b59f) to hopefully fix several of these sorts of issues.

PR #342, opened 2017-10-05
CB-13382: (ios) Added destroyWebView method for CDVViewController

surajpindoria:
### Platforms affected
iOS

### What does this PR do?
Adds a method to destroy CDVViewController webview object

### Checklist


shazron:
Thanks Suraj, since it's a new feature, the release that includes this should be 4.6.0? (minor version update)

CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -767,13 +767,23 @@ - (void)onAppDidEnterBackground:(NSNotification*)notification
 
 // ///////////////////////
 
+- (void)destroyWebView
+{
+    self.webViewEngine = nil;
```
jcesarmobile: I think this won't do anything if we don't call all what you put on the dealloc. If we are going to offer this new method, it should do everything what is on the dealloc and dealloc just call it. 

Anyway, the main problem of CB-13382 is that dealloc is not doing the complete cleanup of the webview, with the new code it will do it, so we don't really need to offer this new method as dealloc will do the cleaning automatically.

shazron: Woops I forgot about the dealloc in there since I haven't thought about it since everything is ARC. I think that we should just put `self.webviewEngine = nil` in the `dealloc` method, that would free it up I think.

PR #344, opened 2017-10-24
CB-13290: (ios) Update splash screen mapping for missing devices

surajpindoria:
### Platforms affected
iOS

### What does this PR do?
Adds splash screen support for iPhone X and other missing devices.

Created a project with all supported splash screens, opened in Xcode and viewed the Images.xcassets to see that the splash screens were correctly added.

### Checklist

### DIFF
--- a/bin/templates/scripts/cordova/lib/prepare.js
+++ b/bin/templates/scripts/cordova/lib/prepare.js
@@ -411,7 +411,9 @@ function mapSplashScreenResources (splashScreens, splashScreensDir) {
         {dest: 'Default-568h@2x~iphone.png', width: 640, height: 1136},
         {dest: 'Default-667h.png', width: 750, height: 1334},
         {dest: 'Default-736h.png', width: 1242, height: 2208},
-        {dest: 'Default-Landscape-736h.png', width: 2208, height: 1242}
+        {dest: 'Default-Landscape-736h.png', width: 2208, height: 1242},
+        {dest: 'Default-2436h.png', width: 1125, height: 2436},
+        {dest: 'Default-Landscape-2436h.png', width: 2436, height: 1125}
     ];
 
     var pathMap = {};
--- a/bin/templates/project/__PROJECT_NAME__/Images.xcassets/LaunchImage.launchimage/Contents.json
+++ b/bin/templates/project/__PROJECT_NAME__/Images.xcassets/LaunchImage.launchimage/Contents.json
@@ -1,6 +1,24 @@
 {
   "images" : [
     {
+      "extent": "full-screen",
+      "idiom": "iphone",
+      "subtype": "2436h",
+      "filename": "Default-2436h.png",
+      "minimum-system-version": "11.0",
+      "orientation": "portrait",
+      "scale": "3x"
+    },
+    {
+      "extent": "full-screen",
+      "idiom": "iphone",
+      "subtype": "2436h",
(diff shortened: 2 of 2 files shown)


bin/templates/scripts/cordova/lib/prepare.js
```diff
@@ -411,7 +411,9 @@ function mapSplashScreenResources (splashScreens, splashScreensDir) {
         {dest: 'Default-568h@2x~iphone.png', width: 640, height: 1136},
         {dest: 'Default-667h.png', width: 750, height: 1334},
         {dest: 'Default-736h.png', width: 1242, height: 2208},
-        {dest: 'Default-Landscape-736h.png', width: 2208, height: 1242}
+        {dest: 'Default-Landscape-736h.png', width: 2208, height: 1242},
+        {dest: 'Default-2436h.png', width: 1125, height: 2436},
+        {dest: 'Default-Landscape-2436h.png', width: 2436, height: 1125}
```
dpogue: This seems to be missing 1668×2224 (and landscape) & 2048×2732 (and landscape) for larger iPad Pro devices.

We really should be encouraging Launch Storyboards for those, but if it's as easy as adding more numbers here then it would be nice to support them.

Stefano1964: We have also to modify splashscreen plugin but there is a bug in xCode actually for generetad image (it's seem to be solve in xCode 9.1):
https://stackoverflow.com/questions/46263795/iphone-x-incorrect-launch-screen-orientation-used
I make it work (portrait only) with launch-screen image and by modifying splashscreen plugin, no reason to use launch storyboard, with launch-screen you have more granular control over your splash

surajpindoria: I did try adding those in, but when I went to check it in Xcode it shows those splash screen sizes as "unassigned", whereas the others are all tied to a device/OS version. Which made me think that they aren't properly supported for those larger screens.

dpogue: hmm, okay

Stefano1964: There is no support for iPad Pro (12.9) in xCode, there is no 2048x2732 and landscape asset catalog, this is Apple you can't do anything

kerrishotts: Yeah, the lack of support for larger screens is the reason we added launch storyboard support. IIRC there are hacks to use launch images for the 12.9” iPad Pro, but I don’t think you’d ever get Xcode to be happy with them. (Personally I fully expected Apple to require launch storyboards for the new iPhone X. Surprised they aren’t!)

Stefano1964: There is support for iPhone X in asset catalog (LaunchImage), just check iOS 8 and later, but there is a bug in xCode 9 (seems to be fixed in 9.1) that prevent to use LaunchImage for both portrait and landscape: https://stackoverflow.com/questions/46263795/iphone-x-incorrect-launch-screen-orientation-used
I modified the splashscreen plugin to support iPhoneX and is working, portrait only, for now.

mkn20: I am facing the same issue. Can you please share what specific changes you made in the splash screen plugin to get it working? Thanks in advance and regards.

PR #347, opened 2017-10-31
CB-13523: Add flag for Xcode-managed provisioning

dpogue:
### Platforms affected
iOS with Xcode 9

### What does this PR do?
This adds a `automaticProvisioning` option to the build config which passes flags to xcodebuild during the exporting step that allow it to automatically create and update provisioning profiles for automatic code signing.

I don't love that we're adding even more random stuff to the build config, but I suspect most people don't want this to be enabled by default.

Added a case to existing spec tests.

Tested with a Cordova project at work and confirmed to be working.

### Checklist
- [x] Reported [CB-13523](https://issues.apache.org/jira/browse/CB-13523)
- [x] Commit message follows the format

### DIFF
--- a/bin/templates/scripts/cordova/build
+++ b/bin/templates/scripts/cordova/build
@@ -42,6 +42,7 @@ var buildOpts = nopt({
     'codeSignIdentity': String,
     'codeSignResourceRules': String,
     'provisioningProfile': String,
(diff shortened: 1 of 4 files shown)


janpio:
Nice. 

(Side note: Please make sure that the new option also gets documented - these definitely need documentation)

PR #348, opened 2017-11-03
CB-13535: (ios) Add CallStart/CallEnd events to fire with Resign/Active events

SpencerEJameson:
### Platforms affected
iOS

### What does this PR do?
Adds new events to trigger alongside the current resign/active events to specifically denote when a call started and ended. The new events are **callstart** and **callend**

Tested against iOS 10.x+ and verified that upon calls connecting and disconnecting the events were triggering correctly. I did not have an iOS device less than 10.x to test with.

### Usage in App
`document.addEventListener("callstart", function() {
        console.log("Call Start");
      });`

`document.addEventListener("callend", function() {
        console.log("Call End);
      });`

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -28,6 +28,10 @@
 #import "CDVLocalStorage.h"
 #import "CDVCommandDelegateImpl.h"
 #import <Foundation/NSCharacterSet.h>
+#import <CoreTelephony/CTCallCenter.h>
+#import <CoreTelephony/CTCall.h>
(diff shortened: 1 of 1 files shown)


shazron:
This shouldn't be in the platform but in a plugin, since adding new platform features should either be:
1. cross-platform (all relevant platforms support it, and this has to be decided by consensus on the dev list)
2. part of a standard spec, i.e w3c (but still decided by point 1 above). 

Not all apps need this functionality thus it should be optional in a plugin, and this can be achieved in a plugin.

SpencerEJameson:
@shazron Totally, that makes sense. Thanks for providing the feedback!

PR #350, opened 2017-11-22
iOS inputfield bug

ngt14:
I developed a cordova app with a iframe, inside my iframe, i have input fields.

Sometimes i try to select an input field in my iframe page, the iPad/iPhone keyboard is show and focus on a input field selected but when i tapped it,is nothing fill on my input field.

How can i resolved it ? Thanks


janpio:
Definitely not by creating a Pull Request on the repo to "abuse" it as an issue.

If this is a real issue, post it here: https://issues.apache.org/jira/browse/CB/
If it is a question, ask it here: https://slack-cordova-io.herokuapp.com/ or here http://stackoverflow.com/questions/tagged/cordova
Thanks.

PR #352, opened 2017-12-06
CB-13510: (iOS) Removed clipboard workaround, which was braking copy/paste functional iOS 11

ssraza:
### Platforms affected
iOS

### What does this PR do?
Removed `[pasteboard setValue:string forPasteboardType:@"public.text"]` clipboard workaround, which is not need in iOS 11, but breaking copy/past functionality from outside the app (i.e notes/safari) into (Cordova) consumer app using native UI as well.

Manually tested copy/pastes functionality for both cordova (web UI) components as well as with native UI in a container app.

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database. [CB-13510](https://issues.apache.org/jira/browse/CB-13510)

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -739,6 +739,15 @@
 {
     // NSLog(@"%@",@"applicationWillEnterForeground");
     [self.commandDelegate evalJs:@"cordova.fireDocumentEvent('resume');"];
+    
+    if (!IsAtLeastiOSVersion((@"11.0"))) {
+        /** Clipboard fix **/
+        UIPasteboard* pasteboard = [UIPasteboard generalPasteboard];
+        NSString* string = pasteboard.string;
+        if (string) {
+            [pasteboard setValue:string forPasteboardType:@"public.text"];
+        }
+    }
 }
 
 // This method is called to let your application know that it moved from the inactive to active state.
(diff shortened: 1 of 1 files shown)


bhariharan:
@shazron @infil00p @stevengill Would you mind reviewing this PR? Thanks!

ssraza:
Not that I'm aware of...

purplecabbage:
```
#if __IPHONE_OS_VERSION_MAX_ALLOWED < 110000
    /** Clipboard fix **/		
    UIPasteboard* pasteboard = [UIPasteboard generalPasteboard];		
     NSString* string = pasteboard.string;		
     if (string) {		
         [pasteboard setValue:string forPasteboardType:@"public.text"];		
     }
 #endif
```
Can you try that on iOS 11 ?

ssraza:
Yep, let me try with this check shortly...

shazron:
You need to do this at runtime. Jesse's check above will omit the code needed on older iOSes during compile time on Xcode 9, thus break the feature at runtime for iOS 10 devices.

The pasteboard code above should only be loaded if the iOS version is less than 11.
```obj-c
if (!IsAtLeastiOSVersion('11.0')) {
  // put pasteboard code here
}
```
(IsAtLeastiOSVersion is in CDVAvailability.h)

ssraza:
@shazron @purplecabbage @bhariharan updated PR based on your suggestions... Thanks

purplecabbage:
Yep, just waiting on Travis ...

ssraza:
@purplecabbage @shazron if these changes work for you folks...

purplecabbage:
Thanks @ssraza !

PR #357, opened 2018-02-25
remove exit for start sim

hosseinamin:
There's no reason to exit after launch.

Also this action stops cordova-plugin-browsersync to keep listening.

### DIFF
--- a/bin/templates/scripts/cordova/lib/run.js
+++ b/bin/templates/scripts/cordova/lib/run.js
@@ -197,7 +197,7 @@ function deployToSim (appPath, target) {
 function startSim (appPath, target) {
     var logPath = path.join(cordovaPath, 'console.log');
 
-    return iossim.launch(appPath, 'com.apple.CoreSimulator.SimDeviceType.' + target, logPath, '--exit');
+    return iossim.launch(appPath, 'com.apple.CoreSimulator.SimDeviceType.' + target, logPath, false);
 }
 
 function listDevices () {


shazron:
Considering this for next release, after testing it.
Two things are needed:
1. File an issue at issues.cordova.io
2. Prefix the title with the issue tag (i.e. CB-XXXX)

shazron:
Ping @hosseinamin . If no progress on this for this week, this will be bumped until next release.

hosseamin:
Hi @shazron, Did add an issue, https://issues.apache.org/jira/browse/CB-13922

hosseinamin:
Hi @shazron, Did add an issue, https://issues.apache.org/jira/browse/CB-13922

shazron:
I just tested this. Unfortunately this breaks the expectation that after a (for example):
```bash
cordova emulate ios --target="iPhone-7, 11.2"
```
... the command exits, and the user gets back a command prompt -- the current behaviour. I think, what is needed here is a flag called `--no-exit` to handle your special case.

shazron:
Add a flag here: https://github.com/apache/cordova-ios/blob/7397337c932e072c0d68e6d1c61e6a5056d4f376/bin/templates/scripts/cordova/run#L50 and handle it in your code, switch on that flag

hosseinamin:
i don't know much about it. In my view this command exits after running simulator is ok.
Unless cordova-plugin-browsersync is installed and has `-- --live-reload`

```
cordova emulate ios --target="iPhone-7, 11.2"
```
If this exists i don't see in issue with it.

shazron:
@hosseinamin as I already said, this PR will not get in because it changes the default behaviour, which is to exit after running the simulator. If you want it to not exit per your requirements, you have to add the flag as a feature.

PR #358, opened 2018-02-26
CB-13523: Pass automaticProvisioning to build step

dpogue:
### Platforms affected
iOS

### What does this PR do?
Passes `-allowProvisioningUpdates` to the build/archive step. We are already passing it to the export step (https://github.com/apache/cordova-ios/pull/347) but it's also worth having it here.

Testing on Ayogo's Jenkins instance and was able to successfully make an iOS build without any manual steps to set up certificates or provisioning profiles.

### Checklist

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -171,7 +171,7 @@ module.exports.run = function (buildOpts) {
             // remove the build/device folder before building
             return spawn('rm', [ '-rf', buildOutputDir ], projectPath)
                 .then(function () {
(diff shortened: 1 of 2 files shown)


shazron:
This flag is an Xcode 9 feature, so we need to bump this as well: https://github.com/apache/cordova-ios/blob/master/bin/templates/scripts/cordova/lib/check_reqs.js#L29

I'm surprised that we still have that at 7...

dpogue:
Note this isn't enabled by default, it requires specifying `"automaticProvisioning": true` in build.json config, or as a `--automaticProvisioning` flag to the command

PR #288, opened 2017-02-13
CB-12402 CB-12206 Properly encode app name to generate XML files

matrosov-nikita:
### Platforms affected
self

### What does this PR do?
The build fails due to corrupted xml files at platforms/ios/<APPNAME>.xcworkspace/**. This PR adds logic to encode app name. 

### What testing has been done on this change?
Auto-test

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [x] Added automated test coverage as appropriate for this change.

PR #361, opened 2018-03-04
CB-13948 Properly encode app name to generate XML files using cordova-ios 4.3.1

ronaldozanoni:
### Platforms affected
self

### What does this PR do?
This PR merges the https://github.com/apache/cordova-ios/pull/288 and the branch _origin/4.3.x_.
This is needed due to the _cordova-ios 4.4.0_ has removed the _IOS 8_ support and we need it yet.

Auto-test using _Iphone 6s (IOS11)_ and some iPhone devices using IOS 8.

### Checklist


shazron:
Thanks for the PR, but this will not be released by us because we don't do back-releases, only forward looking releases. In any case, I'm not sure why there are many commits in this PR, since if you want that change you can cherry pick the commit (https://github.com/apache/cordova-ios/pull/288/commits/9b7efc9bb2af115cd5efb5eae79cfdad6c54b0d2) into the branch, and your PR will only contain 1 commit.

Also, often version support drops means we don't test on that version because of limited resources; often you can just change DEPLOYMENT_TARGET in the build settings and it should work. It should be something that can be changed in your config.xml as well.

ronaldozanoni:
Hi @shazron 
Thank you for your feedback.
I've closed this PR and created a new one using the _cherry-pick_ to add only that commit.
https://github.com/apache/cordova-ios/pull/362


About the DEPLOYMENT_TARGET, the problem is not it. The app throws an exception before loading the app on IOS <=8.3, as you can see the images below.

deploymenttarget
ios8bug

shazron:
@ronaldozanoni I understand, but my statement still remains "Thanks for the PR, but this will not be released by us because we don't do back-releases, only forward looking releases."

ronaldozanoni:
@shazron sure. I'm going to close the https://github.com/apache/cordova-ios/pull/362 and the JIRA task too.
If someone else needs it, use the branch from my fork https://github.com/ronaldozanoni/cordova-ios/tree/xml-encoded-fix-4.3.1

PR #363, opened 2018-04-23
CB-14045 - reinit url after app freezes

svalaskevicius:
Hi!

We've stumbled on a similar issue to this: https://issues.apache.org/jira/browse/CB-14045 although for us it takes a bit longer (10h or so of app being in background).

In our case the app resumes in a half working state, where the web view has been reinitialised pointing to about:blank with only default Js api available and no cordova js. Then, cordova tries to emit those resume and other events but all of them fail as there is no cordova js var available.

this approach does not fix app getting into that state as I am not fully aware how does it do that (12h debugging cycle is pretty hard).. but it gets it out of the broken state by pointing the webview back to the app's url.

what do you think?

Thanks!

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -720,17 +720,24 @@
     }
 }
 
-- (void)checkAndReinitViewUrl
+- (bool)isUrlEmpty:(NSURL *)url
 {
-    NSString *loadedUrl = [[self.webViewEngine URL] absoluteString];
-    if (loadedUrl == (id) [NSNull null] || [loadedUrl length]==0 || [loadedUrl isEqualToString:@"about:blank"]) {
-        NSURL* appURL = [self appUrl];
-        
-        if (appURL) {
-            NSURLRequest* appReq = [NSURLRequest requestWithURL:appURL cachePolicy:NSURLRequestUseProtocolCachePolicy timeoutInterval:20.0];
-            [self.webViewEngine loadRequest:appReq];
-        }
+    if (!url || (url == (id) [NSNull null])) {
+        return true;
     }
+    NSString *urlAsString = [url absoluteString];
+    return (urlAsString == (id) [NSNull null] || [urlAsString length]==0 || [urlAsString isEqualToString:@"about:blank"]);
+}
+
+- (bool)checkAndReinitViewUrl
+{
+    NSURL* appURL = [self appUrl];
+    if ([self isUrlEmpty: [self.webViewEngine URL]] && ![self isUrlEmpty: appURL]) {
+        NSURLRequest* appReq = [NSURLRequest requestWithURL:appURL cachePolicy:NSURLRequestUseProtocolCachePolicy timeoutInterval:20.0];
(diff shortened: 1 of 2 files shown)


CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -720,12 +720,25 @@ - (void)onAppWillTerminate:(NSNotification*)notification
     }
 }
 
+- (void)checkAndReinitViewUrl
+{
+    if ([[[self.webViewEngine URL] absoluteString] isEqualToString:@"about:blank"]) {
```
shazron: In the unlikely scenario that `appURL` is actually `about:blank`, you should check for this condition.
RIght now there is no probability of an infinite loop, but you never know if there are side-effects with any calls further down the stack in `checkAndReinitViewUrl `

I would return a `BOOL` from this function (`YES` if it re-inited, 'NO' if it did not), and add a unit test in https://github.com/apache/cordova-ios/blob/master/tests/CordovaLibTests/CDVViewControllerTest.m

svalaskevicius: thanks! I'll try to update these when I have a chance :) we've found another case when testing that the loaded url in the webview is (null), so added those checks too

svalaskevicius:
Hi @shazron , I've changed the function as requested, however I'm having problems with the test...

I've tried adding this test:
```
-(void)testIfItLoadsAppUrlIfCurrentViewIsBlank{
    CDVViewController* viewController = [self viewController];
    [viewController view];
    NSString* html = @"<html><body></body></html>";
    [viewController.webViewEngine loadHTMLString:html baseURL:@"about:blank"];
    XCTAssertTrue([viewController checkAndReinitViewUrl]);
    XCTAssertFalse([viewController checkAndReinitViewUrl]);
}
```

However, `loadRequest` does not load anything when running tests, and the url is always null.. 

Any chance you could take this PR from here and add the missing test? I suspect it should not be difficult, however might require some more background knowledge about cordova/ios.. 

I can also confirm that the approach to solve the initial problem (frozen apps for users) has been fixed by this PR (after also checking if it's not null *and* not "about:blank") 

Thanks!

shazron:
@svalaskevicius sure, will do. Can you make sure the checkbox "allow maintainers to make edits" is checked? Thanks: https://blog.github.com/2016-09-07-improving-collaboration-with-forks/

shazron:
Made a mistake - didn't realize this was based off 4.5.x branch so it was pulled into that. Reverted the change to 4.5.x and applied to master instead.

svalaskevicius:
thanks!

GulinSS:
Did it was added for current `4.5.5` public release?

brodybits:
> Did it been added for current `4.5.5` public release?

No it was removed from `4.5.5` since we intended to merge the fix on `master` as discussed in <https://github.com/apache/cordova-ios/pull/363#issuecomment-389401996>. I just raised PR #418 in case we do another cordova-ios@4.5.x patch release.

As a general rule I recommend asking this kind of question in a new issue or by email to avoid losing such comments in closed issues and merged PRs.

PR #365, opened 2018-05-14
CB-14076: (ios) Setting UIWebViewDelegate on CDVWebViewEngineProtocol doesn't work

dpolivy:
Attempts to use `CDVWebViewEngineProtocol::updateWithInfo` to set a new `UIWebViewDelegate` were not working; instead of setting the passed in delegate, it was setting the `CDVViewController` instance as the delegate instead.

### Platforms affected

cordova-ios

### What does this PR do?

This commit fixes the `CDVWebViewEngine` class so that the proper delegate is ultimately passed in to the `UIWebView` when calling `updateWithInfo`.

I used the code to create a new test app, to which I applied a test plugin which attempts to set the `UIWebViewDelegate`. I verified that I was able to successfully set the delegate, and receive the events from the `UIWebView`.

I also ran `npm test` and verified all tests passed.

### Checklist


shazron:
Sorry I take that back, it is already set to the default in `pluginInitialize`

PR #376, opened 2018-06-23
Fixed Get Apple OSX Version method

erisu:
### Platforms affected
iOS

### What does this PR do?
While working on an upcoming PR (#377 ) for improving the test coverage, I noticed that the `get_apple_osx_version` method has been returning `undefined`. 

After further investigation, the RegEx pattern has been looking for OSX SDK version `OS X` no longer exists since Xcode 8+ 

Example print outs:
**Xcode 7.3:** https://travis-ci.org/erisu/macos-test/jobs/395785417
**Xcode 8:** https://travis-ci.org/erisu/macos-test/jobs/395785418
**For other versions:** https://travis-ci.org/erisu/macos-test/builds/395785415

From the above print out, I determined that the RegEx pattern should be updated to `macOS`

### What testing has been done on this change?
- https://travis-ci.org/erisu/cordova-ios/builds/395835801
- https://ci.appveyor.com/project/erisu/cordova-ios/build/1.0.3

**Note:** Oddly Node 4 appears to be failing but I have submitted PR #375 to drop Node 4 support.

##~# Checklist

- ~[Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database~~
- ~~Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.~~
- ~~Added automated test coverage as appropriate for this change.~~

PR #375, opened 2018-06-23
CB-14074 ios: Remove Node 4 support from CI & package.json

erisu:
### Platforms affected
iOS

### What does this PR do?
- Drops Node 4 from travis and appveyor
- Increase the engine requirement of node in package.json
- Added engineStrict to package.json. (Match with other platforms e.g cordova-android)
- Cleanup .travis.yml

### What testing has been done on this change?
- https://travis-ci.org/erisu/cordova-ios/builds/395831324
- https://ci.appveyor.com/project/erisu/cordova-ios/build/1.0.1

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #377, opened 2018-06-23
Increase Code Coverage & use nyc

erisu:
### Platforms affected
iOS

### What does this PR do?
I wanted to help increase the code coverage for iOS so I wrote some additional tests around the missing area.

I also:
- Upgraded all of the testing and linting related dependencies.
- I replaced the outdated Istanbul code coverage npm package `istanbul@^0.4.2` with their new maintained Istanbul nyc package `nyc@^12.0.2`.

This PR requires PR #375 and #376 to be merged before hand. I will rebase this PR as the other PRs are merged.
- PR #375 deprecates Node 4. This PR contains ES6 usage (E.g let, const, etc) which is not supported in Node 4.
- PR #376 fixes an issue with detecting the macOS version. This PR tests around the correction.

**Note:** The links provided below contains results with the required PRs, mentioned above, merged.
- https://travis-ci.org/erisu/cordova-ios/builds/395853027
- https://ci.appveyor.com/project/erisu/cordova-ios
- https://codecov.io/gh/erisu/cordova-ios/tree/dd60a385b4352c3d6285f63bdb236b10f4f79629/bin/templates/scripts/cordova/lib

### Checklist
- ~~[Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database~~
- ~~Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.~~


tests/spec/unit/versions.spec.js
```diff
@@ -0,0 +1,91 @@
...
+            });
+
+            it('should find xcodebuild version.', (done) => {
+                versions.get_tool_version('xcodebuild').then((version) => {
+                    expect(version).not.toBe(undefined);
+                    done();
+                });
+            });
+
+            it('should find ios-sim version.', (done) => {
```
brodybits: For some reason `versions.get_tool_version('ios-sim')`, `versions.get_tool_version('ios-deploy')`, and `versions.get_tool_version('pod')` do not work for me when I test on my local system, don't know why. Any clues?

I would also favor moving these into a separate script.

package.json
```diff
@@ -67,5 +67,9 @@
     "shelljs",
     "xcode",
     "xml-escape"
-  ]
+  ],
+  "nyc": {
```
brodybits: If I would do `npm install` it writes the following change to `package.json`:
```diff
index 7b21884a..ee6b352a 100644
--- a/package.json
+++ b/package.json
@@ -69,7 +69,12 @@
     "xml-escape"
   ],
   "nyc": {
-    "include": ["bin/templates/scripts/**"],
-    "reporter": ["lcov", "text"]
+    "include": [
+      "bin/templates/scripts/**"
+    ],
+    "reporter": [
+      "lcov",
+      "text"
+    ]
   }
 }
```

I would favor having the "nyc" entry committed this way.

erisu: Are you referring to the formatting, each item on its own line? I can make this change.

brodybits: Yes. I have a few other general remarks coming on this nice PR.

brodybits:
Really nice in general. My major comment is that I am personally not ready to break deprecated Node.js 4 compatibility for a couple reasons:
- make it easier to backport to 4.5.x branch if needed
- I think we should drop Node.js 4, remove committed node_modules, and increase the major version all at the same time. I don't really want to do this without getting some agreement from others. I think these items should be done at the same time in its own PR, would be happy to take this one on.

I think we can keep deprecated Node.js 4 compat a couple straightforward changes:
- do not upgrade eslint devDeps in this PR
- use `var` instead of `let` & `const` for now

I think a valid alternative for using `let` & `const` would be to start using "use strict", not sure if we want to keep this for the future or not. Also if we introduce `let` & `const` I think we should consider transitioning the existing `var` declarations.

tests/spec/unit/versions.spec.js
erisu: Are these tools installed locally?  These tests can only run on a Darwin environment and with these tools installed. 

I did not mock the child process exec so it an actual test, but does not care what is the exact version.

Both locally and on the Travis CI environment has these tools are installed. I did not configure Travis CI to install them. Either the environment already came with the tools or was installed via npm install. From package.json or the .travis.yml script step.

I could mock the data, but by mocking, I would not have discovered that the Mac OSX SDK changed to macOS SDK.

erisu: As for separate scripts, do you mean something like a bash file? In `/cordova-ios/bin` there are some individual script files that calls onto `versions.js` to fetch the value for an individual item. 

For example: 
`/cordova-ios/bin/apple_ios_version`
`/cordova-ios/bin/apple_osx_version`
`/cordova-ios/bin/apple_xcode_version`

But it is missing the `ios-sim`, `ios-deploy`, and `pod` specific executable script.

brodybits: No I meant separate .spec.js file.

I will try `npm i --save-dev ios-deploy` and see if that solves at least part of the problem.

brodybits: `npm i --save-dev ios-deploy` seems to do the trick, except for `versions.get_tool_version('pod')`; I just pushed the change to your branch

tests/spec/unit/versions.spec.js
```diff
@@ -0,0 +1,91 @@
...
+                    done();
+                });
+            });
+
+            it('should find pod version.', (done) => {
+                versions.get_tool_version('pod').then((version) => {
+                    expect(version).not.toBe(undefined);
+                    done();
+                });
+            });
```
brodybits: This case assumes that `pod` is installed. Easy solution for me is to install using homebrew as discussed in <https://github.com/CocoaPods/CocoaPods/issues/2238#issuecomment-359082649>.

I have mixed feelings about this test. Positive is that it verifies that `versions.get_tool_version()` works on globally installed `pod` tool. Negative is that it may break some developers ability to try `npm run unit-tests` before raising PRs.

erisu:
@brodybits 

With the recent mailing list discussion, to mark the next major Cordova dev release (cordova-ios@5.0.0-dev),
does it mean we can and will drop Node.js 4 for cordova-io after the release?

Some repos have already dropped Node 4 support and began using  `let`, `const`, `class` etc in master. For example cordova-android.

As for using `var` vs. `let & const`. I would hope to continue to use let and const if we plan to merge this after the bump and the usage at that time is OK. 

Since the other repos are starting to use them, why not all. 
I know the drop support task will take time to complete across all repos. I would like to have a uniform coding syntax across all repos but, I know it will be difficult. Als everyone has their own coding preference anyways. he

As for the transitioning of var declarations to let and const. I had been making these changes as I go, not mass search and replace.

I also agree and excited to see the committed node_modules removed. I have always been wondering behind this.

`do not upgrade eslint devDeps in this PR` 
I suspect this means to revert the eslint only devDeps from my PR, and we will be handled in #382 ?

The only concern I had with the other PR was the `@next` devDeps usage. I don't normally recommend using non-official release deps in production, though its only used for testing/linting and would not affect user projects.

I haven't looked too much into why it was decided.

tests/spec/unit/versions.spec.js
erisu: Adding `ios-deploy` to `package.json` as a devDep will cause AppVeyor to fail. This package can only be installed and used on the Darwin platforms. 

I question if testing iOS/macOS platform on AppVeyor is actually needed. Xcode is a build requirement and Windows cant use Xcode, and AppVeyor is running on Windows, what is achieved with this test.

```
npm ERR! code EBADPLATFORM

npm ERR! notsup Unsupported platform for ios-deploy@1.9.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"ia32"})
npm ERR! notsup Valid OS:    darwin
npm ERR! notsup Valid Arch:  any
npm ERR! notsup Actual OS:   win32
npm ERR! notsup Actual Arch: ia32
```

erisu: Also should this dep be added to the platform? 

Per documentation, it's expected that the user installed it manually and globally. 

Having it bundled with the platform might mean fewer setup steps? 
Could the globally installed ios-deploy version be affected and break when a platform major versions is released? So therefore they should become a dep?

https://cordova.apache.org/docs/en/latest/guide/platforms/ios/#installing-the-requirements

brodybits:
@erisu I will raise additional PRs with proposal to split some things more clearly out of this PR (#377) and #382, then address your comments.

brodybits:
@erisu I think this is good, hope we can deal with a few minor points:
- I would have preferred that we deal with the eslint changes separately. I proposed a couple alternative solutions in #389 & #390, would rather discuss outside this PR.
- Testing `versions.get_tool_version()` on ios-deploy & pod could add a barrier for some developers who may want to do `npm test` or `npm run unit-tests` on local machine. I would kinda rather we test `versions.get_tool_version()` on tools that should be over 99% sure on every app developer machine, such as node, npm, git
- maybe cleanup commits (I think we don't want to do a squash merge on this one)

I would be fine if you want to finish and merge this one yourself, if you want me to review again, or even if you would like me to take this one over.

erisu:
@brodybits When you have some time, can you review this one more time? Especially the last three commit.

The last three commit should cover the items from your previous message.

* Reverted ESLint Updated: Will be covered in another PR.
* Restructured Testing: There are now three jasmine configurations.
(Unit, Component, and Coverage)

Regards with the second item point, I also agree that users may not install external dependencies such as pods and ios-deploy but should be able to run unit testing.

In this case, I introduced component testing. This is a good example of component level testing where it involves two or more components. I didn't use the term e2e as it not a full end-to-end test. 

* Travis CI had been updated to run both unit and component.

* Users can continue to use `npm run unit-test` or `npm test` and not be affected by component testing. 

* Added Jasmin coverage configs which basically combines unit and component test so the code cover report out is accurate.

Regards to the last item, commit cleanup, can you what is intended? Typically I squash my commits but since it was mentioned to possible not squashing, I am trying to figure out cleanup in what way.

Please let me know your thoughts.

brodybits:
@erisu now I understand a few things. My one major criticism is that I think `npm test` should do all existing tests, meaning `npm test` should include e2e (general case) or component (this case). If user does not have standard iOS dev tools including ios-deploy and pod then I think it is fine to say that full `npm test` would not work. I think these tools should not be extremely hard to install for someone working on cordova-ios.

I personally do not really like squashing multiple kinds of changes together, like both migrating to nyc and adding more test coverage. More things at the same time could be more to grasp at the same time. Major factor is that if someone wants to back-port to an earlier release, separate commits would be nicer. But this is just my personal opinion, some others seem more likely to squash.

And a side question: I thought you had made a change to introduce jasmine-spec-reporter but couldn't find it. Maybe it was someone else. I think it would be nice, probably in a separate PR.

brodybits:
I would like to add another comment that I think the concept of unit test vs e2e integration test is currently not consistent between all Cordova repos, and is actually not so clear to me. For example: it is actually not so clear to me what defines "spec" test vs e2e integration test in cordova-lib. Noting this here before it gets forgotten.

erisu:
@brodybits Yes, I also see sometimes inconsistencies across all repos. Basically, I see testing broken into three categories. This is how I view testing in general. 

What tasks or open action items we should do from these comments is something I am not sure about, as of yet. Maybe we just need document the difference and when to pick one over the other. We should update the repos at least to be consistent with the structure.

Sorry if it sounds a bit generic and maybe textbook. 

*Unit Testing*, which are the test specs and I feel in most cases these are done well.

*Component Testing* is testing between two or more components. For example our previous situation with the pods and ios-deploy tools. Component testing may also involve environment configurations to ensure that the test work. One good old example is the Application + Database

*E2E Testing*  is testing the entire system as a whole. How we define E2E, I am also not clear. It seems in this repo, E2E represents if the basic app builds. In most cases, I feel E2E starts from the highest point in how end users typically use Cordova. In this case, I would think all of our E2E tests would be within CLI.

There are other testing categories like *UI testing*.

erisu:
I also want to update package.json. But when, where, and should it be brought up. Some of it could be a bit opinionated though.

1. Not all commands are consistent between each repo. I think they should be.

Example:
`cordova-ios` has `npm run unit-tests`
`cordova-common` has `npm run jasmine`
`cordova-docs` has `npm run test-unit`

2. (opinionated) The script naming convention.
```
"test": "",
"posttest": "",
"cover": "",
"e2e-tests": "",
"objc-tests": "",
"objc-tests-lib": "",
"objc-tests-framework": "",
"preobjc-tests": "",
"unit-tests": "",
"eslint": ""
```
would become
```
"coverage": "",
"test:e2e": "",
"test:objc": "",
"test:objc:lib": "",
"test:objc:framework": "",
"test:objc:pre": "",
"test:unit": "",
"linting": ""
```
This is opinionated so there is no priority and if no action, it is all good.

3. (opinionated) Order of the testing.
```
"test": "npm run e2e-tests && npm run objc-tests && npm run unit-tests",
"posttest": "npm run eslint",
```
becomes
```
"pretest": "npm run eslint",
"test": "npm run test:unit && npm run test:component && npm run test:objc && npm run test:e2e"
```
I think ESlint should be pre instead of post? if linting fails, maybe the test should stop right there. It should be the fastest task to run.

The test order also appears reversed IMO. 
e2e tests will take the longest time to run. It should be the to run. 

If a unit test fails, the likely outcome of the e2e test is also a fail. The CI should basically stop at the unit test if it fails. It would take a longer time for the CI to fail on an e2e when the issue could have been seen at a unit level.

Prefered Order: unit -> component -> e2e

brodybits:
Those testing categories seem to make sense, would need some more time to better understand. Should ideally be discussed somewhere else if it affects multiple repos. Changes approved on my part.

PR #392, opened 2018-08-10
Pin jasmine@3.1.0

erisu:
### Platforms affected
ios (tests)

### What does this PR do?
This PR fixes the recent test failures on `master` branch. 

These tests are failing because of the latest `Jasmine@3.2.0` that has been released within the past 2 days. (https://github.com/jasmine/jasmine-npm/releases/tag/v3.2.0)

At the time the original PR was submitted and tested, Jasmine was at version `3.1.0`. On this version, all tests were passing. https://travis-ci.org/apache/cordova-ios/builds/412472891

By pinning Jasmine@3.1.0 in `package.json`, the tests will continue to pass.

- local `npm test`
- Travis CI: https://travis-ci.org/erisu/cordova-ios/builds/414529064
- AppVeyor: https://ci.appveyor.com/project/erisu/cordova-ios/build/1.0.34

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "eslint-plugin-node": "^5.1.0",
     "eslint-plugin-promise": "^3.5.0",
     "eslint-plugin-standard": "^3.0.1",
-    "jasmine": "^3.1.0",
+    "jasmine": "~3.1.0",
     "nodeunit": "^0.8.7",
     "nyc": "^12.0.2",
     "rewire": "^4.0.1",


brodybits:
Nice temporary solution. TBH I would kinda favor updating the test to pass on jasmine@3.2.0.

erisu:
@brodybits The questions I have: 
- Do we merge this temporary fix so master is green while we can continue to research into Jasmine 3.2.0? 
- Do you have any issues if this was merged in?

The problem could be that these tests have always been broken and with the new Jasmine release, it bubbled up the issue. The last PR does not affect actual code; it only added new tests, updated a couple of testing deps (Jasmine, Rewire, and added NYC), and added component testing.

The other possibility is that Jasmine actually introduced a bug and the tests are OK. If the tests are updated to work with the possible new bug, it is likely that the tests will fail again once Jasmine releases a new version.

Either way, I don't deny that we should investigate to find out where the issues originate and correct if it's our tests.

janpio:
Priority is always to get `master` back to green again. In this case the fastest way to achieve this, is to go back to the state before the jasmine release. Pinning the version achieves exactly that.

Tests can (and should!) be updated for the new jasmine release in an additional PR. But that can happen while `master` is green.

brodybits:
As a minor comment I would favor the following approach:
- specify jasmine@~3.1.0, which would block the jasmine@3.2.0 update in a slightly more general way
- raise a followup PR, maybe a WIP PR, that bumps to jasmine@^3.2.0 to track that we want to work on this (alternative would be to raise this as an issue)

erisu:
Ok, I will:
1. Update `package.json` with `jasmine@~3.1.0` so that patch versions are still acceptable.
2. Create an issue ticket so we can revisit at a later time to investigate and fix tests for `jasmine@^3.2.0`.

Issue #402, opened 2018-09-06
Line separator (U+2028) and paragraph separator (U+2029)  Got JS Exception SyntaxError


crmo:
When Objective-c call JavaScript use function `stringByEvaluatingJavaScriptFromString:` with Line separator (U+2028) or paragraph separator (U+2029), JavaScript will get error.
In `CDVUIWebViewEngine.m`, function `evaluateJavaScript:completionHandler:`. The caller may pass in these special characters.Before call `stringByEvaluatingJavaScriptFromString:`,we can do some filtering.

brodybits:
I found the same bug on cordova-osx. I discovered the same issue came up on cordova-android@6, not sure if it is still an issue on cordova-android@7 or not. I investigate and report next week.

dpogue:
As per the comments and justification in https://github.com/apache/cordova-ios/pull/403, I'm going to close this as "not planned".

PR #403, opened 2018-09-06
fix#402 Line separator (U+2028) and paragraph separator (U+2029) Got …

crmo:
…JS Exception SyntaxError

### Platforms affected
iOS

### What does this PR do?
Fix Issus #402 Line separator (U+2028) and paragraph separator (U+2029) Got JS Exception SyntaxError.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVUIWebViewEngine/CDVUIWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVUIWebViewEngine/CDVUIWebViewEngine.m
@@ -67,6 +67,8 @@
 
 - (void)evaluateJavaScript:(NSString*)javaScriptString completionHandler:(void (^)(id, NSError*))completionHandler
 {
+    javaScriptString = [javaScriptString stringByReplacingOccurrencesOfString:@"\u2028" withString:@""];
+    javaScriptString = [javaScriptString stringByReplacingOccurrencesOfString:@"\u2029" withString:@""];
     NSString* ret = [(UIWebView*)_engineWebView stringByEvaluatingJavaScriptFromString:javaScriptString];
 
     if (completionHandler) {


janpio:
Thanks for the PR! 🥇 

Can you explain a bit why and when someone might call this function with these characters?
Are those the only ones that could be problematic?

crmo:
> Can you explain a bit why and when someone might call this function with these characters?

I custom a plugin in my project, Its function is to get data from the network and return it to JavaScript.I call `sendPluginResult:callbackId:` like this:
```
    // fetch from network, may contain \u2028 or \u2029
    NSString *data = @"...\u2028...";
    CDVPluginResult* pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK messageAsString:data];
    [self.commandDelegate sendPluginResult:pluginResult callbackId:aCallBackId];
```

> Are those the only ones that could be problematic?

Sorry,I am not sure about this problem.But I find a [table](http://tools.jb51.net/table/javascript_escape) about js special characters,and [some discussion about this issue](https://stackoverflow.com/questions/2965293/javascript-parse-error-on-u2028-unicode-character).I verified these special characters.Only these two characters have problems

CordovaLib/Classes/Private/Plugins/CDVUIWebViewEngine/CDVUIWebViewEngine.m
```diff
@@ -67,6 +67,8 @@ - (void)pluginInitialize
 
 - (void)evaluateJavaScript:(NSString*)javaScriptString completionHandler:(void (^)(id, NSError*))completionHandler
 {
+    javaScriptString = [javaScriptString stringByReplacingOccurrencesOfString:@"\u2028" withString:@""];
+    javaScriptString = [javaScriptString stringByReplacingOccurrencesOfString:@"\u2029" withString:@""];
```
brodybits: Why not combine these?

brodybits: P.S. I would have combined these with the line below.

brodybits:
In general I would down-vote the proposed solution since it removes the`U+2028` and `U+2029` characters, which is not always wanted. I think a better solution would be to replace these two characters with `\u2028` and `\u2029`, which is kind of an inverse solution to <https://stackoverflow.com/questions/4901133/json-and-escaping-characters>.

Some bonus items:
- test coverage
- consider applying the fix upstream (in a calling function)
- I think a similar fix is needed on cordova-osx as well.

The `cordova-sqlite-storage` plugin that I maintain has test cases to cover this functionality.

shazron:
I don't think this is will be accepted. For one, the data should be processed before it is sent (as @brodybits suggets) as a workaround, and we shouldn't touch the evaluateJavascript function which is just an abstraction for the various webview functions.

PR #408, opened 2018-09-15
CB-14192: (ios) - Xcode 10 Framework Target Fix - Convert Framework to build source rather than wrap static lib

sgoldberg-sfdc:
-  The new build system in Xcode 10 gets confused by the static library header files copied to $CONFIGURATION_BUILD_DIR/include
-  By making the framework target build the source directly we eliminate this issue
-  If both targets are built, the issue will still occur

### Platforms affected
iOS

### What does this PR do?
Modifies the framework target to compile the source directly, rather than wrap the static library target. This avoids the issue where the static library header files are copied to the $CONFIGURATION_BUILD_DIR confusing the new build system (defaulted on in Xcode 10) resulting in duplicate definition errors. 

Manually verified builds of both targets within xcode and via Carthage.

### Checklist
- [X] Added automated test coverage as appropriate for this change. (None required)

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVGestureHandler/CDVGestureHandler.h
+++ b/CordovaLib/Classes/Private/Plugins/CDVGestureHandler/CDVGestureHandler.h
@@ -17,7 +17,7 @@
  under the License.
  */
 
-#import "CDVPlugin.h"
+#import <Cordova/CDVPlugin.h>
 
 @interface CDVGestureHandler : CDVPlugin
 
--- a/CordovaLib/Classes/Private/Plugins/CDVHandleOpenURL/CDVHandleOpenURL.h
+++ b/CordovaLib/Classes/Private/Plugins/CDVHandleOpenURL/CDVHandleOpenURL.h
@@ -17,7 +17,7 @@
  under the License.
  */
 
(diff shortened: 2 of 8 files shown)


sgoldberg-sfdc:
Closing for now, more at play here than the framework target.

bhariharan:
@dpogue @shazron Could we get this merged? Without this, it's impossible to use `CordovaLib` in a pure `Swift` project because `Swift` ignores the existing linker flag workaround that we use to make it work with an `Objective C` project, i.e. `Allow Non-modular Includes in Framework Modules`.

dpogue:
I had been hoping to merge this (in the form of https://github.com/apache/cordova-ios/pull/409) once I'd figured out why automatic signing was still broken, but it might make sense to fix this one issue and tackle the signing separately since it's not clear that they are related.

bhariharan:
That would be much appreciated, @dpogue!

dpogue:
https://github.com/apache/cordova-ios/pull/409 (which includes the changes from this PR, with credit to sgoldberg-sfdc) has been merged in preparation for the next major version of cordova-ios, so I'm going to close this PR. Thanks for your contribution and for paving the way to support Xcode 10!

PR #409, opened 2018-09-16
Update project files for Xcode 10

dpogue:
This is an alternative proposal to #408, but I lack the knowledge to make any definitive statements about one being better than the other.

/cc @sgoldberg-sfdc 

### Platforms affected
iOS

### What does this PR do?
Updates the CordovaLib.xcodeproj and template xcodeproj files to use the recommended settings from Xcode 10. This appears to solve the problems with building in Xcode 10, but doesn't appear to break building in Xcode 9.

Manual testing done with Xcode 10 GM and Xcode 9.4.1.
Updated unit tests for new behaviour.


brodybits:
+1

Looks much less dramatic than #408, especially good for people with
existing Cordova-iOS projects.

sgoldberg-sfdc:
WIth this branch I still see the Xcode 10 failures

`Testing failed:
	Redefinition of enumerator 'CDVCommandStatus_NO_RESULT'
	Redefinition of enumerator 'CDVCommandStatus_OK'
	Redefinition of enumerator 'CDVCommandStatus_CLASS_NOT_FOUND_EXCEPTION'
	Redefinition of enumerator 'CDVCommandStatus_ILLEGAL_ACCESS_EXCEPTION'
	Redefinition of enumerator 'CDVCommandStatus_INSTANTIATION_EXCEPTION'
	Redefinition of enumerator 'CDVCommandStatus_MALFORMED_URL_EXCEPTION'
	Redefinition of enumerator 'CDVCommandStatus_IO_EXCEPTION'
	Redefinition of enumerator 'CDVCommandStatus_INVALID_ACTION'
	Redefinition of enumerator 'CDVCommandStatus_JSON_EXCEPTION'
	Redefinition of enumerator 'CDVCommandStatus_ERROR'
	Typedef redefinition with different types ('enum CDVCommandStatus' vs 'enum CDVCommandStatus')
	Duplicate interface definition for class 'CDVPluginResult'
	Property has a previous declaration
	Property has a previous declaration
	Property has a previous declaration
	Property has a previous declaration
	Duplicate interface definition for class 'CDVInvokedUrlCommand'
	Property has a previous declaration
	Property has a previous declaration
	Too many errors emitted, stopping now
	linker command failed with exit code 1 (use -v to see invocation)
	Testing cancelled because the build failed.
** TEST FAILED **'
`
type errors. This is with both Carthage and building in the IDE, and from xcode build.

sgoldberg-sfdc:
Updating to recommended settings is something that should happen regardless of the current build issues though.

dpogue:
@sgoldberg-sfdc What commands are you running to test this?

I created a new Cordova project based on the changes in this PR and was able to make builds using Xcode 10, both in Xcode and via the command-line, both debug and enterprise release builds, and in the simulator.

sgoldberg-sfdc:
@dpogue The issue seems to stem from how the new build system handles the include directories, and the fact that the static lib copies the headers to a folder that would get picked up by Xcode on a subsequent build. Depending on the state of your derived data, it may appear to build as it is skipping certain build steps. Try building the Framework App target with a clean derived data and the new build system and you will see the issue.

dpogue:
> Try building the Framework App target

Not sure what you mean by this. There's only one app in my newly-created project.

dpogue:
Ahhh, I didn't even realize that project existed. It probably also needs to be updated to recommended settings, and hopefully that will fix the issues.

Will try to tackle that tonight!

mhartington:
Tested this PR out, only thing I can note is with regards to splashscreens/launch images not showing up when emulating from the command line. Not sure if that is because of the changes or something else though.

Other than that, builds work from the command line. Thanks @dpogue !

sgoldberg-sfdc:
I believe when you generate a project using Cordova it uses the static library, so the framework issues will not appear in the typical workflow.

sgoldberg-sfdc:
We consume the framework as we use Carthage for dependency management. I don't think I would force it on developers, but an option to the CLI might work. 

Another simple (but of unknown impact to downstream developers) solution, would be to change the static library public headers folder such that its not include/Cordova , something like Cordova/include.

dpogue:
@mhartington The splashscreen/icon issue you noticed, is that specific to this branch or does it also affect projects using master (with the `-UseModernBuildSystem=0` flag)? It was only happening with Xcode 10?

mhartington:
Happens with the `UseModernBuildSystem` and the CLI emulate/run commands with master as well as with this PR applied

leshik:
I tried this patch and for me it builds the project successfully, but then fails to launch it (when using `cordova run ios --device`.) For now I've disabled this modern build thing.

PR #413, opened 2018-09-18
remove node_modules from repository

brodybits:
- ignore all node_modules
- completely remove node_modules from repository


brodybits:
My bad, apologies!

PR #418, opened 2018-09-27
CB-14045 - reinit url after app freezes on 4.5.x

brodybits:
Fix for [CB-14045](https://issues.apache.org/jira/browse/CB-14045) (reinit url after app freezes) was mistakenly raised on 4.5.x in PR #363, merged on 4.5.x, reverted, then merged on `master`.

I think this fix should have also been part of 4.5.x.

I am raising this PR just in case we do another cordova-ios@4.5.x patch release.

The commit ID on `master` is: 6e790afb0e8e7e6e427960b6e884462dc981744f

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -720,12 +720,33 @@
     }
 }
 
+- (bool)isUrlEmpty:(NSURL *)url
+{
+    if (!url || (url == (id) [NSNull null])) {
+        return true;
+    }
+    NSString *urlAsString = [url absoluteString];
+    return (urlAsString == (id) [NSNull null] || [urlAsString length]==0 || [urlAsString isEqualToString:@"about:blank"]);
+}
+
+- (bool)checkAndReinitViewUrl
+{
+    NSURL* appURL = [self appUrl];
+    if ([self isUrlEmpty: [self.webViewEngine URL]] && ![self isUrlEmpty: appURL]) {
+        NSURLRequest* appReq = [NSURLRequest requestWithURL:appURL cachePolicy:NSURLRequestUseProtocolCachePolicy timeoutInterval:20.0];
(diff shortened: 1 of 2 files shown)


janpio:
I have no idea what this PR is actually doing without following the links.

Does this cherry pick a commit from somewhere else for this release branch? If so, please say so in the PR description. The PR title should also be about what the commit fixes, not what issue is merged into what. Include that as additional information in the PR description please.

brodybits:
@janpio I just updated the title and description. Does it look OK now?

janpio:
yep, now I can follow

GulinSS:
Please somebody merge it

jacobg:
Does this commit work with CodePush using dynamic url?

brodybits:
> Does this commit work with CodePush using dynamic url?

I suggest you try cordova-ios@nightly with CodePush.

brodybits:
Now merged onto 4.5.x, in case we ever publish 4.5.6 patch release before Cordova 9.

dpogue:
> Does this commit work with CodePush using dynamic url?

I think it'll be fine, unless it gets into the frozen state, in which case it will reload with the original URL instead of the dynamic CodePush one.

I would be inclined to consider that a big enough change in behaviour that it's not a patch-level release anymore

jacobg:
> > Does this commit work with CodePush using dynamic url?
> 
> I think it'll be fine, unless it gets into the frozen state, in which case it will reload with the original URL instead of the dynamic CodePush one.

Thank @dpogue. Do you suggest a different approach to this fix in Cordova 9?

dplewis:
I tried using this fix but it still freezes after a week of being in the background.

Its hard to debug as I don't know how to do crash reporting on a build from XCode.

I'm not sure how to replicate this as a week debug cycle is hard.

janpio:
Step 1: Create an issue instead of commenting here, this will get ignored and forgotten.

PR #450, opened 2018-11-02
UIWebView call loadRequest multiple times lead to deadlock in iOS12 (fixes #447)

### LINKED ISSUE #447: UIWebView call loadRequest multiple times lead to deadlock in iOS12
In iOS 12,if you call `loadRequest` twice in succession,like this:

```
[self.webView loadRequest:[NSURLRequest requestWithURL:[NSURL URLWithString:@"https://github.com"]]];
[self.webView loadRequest:[NSURLRequest requestWithURL:[NSURL URLWithString:@"https://google.com"]]];
```

`webViewDidStartLoad:` will callback twice times,but `webViewDidFinishLoad:` just callbcak once time.like this:

```
1. didStartLoading
2. didStartLoading
3. didFinishLoad
```

Everything is correct in iOS 11.Callback sequence:

```
1. didStartLoading
2. didFinishLoad
3. didStartLoading
4. didFinishLoad
```

This problem will lead to `[CDVUserAgentUtil acquireLock:]` block don`t excute.

crmo:
### Platforms affected
iOS

### What does this PR do?
Fix issue 447 UIWebView call loadRequest multiple times lead to deadlock in iOS12.
The number of callback of `webViewDidStartLoad`  and `webViewDidFinishLoad`  is not equal.`_loadCount` will never equals 1, delegate's `webViewDidFinishLoad` and `[CDVUserAgentUtil releaseLock:vc.userAgentLockToken];` will never call.Then new CDVViewcontroller  won't load page.

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVUIWebViewEngine/CDVUIWebViewDelegate.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVUIWebViewEngine/CDVUIWebViewDelegate.m
@@ -329,9 +329,14 @@ static NSString *stripFragment(NSString* url)
             break;
 
         case STATE_WAITING_FOR_LOAD_FINISH:
-            if (_loadCount == 1) {
+            // fix call loadRequest multiple times just callback webViewDidFinishLoad once time in iOS 12
+            if (@available(iOS 12.0, *)) {
                 fireCallback = YES;
-                _state = STATE_IDLE;
+            } else {
+                if (_loadCount == 1) {
+                    fireCallback = YES;
+                    _state = STATE_IDLE;
+                }
             }
             _loadCount -= 1;
             break;


CordovaLib/Classes/Private/Plugins/CDVUIWebViewEngine/CDVUIWebViewDelegate.m
```diff
@@ -329,9 +329,14 @@ - (void)webViewDidFinishLoad:(UIWebView*)webView
             break;
 
         case STATE_WAITING_FOR_LOAD_FINISH:
-            if (_loadCount == 1) {
+            // fix call loadRequest multiple times just callback webViewDidFinishLoad once time in iOS 12
+            if (@available(iOS 12.0, *)) {
```
janpio: What exactly does this line do? 
Does it check for "iOS 12.0"?

crmo: Yes,check if the system version is "12.0"

janpio: Doesn't that mean that is already broken again in 12.1?

sinno93:
I have the same problem in iOS12.0  .can these codes be merged?

CordovaLib/Classes/Private/Plugins/CDVUIWebViewEngine/CDVUIWebViewDelegate.m
shazron: Like @janpio said, you need to test for iOS 12 and greater. Use `IsAtLeastiOSVersion` https://github.com/apache/cordova-ios/blob/1cf92246d5463f519bab76e573078a76d58f9c68/CordovaLib/Classes/Public/CDVAvailability.h#L96

shazron: @crmo So this fix is just for 12.0 specifically, you mean?

jcesarmobile: The `@available` check includes newer versions that the one specified.

dpogue:
We were running into the same issue in one of our apps (symptom being that the splashscreen stayed up forever because the UIWebView never fired its loaded callback). @jeff2013 did some investigating and determined that the changes in this PR solve the issue.

So I'm going to approve this for merge and hopefully get it into cordova-ios@5.0.1.

Issue #477, opened 2018-12-18
CLI usage is not emitted to log


oliversalzburg:
None of command lines (like `xcodebuild`) that are invoked on the shell are logged. This is essential information when attempting to troubleshoot build issues and the information is completely lost.

At the very least, the essential `xcodebuild` commands should be logged.

janpio:
`--verbose` or `--debug` has no influence there?

oliversalzburg:
Correct. This is not logged at all. For example:

https://github.com/apache/cordova-ios/blob/004b8300456816cc614d229e85a35a8bc37d363d/bin/templates/scripts/cordova/lib/build.js#L174

dpogue:
Note: cordova-ios is currently using [its own `spawn` method](https://github.com/apache/cordova-ios/blob/004b8300456816cc614d229e85a35a8bc37d363d/bin/templates/scripts/cordova/lib/spawn.js), when it should probably be using [cordova-common's `superspawn`](https://github.com/apache/cordova-common/blob/master/src/superspawn.js)

oliversalzburg:
@dpogue I created https://github.com/apache/cordova-ios/pull/478 to resolve that

oliversalzburg:
Fixed by #479

PR #478, opened 2018-12-18
Use cross-spawn & shelljs instead of child-process

oliversalzburg:
### Platforms affected
iOS

### What does this PR do?
Applies `superspawn` and `shelljs` consistently. Both modules are used in other places in this module. The existing `spawn.js` in this project should be obsolete now. I preserved it just in case.

**Details by @brodybits:**

Change details:
* Use `cross-spawn` ("superspawn") instead of `child-process` spawn
* Use `shelljs` instead instead of `child-process` spawn in other places
* Add a comment that the `spawn` function provided by `spawn.js` is deprecated

Motivation: this change will allow us to spawn with `printCommand: true` option, to print the `xcodebuild` and other shell commands as proposed in #479.

~~Nothing at all.~~

- [x] Green Travis CI build

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -20,7 +20,7 @@
 var Q = require('q');
 var path = require('path');
 var shell = require('shelljs');
-var spawn = require('./spawn');
+var superspawn = require('cordova-common').superspawn;
 var fs = require('fs');
 var plist = require('plist');
 var util = require('util');
@@ -168,10 +168,10 @@ module.exports.run = function (buildOpts) {
             var buildOutputDir = path.join(projectPath, 'build', (buildOpts.device ? 'device' : 'emulator'));
 
             // remove the build/device folder before building
-            return spawn('rm', [ '-rf', buildOutputDir ], projectPath)
+            return superspawn.spawn('rm', [ '-rf', buildOutputDir ], { cwd: projectPath })
                 .then(function () {
                     var xcodebuildArgs = getXcodeBuildArgs(projectName, projectPath, configuration, buildOpts.device, buildOpts.buildFlag, emulatorTarget, buildOpts.automaticProvisioning);
-                    return spawn('xcodebuild', xcodebuildArgs, projectPath);
+                    return superspawn.spawn('xcodebuild', xcodebuildArgs, { cwd: projectPath });
                 });
 
         }).then(function () {
@@ -225,7 +225,7 @@ module.exports.run = function (buildOpts) {
 
             function packageArchive () {
                 var xcodearchiveArgs = getXcodeArchiveArgs(projectName, projectPath, buildOutputDir, exportOptionsPath, buildOpts.automaticProvisioning);
-                return spawn('xcodebuild', xcodearchiveArgs, projectPath);
+                return superspawn.spawn('xcodebuild', xcodearchiveArgs, { cwd: projectPath });
             }
 
             return Q.nfcall(fs.writeFile, exportOptionsPath, exportOptionsPlist, 'utf-8')
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -168,11 +168,10 @@ module.exports.run = function (buildOpts) {
             var buildOutputDir = path.join(projectPath, 'build', (buildOpts.device ? 'device' : 'emulator'));
 
             // remove the build/device folder before building
-            return superspawn.spawn('rm', [ '-rf', buildOutputDir ], { cwd: projectPath })
-                .then(function () {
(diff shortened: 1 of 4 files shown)


bin/templates/scripts/cordova/lib/build.js
```diff
@@ -168,11 +168,10 @@ module.exports.run = function (buildOpts) {
...
             // remove the build/device folder before building
-            return spawn('rm', [ '-rf', buildOutputDir ], projectPath)
-                .then(function () {
-                    var xcodebuildArgs = getXcodeBuildArgs(projectName, projectPath, configuration, buildOpts.device, buildOpts.buildFlag, emulatorTarget, buildOpts.automaticProvisioning);
-                    return spawn('xcodebuild', xcodebuildArgs, projectPath);
-                });
+            shell.rm('-rf', buildOutputDir);
+
+            var xcodebuildArgs = getXcodeBuildArgs(projectName, projectPath, configuration, buildOpts.device, buildOpts.buildFlag, emulatorTarget, buildOpts.automaticProvisioning);
+            return superspawn('xcodebuild', xcodebuildArgs, { cwd: projectPath });
```
dpogue: If you set `printCommand: true` in the options, it will log the xcodebuild command as a log level instead of a verbose level, which might be what we want to fix #477

bin/templates/scripts/cordova/lib/spawn.js
```diff
@@ -27,6 +27,7 @@ var proc = require('child_process');
  * @param  {String} opt_cwd       Working directory for command
  * @param  {String} opt_verbosity Verbosity level for command stdout output, "verbose" by default
  * @return {Promise}              Promise either fullfilled or rejected with error code
+ * @deprecated Use `require('cordova-common').superspawn` instead.
```
dpogue: I'd be inclined to just remove this outright in the next major, but I guess technically we don't know if anything else used it so we should go through one major cycle with it marked deprecated 🙁

bin/templates/scripts/cordova/lib/build.js
oliversalzburg: Awesome! Thanks for the note, I was about to manually implement that in another PR

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -87,15 +88,13 @@ module.exports.run = function (runOptions) {
...
-                        return spawn('rm', [ '-rf', appFile ], buildOutputDir)
-                            .then(function () {
-                                // move the platform/ios/build/device/Payload/appname.app to parent
-                                return spawn('mv', [ '-f', appFileInflated, buildOutputDir ], buildOutputDir);
-                            })
-                            .then(function () {
-                                // delete the platform/ios/build/device/Payload folder
-                                return spawn('rm', [ '-rf', payloadFolder ], buildOutputDir);
-                            });
+                        shell.rm('-rf', appFile);
```
brodybits: Just a quick comment that we have taken efforts to use `fs-extra` instead of `shelljs` in some other parts of the Cordova project. I would favor we do something similar in `cordova-ios` someday in the future (should not block Cordova 9 release).

bin/templates/scripts/cordova/lib/spawn.js
brodybits: I also would favor updating `spawn.js` to output a warning message that it is deprecated and will be removed in the future.

bin/templates/scripts/cordova/lib/run.js
oliversalzburg: Good call. I was mostly going for consistency here, which should make it easier to properly replace any usages in the future.

bin/templates/scripts/cordova/lib/spawn.js
oliversalzburg: I'm somewhat resistant to just `console.log` in this context. If we're assuming that someone is just going to `require()` this file, we can't assume other logging constructs to be in place. And logging directly to console may be more disturbing than helpful because it might break code that parses console output (which wouldn't be unreasonable with this code).

So I'd rather leave it as is, unless someone has a very specific suggestion on how to approach the issue.

brodybits:
Wonders of interactive editing:)

I made the terser title to get a "squash merge" commit message within 50 characters. I will probably use the terser title in the commit message and leave it up to you what to do with the title.

I just edited the description to add a more descriptive list of the actual changes, which I would like to use in the squash merge commit message. Please feel free to edit as needed.

I will probably merge it in the next 20-30 minutes or so. Thanks again!

bin/templates/scripts/cordova/lib/spawn.js
brodybits: Got it. I would like to raise to finish raising another PR to add the deprecation message, for separate discussion, before we resolving this conversation.

PR #480, opened 2018-12-18
superspawn/cross-spawn comment in build.js

brodybits:
to clarify that the recent change in 4694547 is using superspawn
from cordova-common, which in turn uses cross-spawn

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -20,7 +20,6 @@
 var Q = require('q');
 var path = require('path');
 var shell = require('shelljs');
-var superspawn = require('cordova-common').superspawn;
 var fs = require('fs');
 var plist = require('plist');
 var util = require('util');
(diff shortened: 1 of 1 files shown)


dpogue:
I'm not sure that this matters?
`superspawn` is part of cordova-common API, it's our spawning abstraction. Whether it uses `child_process.spawn` or `cross-spawn` internally shouldn't be a concern to downstream users. If cordova-ios cares that it's using `cross-spawn`, then there's a problem with the API.

brodybits:
Understood. I landed PR #478 using the term "cross-spawn", which is the change that would appear in RELEASENOTES.md, and forgot to explain the relationship with "superspawn" (which you know is what we use in the code). I was hoping to clarify it somehow. (I just updated the description to point to apache/cordova-common#50 where "superspawn" was updated to use `cross-spawn`.)

And from <https://github.com/apache/cordova/issues/16> I see a desire to remove "superspawn" in favor of a package such as [`execa`](https://github.com/sindresorhus/execa) over time.

How should we proceed?

dpogue:
I think we should just correct it in the curated changelog notes (if we even deem it important enough to need correcting)

brodybits:
> I think we should just correct it in the curated changelog notes

Whoever makes the release notes will probably be pretty busy cleaning it up for the major release, not sure if they will take much time to check what each change does in this level of detail.

We can close this one, probably too pedantic.

PR #485, opened 2018-12-19
Ensure test failure on unexpected promise resolution

oliversalzburg:
### Platforms affected
iOS

### What does this PR do?
It makes sure that the tests are failed correctly if the result is not as expected. Previously, these tests would simply time out after the Jasmine default timeout period. Now the tests will fail right away and give a better indication towards the reason.

I've ran the tests.

```shell
Started
....F

Failures:
1) versions get_tool_version method should find pod version.
  Message:
    Failed: expected promise resolution
  Stack:
    Error: expected promise resolution
        at versions.get_tool_version.then.catch (/Users/oliver/cordova-ios/tests/spec/component/versions.spec.js:61:37)
```

### DIFF
--- a/tests/spec/component/versions.spec.js
+++ b/tests/spec/component/versions.spec.js
@@ -25,38 +25,40 @@ if (process.platform === 'darwin') {
     describe('versions', function () {
         describe('get_tool_version method', () => {
             it('should not have found tool by name.', (done) => {
-                versions.get_tool_version('unknown').catch((error) => {
-                    expect(error).toContain('is not valid tool name');
-                    done();
-                });
+                versions.get_tool_version('unknown')
+                    .then(() => done.fail('expected promise rejection'))
+                    .catch((error) => {
+                        expect(error).toContain('is not valid tool name');
+                        done();
+                    });
             });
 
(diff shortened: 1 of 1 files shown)


tests/spec/component/versions.spec.js
```diff
@@ -25,38 +25,40 @@ if (process.platform === 'darwin') {
     describe('versions', function () {
         describe('get_tool_version method', () => {
             it('should not have found tool by name.', (done) => {
-                versions.get_tool_version('unknown').catch((error) => {
-                    expect(error).toContain('is not valid tool name');
-                    done();
-                });
+                versions.get_tool_version('unknown')
+                    .then(() => done(new Error('expected promise rejection')))
```
brodybits: I would favor using done.fail('...') instead of done(new Error(...)) throughout this test script, like this:

```suggestion
                    .then(() => done.fail('expected promise rejection'))
```

oliversalzburg: Didn't know about that syntax. Let me give it a try...

oliversalzburg: Looks good. Thanks!

Issue #486, opened 2018-12-23
Cordova run does not work


knight9999:
When doing `cordova run`,  the following errors happened.

```
Running command: unzip -o -qq /Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/platforms/ios/build/device/HelloCordova.ipa
shell.js: internal error
Error: ENOTDIR: not a directory, rename '/Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/platforms/ios/build/device/Payload/HelloCordova.app' -> '/Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/platforms/ios/build/device/HelloCordova.app'
    at Object.fs.renameSync (fs.js:734:3)
    at /Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/node_modules/shelljs/src/mv.js:77:8
    at Array.forEach (<anonymous>)
    at Object._mv (/Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/node_modules/shelljs/src/mv.js:53:11)
    at Object.mv (/Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/node_modules/shelljs/src/common.js:186:23)
    at /Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/platforms/ios/cordova/lib/run.js:93:31
    at _fulfilled (/Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/node_modules/q/q.js:854:54)
    at /Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/node_modules/q/q.js:883:30
    at Promise.promise.promiseDispatch (/Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/node_modules/q/q.js:816:13)
    at /Users/knaito/Documents/work/cordova-develop/coho/xcode10Test/node_modules/q/q.js:624:44
```
`shelljs.rm` introduced by the commit/4694547bcc360c4c68331285593e69aae7232752 can not remove symbolic link `Application.app` (`HelloCordova.app` in this case) due to  the specification of `shelljs` itself. Therefore cordova/lib/run.js can not remove the symbolic link `Application.app` (`HelloCordova.app`) in cordova/build/device (or cordova/build/emulator) and the errors happen.

If we modify
```
shell.rm('-rf', appFile);
```
by 
```
fs = require('fs-extra');
fs.removeSync(appFile);
```
[...]

PR #487, opened 2018-12-24
Fix issue #486. Distinguish xcodebuild `build` command and `archive` command.

knight9999:
distinguish `xcodebuild build` and `xcodebuild archive` according to `cordova build` and `cordova run`

### Platforms affected
cordova-ios 

### What does this PR do?
Resolve this the #486

1. removing `shelljs` in cordova/lib/run.js.  This is because `shelljs.rm` can not remove symbolic link.
2. using `xcodebuild build` at `cordova run --device`. Because `ios-deploy` can not work normally with `.app` pacakge obtained by`xcodebuild archive` and `xcodebuild -exportArchive` command.  
3. new option `--nobuild-device` is added. This is force skip build process evenwhen `cordova run --device`.

In this PR, `build.js` is called from `run.js` with the build-action `build`. This is because we should use internal `xcodebuild build` command to get the `.app` package instead of `xcodebuild archive` command with Xcode 10. 

Before this PR,
`cordova run --device` runs as follows.

1. `cordova-lib` calls `build.js` 
2. `cordova-lib` calls `run.js` with `--nobuild` option.  => But `ios-deploy` does not work normally.

After this PR,
`cordova run --device` runs as follows.

1. `cordova-lib` calls `build.js` (This build process is unnecessary. Because the `ios-deploy` does not work normally with the obtained .app package ) 
2. `cordova-lib` calls `run.js` with `--nobuild` option. => `run.js` calls `build.js` with option `buildFlag: CONFIGURATION_BUILD_DIR=[projectPath]/build/device-run` and `buildAction: build`.
 
Although this PR works with the latest cordova-lib, there is one unnecessary build process.
To remove unnecessary build process, we should update `cordova-lib`. I will send new PR for cordova-lib.

I checked as follows with both Xcode9 and Xcode10 envrionment.

```
$ npx cordova@nightly create xcode10Test com.foo.bar
$ cd xcode10Test
$ vi build.json
(set the certificate and provisioning file, development for debug build and ad-hoc for release build)
$ npx cordova@nightly platform add github:knight9999/cordova-ios#fix_cordova_run
$ npx cordova@nightly run --emulator --debug 
(In some environment, you should specify the target like
`--target="iPhone-8, 11.3"`.  The available targets are shown by `npx cordova@nightly run --list` This is because the selecting default simulator logic is different between build.js and run.js. This is another issue we should solve. but is out of this PR.) 
(confirm the app on emulator)
$ npx cordova@nightly run --device --debug
(confirm the app on device)
$ npx cordova@nightly run --device --release
(confirm the app on device)
$ npx cordova@nightly run --device --release --nobuild --nobuild-device
(This skip build processes. confirm the app on device)
$ npx cordova@nightly build --device
(install .ipa file and confirm the app on device)
$ npx cordova@nightly build --release
(install .ipa file and confirm the app on device)
```

### Checklist

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -170,7 +170,7 @@ module.exports.run = function (buildOpts) {
             // remove the build/device folder before building
             shell.rm('-rf', buildOutputDir);
 
-            var xcodebuildArgs = getXcodeBuildArgs(projectName, projectPath, configuration, buildOpts.device, buildOpts.buildFlag, emulatorTarget, buildOpts.automaticProvisioning);
+            var xcodebuildArgs = getXcodeBuildArgs(projectName, projectPath, configuration, buildOpts.device, buildOpts.buildFlag, emulatorTarget, buildOpts.automaticProvisioning, buildOpts.buildAction);
             return superspawn.spawn('xcodebuild', xcodebuildArgs, { cwd: projectPath, printCommand: true, stdio: 'inherit' });
 
         }).then(function () {
@@ -269,7 +269,7 @@ module.exports.findXCodeProjectIn = findXCodeProjectIn;
  * @param  {Boolean} autoProvisioning   Whether to allow Xcode to automatically update provisioning
  * @return {Array}                  Array of arguments that could be passed directly to spawn method
  */
-function getXcodeBuildArgs (projectName, projectPath, configuration, isDevice, buildFlags, emulatorTarget, autoProvisioning) {
+function getXcodeBuildArgs (projectName, projectPath, configuration, isDevice, buildFlags, emulatorTarget, autoProvisioning, buildAction) {
     var xcodebuildArgs;
     var options;
     var buildActions;
@@ -295,7 +295,7 @@ function getXcodeBuildArgs (projectName, projectPath, configuration, isDevice, b
             '-destination', customArgs.destination || 'generic/platform=iOS',
             '-archivePath', customArgs.archivePath || projectName + '.xcarchive'
         ];
-        buildActions = [ 'archive' ];
+        buildActions = buildAction ? [ buildAction ] : [ 'archive' ];
         settings = [
             customArgs.configuration_build_dir || 'CONFIGURATION_BUILD_DIR=' + path.join(projectPath, 'build', 'device'),
             customArgs.shared_precomps_dir || 'SHARED_PRECOMPS_DIR=' + path.join(projectPath, 'build', 'sharedpch')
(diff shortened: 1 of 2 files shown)


knight9999:
Thanks for the valuable comment @brodybits .
I have updated my PR as you mentioned.

knight9999:
@brodybits I have already updated my PR and now my PR passes all CI Checks.
Would you please re-review my PR?

shazron:
I'm having trouble at the deploy stage since the `.app` is not found (you removed all the code that unzips the .ipa)
```
cordova build --developmentTeam=XXXX --device
cordova build --developmentTeam=XXXX --nobuild-device
```

dpogue:
I am still having trouble with Xcode 10 not making valid builds, which means the generated .ipa file is wrong. I have a suspicion that the invalid .ipa is why unzipping it and running the .app doesn't work.

I'd slightly prefer to try to fix that underlying Xcode 10 build issue before we merge this, because I feel like this PR is fixing a symptom of the problem without fixing the root of the problem.

knight9999:
@shazron Thanks for the comment.
You means `cordova run` or `cordova build`?
I removed all the code that unzips the .ipa in `lib/run.js`. This is because unzipped .app file does not work in ios-deploy. (As you pointed out in https://github.com/ios-control/ios-deploy/issues/364, it stalls after showing splashscreen.) 
Instead of unzipping .ipa file, I use `xcodebuild build` command to get `.app` files in `lib/run.js`. (where calling lib/build.js internally)
The resulting .app files are in `platforms/ios/build/device-run` instead of `platforms/ios/build/device`.

knight9999:
@dpogue Thanks for the comment.
> I am still having trouble with Xcode 10 not making valid builds, which means the generated .ipa file is wrong. I have a suspicion that the invalid .ipa is why unzipping it and running the .app doesn't work.
I feel this is a root of this problem, just as you said.

This may be the Xcode 10's bug. Or Apple chagnes the build specification for some reason. (security reason?)

This PR is dirty in this meaning because of inconsistent way of `cordova build` and `cordova run`.
But for now, this is only the solution I found.

shazron:
The ios-deploy issue is a red-herring. Turns out the .app wasn't built with the `--buildFlag='-UseModernBuildSystem=0'` flag. Once I did that, it could deploy fine.

knight9999:
@shazron Thanks you for the great work https://github.com/apache/cordova-ios/pull/494 .
This resolves the issue about ios-deploy (in 'cordova run' command) when using explicit provisioning. 
Although there is an issue with automatic provisioning case, I close this PR.

Issue #407, opened 2018-09-14
[ANNOUNCEMENT] Status of Xcode 10 support


dpogue:
A few issues have started to come in regarding the state of Cordova projects on Xcode 10. This is a rough summary of the situation:

Xcode 10 uses a new build system by default (previously available on an opt-in basis in Xcode 9). The cordova-ios project structure is not compatible with this new build system and results in failures.
Officially, we do not claim to support Xcode 10.

Currently the best workaround is to opt-out of the new build system:

* If you're building on the command-line, you can specify `--buildFlag="-UseModernBuildSystem=0"`:
   ```
   # Cordova CLI
   cordova run ios --buildFlag='-UseModernBuildSystem=0'
   cordova build ios --buildFlag='-UseModernBuildSystem=0'
   
   # Ionic CLI
   ionic cordova run ios -- --buildFlag="-UseModernBuildSystem=0"
   ionic cordova build ios -- --buildFlag="-UseModernBuildSystem=0"
   ```

* If you're building with a build.json config file, you can add the following under the iOS release or debug config:
   ```json
   "buildFlag": [
     "-UseModernBuildSystem=0"
   ]
   ```

* If you are opening the project in the Xcode IDE, you need to change the build system in Workspace Settings to "Legacy Build System"

  <details><summary>Xcode example</summary>
  <img src="https://user-images.githubusercontent.com/781366/45737593-f7e30780-bbf6-11e8-86f8-207dd596ad36.png" alt="Workspace Settings is under the File menu">
  <img src="https://user-images.githubusercontent.com/781366/45737603-fc0f2500-bbf6-11e8-84c9-8dcb17e7c691.png" alt="Change the Build System option to Legacy Build System">

  Thanks [**kitolog**](https://github.com/apache/cordova-ios/issues/412#issuecomment-422689481) for the screenshots!
  </details>

We're going to investigate what's required to make cordova-ios compatible with the new build system, and hope to include that in the next major version (cordova-ios@5.0.0).

sambegin:
Just note here that the build.json file should look like this : 

```
  "ios": {
    "release": {
      "buildFlag": [
        "-UseModernBuildSystem=0"
      ]
    }
  }
```

not 
```
  "iOS": {
    "release": {
      "buildFlag": [
        "-UseModernBuildSystem=0"
      ]
    }
  }
```

like "mentioned" in the previous comment ;p

robjs827:
Thanks, @dpogue 
[...]

PR #386, opened 2018-08-01
WIP: README.md drop support for Xcode pre-9.0 - WIP FOR DISCUSSION

brodybits:
### Platforms affected

iOS

### What does this PR do?

- Drop support for Xcode pre-9.0 in README.md
- add notice that Xcode 10 (beta) is not supported, with known issue ref:
  - <https://issues.apache.org/jira/browse/CB-14192>
  - apache/cordova-docs#862
  - #378 (proposed workaround solution for Xcode 10)

TODO:
- [ ] needs discussion
- [ ] raise JIRA issue if agreed
- [ ] update in cordova-docs
- [ ] check if we should drop support for Xcode pre-9.0 in `bin` and `tests`?

### What testing has been done on this change?

TODO

### Checklist

TODO:

- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- ~~Added automated test coverage as appropriate for this change.~~

PR #491, opened 2019-01-08
[WIP] Update Xcode version supported (9.x/10.x/...)

brodybits:
Continuation from PR #386: show minimum Xcode version as 9.x (deprecated) or 10.

This is assuming that the issues with Xcode 10 discussed in #407 are resolved by the time we make the new release of cordova-ios.

~~P.S. Linking to apache/cordova-docs#863: Minimum Xcode version in cordova-docs & cordova-ios not consistent~~ (sorry apache/cordova-docs#863 was an old issue, will raise a new issue to track this one)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Cordova iOS is an iOS application library that allows for Cordova-based projects
 
 Requires:
 
-* Xcode 8.x or greater. Download it at [http://developer.apple.com/downloads](http://developer.apple.com/downloads) or the [Mac App Store](http://itunes.apple.com/us/app/xcode/id497799835?mt=12).
+* Xcode 9.x (deprecated), 10, or greater. Download it at [http://developer.apple.com/downloads](http://developer.apple.com/downloads) or the [Mac App Store](http://itunes.apple.com/us/app/xcode/id497799835?mt=12).
 * [node.js](https://nodejs.org)
 
 Create a Cordova project


dpogue:
I wonder if we want to say we only officially support Xcode 10 (which will be required for app store submissions as of March 2019), but unofficially keep the project files working in Xcode 9 for at least the first release of the next major?

I don't especially want us to get into a situation where Xcode 9 is holding us back from fixing bugs, but I don't want to outright drop it while people are still using it. I don't really think supporting both will be a problem, but that could change with future Xcode (11, 12) versions.

If we don't explicitly claim to support Xcode 9, then we wouldn't have any obligation to do a major version bump if/when we break Xcode 9.

brodybits:
If we want to drop Xcode 9 support, which I am starting to favor, then I would favor that we do something small that we know would break the build on Xcode 9.

I can imagine an Xcode 9 user doing an upgrade without reading the notes, and then someday boom!

I think it should be no problem to support both Xcode 9 and Xcode 10 in 4.x.x.

Another alternative could be to release cordova-ios@5 to drop Node.js 4 support, and then cordova-ios@6 to drop Xcode 9 support, before we finish releasing Cordova 9.

brodybits:
I just marked this as a bug, linked to apache/cordova-docs#863 (minimum Xcode version in cordova-docs & cordova-ios not consistent).

dpogue:
> I think it should be no problem to support both Xcode 9 and Xcode 10 in 4.x.x.

Do you mean 5.x.x? I think it **will** be a problem to support Xcode 10 in 4.x.x (it currently doesn't work, and I don't think we can fix that without breaking Xcode 8 (which is supported by 4.x.x))

brodybits:
> > I think it should be no problem to support both Xcode 9 and Xcode 10 in 4.x.x.
> 
> Do you mean 5.x.x?

I would favor supporting both Xcode 9 and Xcode 10 in 5.x.x.

I actually did not encounter any issues when testing my own work with cordova-ios@4.5.x and recent Xcode 10 upgrade. But yes I would favor that we retire cordova-ios@4 soon if possible.

I will raise a new issue to keep this consistent between cordova-ios and cordova-docs.

Issue #523, opened 2019-02-13
With cordova-ios 5.0.0, phonegap-plugin-push cannot be added


martispaul:
# Bug Report

## Problem

With cordova-ios 5.0.0, phonegap-plugin-push cannot be added

### What is expected to happen?

Created new cordova ios app with ios platform 5.0.0, adding phonegap-plugin-push should be successfully. 

### What does actually happen?

Adding the phonegap-plugin-push fails on a new created app (with cordova-ios@5.0.0).
On the older cordova-ios version (4.5.4), adding the plugin was successfully. 

## Information

Using windows PC, create a blank cordova app
Add platform ios@5.0.0
Add plugin phonegap-plugin-push@2.2.3

C:\Users\martispa\Desktop\TestIOS\hello>**cordova platform add ios@5.0.0**
Using cordova-fetch for cordova-ios@5.0.0
Adding ios project...
Creating Cordova project for the iOS platform:
        Path: platforms\ios
        Package: com.example.hello
        Name: HelloWorld
iOS project created with cordova-ios@5.0.0
Installing "cordova-plugin-whitelist" for ios
--save flag or autosave detected
Saving ios@~5.0.0 into config.xml file ...

C:\Users\martispa\Desktop\TestIOS\hello>**cordova plugin add phonegap-plugin-push@2.2.3**
Installing "phonegap-plugin-push" for ios
"framework" tag with type "podspec" is deprecated and will be removed. Please use the "podspec" tag.
Failed to install 'phonegap-plugin-push': Error: ENOENT: no such file or directory, open 'C:\Users\martispa\Desktop\TestIOS\hello\platforms\ios\Pods\Pods.xcodeproj\project.pbxproj'
    at Object.fs.openSync (fs.js:646:18)
    at Object.fs.readFileSync (fs.js:551:33)
    at pbxProject.parseSync (C:\Users\martispa\Desktop\TestIOS\hello\node_modules\xcode\lib\pbxProject.js:61:28)
    at Api.setSwiftVersionForCocoaPodsLibraries (C:\Users\martispa\Desktop\TestIOS\hello\platforms\ios\cordova\Api.js:587:18)
    at C:\Users\martispa\Desktop\TestIOS\hello\platforms\ios\cordova\Api.js:448:26
    at _fulfilled (C:\Users\martispa\Desktop\TestIOS\hello\node_modules\q\q.js:854:54)
[...]

PR #528, opened 2019-02-14
GH-523 : ignore swift version settings of cocoapods libraries on windows

knight9999:
### Platforms affected
cordova-ios

### Motivation and Context
Fixing the issue #523 

### Description
To avoid error on windows, adding checks cocoapods is available or not before changing swift version of cocoapods libraries.

### Testing
In my local Mac,
create project and adding phonegap-plugin-push.
Checking build.
Checking the application can receive push notifications.
(by using cordova@nightly)

In my local Windows 10,
create project and adding phonegap-plugin-push.
Checking to skip the cocoapods specific process.
(by using cordova@8.1.0)

### Checklist

- [x] Commit is prefixed with `ios` if this change only applies to one platform (e.g. `(android)`)

### DIFF
--- a/bin/templates/scripts/cordova/Api.js
+++ b/bin/templates/scripts/cordova/Api.js
@@ -445,7 +445,7 @@ Api.prototype.addPodSpecs = function (plugin, podSpecs, frameworkPods) {
 
             return podfileFile.install(check_reqs.check_cocoapods)
                 .then(function () {
-                    self.setSwiftVersionForCocoaPodsLibraries(podsjsonFile);
+                    return self.setSwiftVersionForCocoaPodsLibraries(podsjsonFile);
                 });
         } else {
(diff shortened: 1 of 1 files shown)


janpio:
Awesome, super quick response @knight9999! 

Just for my understanding: When is `pod install` and the code that you guarded in this PR run for the project then? First time being used on macOS?

knight9999:
@janpio I simply added `check_reqs.check_cocoapods()` to cocoapod related function as same as the case of `pod install` in cordova 4.5.*.

`check_reqs.check_cocoapods` checks os type, cocoapods is installed or not, and cocoapods version is suitable or not. It shows appropriate warning message if necessary.

janpio:
Ok, so when I add some plugin on Windows and commit, then check out on a Mac and keep working there it will still run `pod install` and do all the things sooner or later? No way to break the project?

PR #531, opened 2019-02-15
Compress assets in Images.xcasset (tinypng.com)

janpio:
### Motivation and Context

Projects include a lot of assets that are quite big. Can they be made smaller?

### Description

Before: 1.07MB
After: 401kB

(used https://www.npmjs.com/package/tinypng-cli, `tinypng -r -k ...`)

### Testing

none

We should make _really_ sure the images aren't made worse.

### Checklist


dpogue:
:astonished: What did tinypng do to these to shrink them down?

I'd run them through `pngquant --strip --speed 1` and then `zopflipng` on the command-line as part of generating them, but I'd love to know what exactly they were using to accomplish this greater reduction in filesize

janpio:
From https://tinypng.com/

> What does TinyPNG do?
>
> TinyPNG uses smart lossy compression techniques to reduce the file size of your PNG files. By selectively decreasing the number of colors in the image, fewer bytes are required to store the data. The effect is nearly invisible but it makes a very large difference in file size!
> 
> When you upload a PNG (Portable Network Graphics) file, similar colors in your image are combined. This technique is called “quantization”. By reducing the number of colors, 24-bit PNG files can be converted to much smaller 8-bit indexed color images. All unnecessary metadata is stripped too. The result better PNG files with 100% support for transparency. Have your cake and eat it too!

---

Usually I would not go for "smart lossless" compression, but as those assets are not actually meant to be shipped to end customers - those are Cordova branded assets after all - I think this is fine. And I couldn't see any difference to be honest 🖼

Issue #538, opened 2019-02-19
Cordova overwrites bundleIdentifier for all existing targets


wolfmanfx:
# Bug Report
When cordova (using 5.0.0) projects contains watch targets (watchkit, watch extension) which have their own bundle id setup correctly, cordova overwrites all targets with the same id.

## Problem
Its not possible to pack new targets in the project and use cordova cli build 
cordova build --platform ios --device --release

### What is expected to happen?
It should not overwrite the PRODUCT_BUNDLE_IDENTIFIER of all targets.

### What does actually happen?
Signing does not work anymore.

- [ x] I searched for existing GitHub issues
- [ x] I updated all Cordova tooling to most recent version
- [x ] I included all the necessary information above

---

Updates by @brodybits:

This issue seems to also break functionality of plugins that do extensions to Xcode projects by using hooks. See below for discussions.

janpio:
So did I understand correctly that using `cordova build ios --device --release` overwrite the bundle identifiers of other targets? 

Can you pinpoint which files are changed to do this (quickly create a temp git project in `platform/ios` with `git init`, commit everything, then do the build, then look at the changed files to exactly pinpoint what files got edited)?

wolfmanfx:
Hi, 
I create the project with cordova (I mean the xcode project) than I am patching the project.pbxproj and add 2 targets (watchapp, watchapp extension) when I stay within xcode everything is correct. But we build within travis so I am using the cli to create and export the ipa - but the signing process fails.

When I open the project.pbxproj the PRODUCT_BUNDLE_IDENTIFIER for all 3 targets are different and correct when I run 
cordova build --platform ios --device --release
All PRODUCT_BUNDLE_IDENTIFIER within the project.pbxproj are overwritten which breaks all other targets.

Can you point to the code where the updates happen?

dpogue:
That happens [here](https://github.com/apache/cordova-ios/blob/5e688ca4ddf698fad5d10185f7583dec49abcfe1/bin/templates/scripts/cordova/lib/prepare.js#L301-L304): https://github.com/apache/cordova-ios/blob/5e688ca4ddf698fad5d10185f7583dec49abcfe1/bin/templates/scripts/cordova/lib/prepare.js#L301-L304

wolfmanfx:
Ok - 
pbxProject.prototype.updateBuildProperty = function(prop, value, build) calls 
    var configs = this.pbxXCBuildConfigurationSection();
[...]

PR #544, opened 2019-02-21
Fix https://github.com/apache/cordova-ios/issues/538 now we update on…

wolfmanfx:
…ly the app target

### Platforms affected
iOS

### Motivation and Context
https://github.com/apache/cordova-ios/issues/538

### Description
Now we retrieve the correct target and update only these properties without touching other targets.

### Testing
I have tested unit tests and the creation of standard apps + the more complex targets where I have added apple watch targets (via script using apchae xcode)

### Checklist

- [ X] I've run the tests to see all new and existing tests pass
- [ x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))

### DIFF
--- a/bin/templates/scripts/cordova/build
+++ b/bin/templates/scripts/cordova/build
@@ -19,21 +19,21 @@
        under the License.
 */
 
-var args  = process.argv;
+var args = process.argv;
 var Api = require('./Api');
 var nopt = require('nopt');
 var path = require('path');
 
 // Support basic help commands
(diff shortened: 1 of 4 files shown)


dpogue:
This is incredibly hard to review because of all the formatting changes, and it's going to fail tests because those changes break our linting rules.

Can you re-submit with just the additions and no formatting changes?

msari-ipe-ext-1:
Yes I will do another PR the auto format kicked in 🙄

PR #545, opened 2019-02-22
Do not overwrite all targets with same id

wolfmanfx:
* Prepare script update build properties for the main target only. Wi…th the old behavior it were not possible to add apple watch targets and build it because cordova bricked the bundleids for these targets.

* Add new build parameter multipleProvisioningProfiles. Here the usage within the build.json:

            "multipleProvisioningProfiles": [
                { "key": "de.demo.html5", "value": "DemoME-Enterprise" },
                { "key": "de.demo.html5.watchkitapp", "value": "DemoWE-Enterprise" },
                { "key": "de.demo.html5.watchkitapp.watchkitextension", "value": "DemoWE Extension-Enterprise" }
            ],

This one is needed too otherwise the signing step fails because cordova generates a wrong bundleid to prov.profile mapping. With these settings the exportOptions.plist is written correctly for multiple targets.
This patch is needed to make the apple watch targets work with the cordova-ios workflow (apple watch targets added within a post process scripts using npm xcode)

### Platforms affected
iOS

### Motivation and Context
When cordova (using 5.0.0) projects contains watch targets (watchkit, watch extension) which have their own bundle id setup correctly, cordova overwrites all targets with the same id.
So the motivation is clear to have an hybrid app as watchos host without maintaining a manual created xcode project.

### Description
See commit message

### Testing
Tested against a real project with an watchos app.

### Checklist


wolfmanfx:
Link #538

bin/templates/scripts/cordova/build
```diff
@@ -42,6 +42,7 @@ var buildOpts = nopt({
     'codeSignIdentity': String,
     'codeSignResourceRules': String,
     'provisioningProfile': String,
+    'multipleProvisioningProfiles': Array,
```
dpogue: I wonder if we should add a new property here for `multipleProvisioningProfiles` vs changing the existing `provisioningProfile` property to accept either a single string or an array.

bin/templates/scripts/cordova/lib/prepare.js
```diff
@@ -275,12 +275,59 @@ function handleOrientationSettings (platformConfig, infoPlist) {
     }
 }
 
+// Make sure only update properties from our target project
+function updateBuildPropertyLocal(proj, displayName, prop, value, build) {
+    try {
+        // Check if we have a valid target - during prepare we do not have it
+        var target = proj.pbxTargetByName(displayName);
```
dpogue: Does this definitely use the display name, and not the full name? In Cordova apps those are usually the same, but it's not guaranteed (and we provide a way to change the display name).

msari-ipe-ext-1: Can you give me an example how to change the name - I will test that case.

bin/templates/scripts/cordova/build
msari-ipe-ext-1: I guess thats a better approach - i will change that.

bin/templates/scripts/cordova/lib/prepare.js
dpogue: In config.xml:

```xml
<name short="MyApp">My Full Application Name</name>
```

In that example, `"MyApp"` would be the short name (which is what you're putting into the `displayName` variable).

wolfmanfx: Yes you were right - I have fixed the pbx target retrieval (tried both variants)

bin/templates/scripts/cordova/build
wolfmanfx: I have to revert the idea - merging the two options do not work and require more refactoring. So I suggest to add this additional parameter which is only used for the export plist.

msari-ipe-ext-1:
So we are using the fork on a day by day basis everythinh working (with the embeded watch app).
I would love to switch back to the main repo anytime :)

janpio:
Maybe create a companion PR that adds some documentation on that at https://github.com/apache/cordova-docs/blob/master/www/docs/en/dev/guide/platforms/ios/index.md#signing-an-app (I guess?).

NiklasMerz:
What's the state of this PR? I am encountering this issue as well.

timbru31:
> What's the state of this PR? I am encountering this issue as well.
> 
> This PR solved the issue.

A code review or or confirmation that this PR solves the issue would help us to merge this faster :)

msari-ipe-ext-1:
The flag overwrites the correct profiles for the given bundle ids - I can build my app within the console without to open xcode. The value is name of the provision profile.
It does make sense to split the PR because no CI would able to build apps without manual editing which is not my usecase (must be builded by travis)

dpogue:
The more I think about this, the more I feel like this is probably outside the scope of what Cordova should be handling. 

I'm hesitant to add this for several reasons:

1. We know that provisioning profiles are already a source of a lot of frustration for Cordova users, and a source of a lot of problems every time a new Xcode version is released. We're trying to encourage everyone to move to automatic signing instead of specifying provisioning profiles manually. Adding this support makes the documentation more confusing and seems to encourage the practice we're trying to move away from.

2. As you can see from the ongoing discussions about the future of UIWebView and WKWebView, there's a real shortage of people able to contribute time to maintaining and upgrading Cordova. Adding this support might seem simple, but it's yet more code that needs to be maintained. This code has no tests, its use-case is far from the standard usage, and there's a danger that every cleanup effort anyone takes in the future will break it in non-obvious ways.

3. The purpose of Cordova is ostensibly to package web applications up in a native app wrapper. Yes, it supports native code plugins, but those are typically meant to expose a JS interface to the web application.
  
    If you're at a point where you have multiple targets in Xcode and are juggling multiple provisioning profiles, you should probably be managing the Xcode project yourself (either based on the one generated by Cordova or using a similar tool like [Capacitor](https://capacitor.ionicframework.com/)).

NiklasMerz:
@dpogue I can agree you with some points but have to disagree with others.

I am with you that this change should be minimal to avoid breaking stuff and keep maintainance cost low. That's why I would suggest to simplify this PR and remove the provisioning profile stuff. I am happy to do and/or test this. Is this viable?

But I would strongly argue to fix the core issue "Do not overwrite bundle ids". Cordova iOS 5 does not work for our app but 4.5.5 did before these changes.

You are probably right that at some point projects with multiple targets grow out of the original scope of Cordova. But this can happen organically. Our app now has two plugins which add extensions, because we need those features. Switching to Capacitor is not an option at this point because of other design choices.

Managing profiles and signing manually in Xcode is fine for me, but everything else should work with cordova build. That's why I want to fix the bundleid issue to make upgrading to cordova-ios 5 possible.

msari-ipe-ext-1:
I see your points - but overwriting existing profiles is a bug, I have combined it with a feature to automate profile settings.
Supporting iOS extensions would be great but I also see that the cordova repos (including plugins) in general are loosing ground (maintainers).
So Capacitor is also or midterm target - we have wasted a lot of time with broken CI builds or forking unmaintained plugins to make it work.

You could split this PR and extract the bug fix and remove the feature.

best regards Murat

NiklasMerz:
Thanks Murat. I would like to see this bug fixed. I took you branch and removed the `multipleProvisioningProfiles` part. Now this change looks good to me for default Cordova projects with no extensions. I will do some tests and update #708 accordingly.

NiklasMerz:
Closing this because of the reasons above and many conflicts. We need to address this differently. We should discuss more in the issue when time is ready.

HarelM:
@NiklasMerz @msari-ipe-ext-1 is the multiple provisioning profile available in latest release of cordova-ios? I see that when using the fovea's openwith share extension the `exportOptions.plist` is create with a single provisioning profile of the shared extention...
Any chance to update the docs of iOS into how to use this...?
https://cordova.apache.org/docs/en/latest/guide/platforms/ios/index.html

PR #546, opened 2019-02-22
(ios) Fix setting of target-device to handset in combination with plugins and resource files

skocks:
### Platforms affected
IOs

### Motivation and Context
when:
- setting the target device `<preference name="target-device" value="handset" />`
- using (any) plugins
- Adding a resource file `<resource-file src="GoogleService-Info.plist" />`

then:
`TARGETED_DEVICE_FAMILY` is reset to the default value of "1,2" while adding the resource file after handling the target device.

### Description
Instead of creating a new instance of the xcode project via `new xcode.project([...])` use the wrapped xcode project via `projectFile.parse([...])`.
This will reuse the cached xcode project, which is created when adding the plugins, and therefore maintain the changes when setting the target device.
The cached xcode project will then also be used when adding the resource file.

### Testing
- add plugins
- set target-device to handheld
- add a resource file

verify, that `TARGETED_DEVICE_FAMILY = 1;` is written to the project.pbxproj file.

### Checklist

- [X] I added automated test coverage as appropriate for this change
- ~~If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))~~
- ~~I've updated the documentation if necessary~~

### DIFF
--- a/bin/templates/scripts/cordova/lib/prepare.js
+++ b/bin/templates/scripts/cordova/lib/prepare.js
@@ -22,7 +22,6 @@ var Q = require('q');
 var fs = require('fs');
 var path = require('path');
 var shell = require('shelljs');
-var xcode = require('xcode');
 var unorm = require('unorm');
 var plist = require('plist');
 var URL = require('url');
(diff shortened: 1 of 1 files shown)


janpio:
What issue does this PR refer to? You checked that you "linked to the issue in the text above" but I couldn't find an issue ID.

skocks:
> 
> 
> What issue does this PR refer to? You checked that you "linked to the issue in the text above" but I couldn't find an issue ID.

That was a misunderstanding.
Since it reads "If this Pull Request resolves an issue, [...]" I checked it, because no issue is linked and therefore not referenced. - I can uncheck or delete the passage.

PR #577, opened 2019-03-20
Don't write spec to podfile when it's an empty string

piotr-cz:
### Platforms affected
iOS

### Motivation and Context

When plugin framework tag contains an empty spec value (`spec=""`), Podfile becomes malformed and `pod install` command fails with
```
ArgumentError - Illformed requirement `""`
```

Example:
```xml
    <framework src="Fabric" type="podspec" spec="~> 1.7.6"/>
    <framework src="Crashlytics" type="podspec" spec="~> 3.10.1"/>
    <framework src="Firebase/Core" type="podspec" spec=""/>
```

Creates
```
	pod 'Fabric', '~> 1.7.6'
	pod 'Crashlytics', '~> 3.10.1'
	pod 'Firebase/Core', ''
```

Proposed change checks for an empty spec value and doesn't write version number:

```
	pod 'Fabric', '~> 1.7.6'
	pod 'Crashlytics', '~> 3.10.1'
	pod 'Firebase/Core'
```

### Description

I'm not sure if it's a bug in cordova-ios or specific plugin ([cordova-plugin-firebase-crashlytics](https://github.com/ReallySmallSoftware/cordova-plugin-firebase-crashlytics)).
However cordova documentation for the [framework tag](https://cordova.apache.org/docs/en/latest/plugin_ref/spec.html#framework) doesn't specify if the spec attribute can be omitted so it seems valid to me to use empty value.

### Testing

- Install cordova-plugin-firebase-crashlytics plugin in verbose mode:
  `cordova plugin add cordova-plugin-firebase-crashlytics --verbose`
- Installation doesn't fail

### Checklist

### DIFF
--- a/bin/templates/scripts/cordova/lib/Podfile.js
+++ b/bin/templates/scripts/cordova/lib/Podfile.js
@@ -313,7 +313,7 @@ Podfile.prototype.write = function () {
             }
         } else {
             var list = ['\'' + name + '\''];
-            if ('spec' in json) {
+            if ('spec' in json  && json.spec.length) {
                 list.push('\'' + json.spec + '\'');
             }
 
--- a/bin/templates/scripts/cordova/lib/Podfile.js
+++ b/bin/templates/scripts/cordova/lib/Podfile.js
@@ -313,7 +313,7 @@ Podfile.prototype.write = function () {
             }
         } else {
             var list = ['\'' + name + '\''];
-            if ('spec' in json  && json.spec.length) {
+            if ('spec' in json && json.spec.length) {
                 list.push('\'' + json.spec + '\'');
             }
 


piotr-cz:
Travis test fails only on Xcode 10.1 and node 8, I have no idea why
```
** INTERNAL ERROR: Uncaught exception **
Uncaught Exception: Invalid parameter not satisfying: sessionManager
```

janpio:
restarted

dpogue:
Thanks for finding and fixing this! I'll merge it in for the upcoming 5.0.1 patch release :)

piotr-cz:
After thinking about it a little more, it's supposed to be considered as bug in cordova-ios v5 and not in plugins as v4 has evaluating emptiness of the spec value before adding it to the Podfile

PR #596, opened 2019-04-14
More logging for simulator selection and deployment

janpio:
### Motivation and Context

Currently the simulator selection and app deployment on the selected simulator is a bit of a black box, the logging is insufficient to properly understand and retrace what is happening.

### Description

- When building for simulator
  - display more information about the simulator being built for (simulator ID, simulator identifier string):  
    ```
    $ cordova run ios --no-telemetry --no-update-notifier --target iPhone-XR --emulator
    Building for "iPhone X" Simulator (com.apple.CoreSimulator.SimDeviceType.iPhone-X, iPhone-X)
    ```
  - before executing `xcodebuild` command, output the `emulatorTarget` that will be used as part of `-destination` option:  
    ```
    Building project: /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp-2604oc4wBu2VePt9/platforms/ios/HelloCordova.xcworkspace
	Configuration: Debug
	Platform: emulator
	Target: iPhone X
    ```
- When deploying
  - log when deploy (both device and simulator) starts:  
    `Deploying to simulator`
  - to simulator
    - use superspawn to trigger `ios-sim`
	    - output the command used to start simulator:  
          `Running command: /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp-2604oc4wBu2VePt9/node_modules/ios-sim/bin/ios-sim launch /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp-2604oc4wBu2VePt9/platforms/ios/build/emulator/HelloCordova.app --devicetypeid com.apple.CoreSimulator.SimDeviceType.iPhone-XR --log /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp-2604oc4wBu2VePt9/platforms/ios/cordova/console.log --exit`
	    - prefix logging coming from `ios-sim` with `ios-sim log: ` for context:
          ```
          ios-sim log: device.name: iPhone X
          ios-sim log: device.runtime: iOS 11.0
          device.id: 5F22BAE5-1DB2-43CC-8983-5C4EE1BA37E7
		  ```
		- output any errors returned from `ios-sim` and fail (instead of hanging):  
          ```
           No available runtimes could be found for "iphone xʀ".
          /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp-2604oc4wBu2VePt9/node_modules/ios-sim/bin/ios-sim: Command failed with exit code 1 Error output:
            No available runtimes could be found for "iphone xʀ".
           ```

(Related: The output this creates would benefit from changes to `ios-sim`: https://github.com/ios-control/ios-sim/pull/255 + https://github.com/ios-control/ios-sim/pull/256 (8.x))

### Testing

This has been used in `cordova-paramedic` to better understand cordova-ios behaviour and uncover a few problems around simulator selection and deployment.

### Checklist

### DIFF
--- a/bin/templates/scripts/cordova/lib/run.js
+++ b/bin/templates/scripts/cordova/lib/run.js
@@ -19,7 +19,6 @@
 
 var Q = require('q');
 var path = require('path');
-var cp = require('child_process');
 var build = require('./build');
 var shell = require('shelljs');
 var superspawn = require('cordova-common').superspawn;
@@ -208,15 +207,15 @@ function iossimLaunch (appPath, devicetypeid, log, exit) {
 
     return superspawn.spawn(f, params, { cwd: projectPath, printCommand: true, stdio: 'pipe' })
         .progress(function (stdio) {
-            if (stdio.stderr) { 
-                console.error(stdio.stderr); 
+            if (stdio.stderr) {
+                console.error(stdio.stderr);
             }
             if (stdio.stdout) {
-                console.log('ios-sim log: ' + data.toString());
+                console.log('ios-sim log: ' + stdio.stdout);
             }
         })
-        .then(function(result){
-            console.log('Simulator successfully started via `ios-sim`.');
+        .then(function (result) {
+            console.log('Simulator successfully started via `ios-sim`.', result);
         });
 }
 
--- a/bin/templates/scripts/cordova/lib/run.js
+++ b/bin/templates/scripts/cordova/lib/run.js
@@ -174,7 +174,6 @@ function deployToDevice (appPath, target, extraArgs) {
  * @return {Promise}        Resolves when deploy succeeds otherwise rejects
  */
 function deployToSim (appPath, target) {
-    // Select target device for emulator. Default is 'iPhone-6'
     if (!target) {
         return require('./list-emulator-images').run()
             .then(function (emulators) {
--- a/bin/templates/scripts/cordova/lib/run.js
+++ b/bin/templates/scripts/cordova/lib/run.js
@@ -177,6 +177,7 @@ function deployToDevice (appPath, target, extraArgs) {
 function deployToSim (appPath, target) {
     events.emit('log', 'Deploying to simulator');
     if (!target) {
+        // Select target device for emulator
         return require('./list-emulator-images').run()
             .then(function (emulators) {
                 if (emulators.length > 0) {
--- a/bin/templates/scripts/cordova/lib/run.js
+++ b/bin/templates/scripts/cordova/lib/run.js
@@ -202,9 +202,13 @@ function startSim (appPath, target) {
     return iossimLaunch(appPath, 'com.apple.CoreSimulator.SimDeviceType.' + target, logPath, '--exit');
 }
 
-function iossimLaunch (app_path, devicetypeid, log, exit) {
+function iossimLaunch (appPath, devicetypeid, log, exit) {
     var f = path.resolve(path.dirname(require.resolve('ios-sim')), 'bin', 'ios-sim');
-    var proc = cp.spawn(f, ['launch', app_path, '--devicetypeid', devicetypeid, '--log', log, exit]);
+    
+    var params = ['launch', appPath, '--devicetypeid', devicetypeid, '--log', log, exit];
+    console.log("$ " + f + params.join(' '));
+    var proc = cp.spawn(f, params);
+    
     proc.stdout.on('data', (data) => {
         console.log(data.toString());
     });
--- a/bin/templates/scripts/cordova/lib/run.js
+++ b/bin/templates/scripts/cordova/lib/run.js
@@ -159,6 +159,7 @@ function checkDeviceConnected () {
  * @return {Promise}        Resolves when deploy succeeds otherwise rejects
  */
 function deployToDevice (appPath, target, extraArgs) {
+    events.emit('log', 'Deploying to device');
     // Deploying to device...
     if (target) {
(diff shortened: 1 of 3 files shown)


bin/templates/scripts/cordova/lib/run.js
```diff
@@ -200,12 +201,22 @@ function startSim (appPath, target) {
...
-function iossimLaunch (app_path, devicetypeid, log, exit) {
+function iossimLaunch (appPath, devicetypeid, log, exit) {
     var f = path.resolve(path.dirname(require.resolve('ios-sim')), 'bin', 'ios-sim');
-    var proc = cp.spawn(f, ['launch', app_path, '--devicetypeid', devicetypeid, '--log', log, exit]);
-    proc.stdout.on('data', (data) => {
-        console.log(data.toString());
-    });
+    var params = ['launch', appPath, '--devicetypeid', devicetypeid, '--log', log, exit];
+
+    return superspawn.spawn(f, params, { cwd: projectPath, printCommand: true, stdio: 'pipe' })
```
raphinesse: If you use `stdio: inherit` here, you should be able to drop the whole `progress` block.

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -159,6 +158,7 @@ function checkDeviceConnected () {
  * @return {Promise}        Resolves when deploy succeeds otherwise rejects
  */
 function deployToDevice (appPath, target, extraArgs) {
+    events.emit('log', 'Deploying to device');
```
raphinesse: It's it usually unclear where you're deploying to? If not, this should go to the verbose log IMHO

raphinesse:
One other minor thing: would be nice if the paths for the ios-sim logging were relative to the project. That would improve readability but it could be more trouble than it's worth too

bin/templates/scripts/cordova/lib/run.js
janpio: The initial idea was to be able to handle the error case different than the normal output that ios-sim "logs". Will that work without all this as well?

bin/templates/scripts/cordova/lib/run.js
janpio: This was meant more as a "structuring" comment between the "build" and the "deploy" phase, but as deploying to devices usually doesn't give any more output (although one might think about changing that) we could move this to verbose as well. Thoughts?

janpio:
The apppath and iossim binary? I didn't touch those, and assumed there was a reason for the absolute paths here :/

bin/templates/scripts/cordova/lib/run.js
raphinesse: Ah right. You do the prefixing. Then nevermind my comment. You can drop the stdio option though, since pipe is the default.

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -200,12 +201,22 @@ function startSim (appPath, target) {
...
-    var proc = cp.spawn(f, ['launch', app_path, '--devicetypeid', devicetypeid, '--log', log, exit]);
-    proc.stdout.on('data', (data) => {
-        console.log(data.toString());
-    });
+    var params = ['launch', appPath, '--devicetypeid', devicetypeid, '--log', log, exit];
+
+    return superspawn.spawn(f, params, { cwd: projectPath, printCommand: true, stdio: 'pipe' })
+        .progress(function (stdio) {
+            if (stdio.stderr) {
+                console.error(stdio.stderr);
```
raphinesse: I think this should probably be prefixed too. Unless there's only stderr output if an error occurs. Then you can drop this completely

bin/templates/scripts/cordova/lib/run.js
raphinesse: I see. Sounds sensible. I can't play with iOS builds myself to get a feeling for it so I'll leave it to your judgement. Do as you see fit.

raphinesse:
> The apppath and iossim binary? I didn't touch those, and assumed there was a reason for the absolute paths here :/

That might well be the case. It's safest to use absolute paths but it _should_™ also work with relative paths since you cwd to the project dir. But you never know that they don't mess up in ios-sim.  And we would need three extra lines just to have prettier paths for logging...

bin/templates/scripts/cordova/lib/run.js
```diff
@@ -200,12 +201,22 @@ function startSim (appPath, target) {
...
+            }
+            if (stdio.stdout) {
+                console.log('ios-sim log: ' + stdio.stdout);
```
dpogue: Should these use `console.log` directly, or go through `events.emit('log', '...')`?

raphinesse: Right, I completely oversaw that

janpio: What is the equivalent of `console.error` then?

janpio:
Thanks for the feedback, incorporated the one I understood.

Please feel free to edit the files yourself to get this ready to merge (as I won't be available the next few days).

dpogue:
Okay, if the tests (eventually) pass, let's merge this as-is and ship it in 5.0.1.
We can adjust the stderr output in the future based on whether we see duplicate errors in the output.

janpio:
Thanks for the cleanup @dpogue!

PR #614, opened 2019-05-08
Don't use whitespace as an indent indicator

oliversalzburg:
This is extremely confusing if your editor is not configured to clearly distinguish different sorts of whitespce

### Platforms affected

iOS

### Motivation and Context

This is just clean code style IMHO

### Checklist


janpio:
Is it common or convention to use tabs in plist files vs spaces?

oliversalzburg:
@janpio What I understand from comments in the code and from observing XCode output files, it appears to be the default behavior of current XCode versions.

PR #622, opened 2019-05-21
Remove outdated component/duo package manager package definition

janpio:
Both https://github.com/componentjs/component and https://github.com/duojs/duo are no more


janpio:
Generally I would agree, but as this refers to `3.7.0-dev` and hasn't been updated for 5 years (2014) I think it is safe to assume that this was broken for a long time and wasn't usable - and hasn't been used by anyone.

PR #631, opened 2019-06-18
Added new platforms.

mrsoneji:
### Platforms affected

ios

### Motivation and Context

Splash Screen isn't supported for iPhone XR and iPhone XS Max. The idea behind this Pull Request is to add support to those iPhone versions.

### Checklist

### DIFF
--- a/bin/templates/project/__PROJECT_NAME__/Images.xcassets/LaunchImage.launchimage/Contents.json
+++ b/bin/templates/project/__PROJECT_NAME__/Images.xcassets/LaunchImage.launchimage/Contents.json
@@ -1,22 +1,54 @@
 {
   "images" : [
     {
-      "extent": "full-screen",
-      "idiom": "iphone",
-      "subtype": "2436h",
-      "filename": "Default-2436h.png",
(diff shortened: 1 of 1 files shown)


brodybits:
Needs review

janpio:
> Splash Screen isn't supported for iPhone XR and iPhone XS Max.

Can someone actually confirm this? How would I notice - just not splash at all? Or broken? Or not correctly positioned?

breautek:
> > Splash Screen isn't supported for iPhone XR and iPhone XS Max.
> 
> Can someone actually confirm this? How would I notice - just not splash at all? Or broken? Or not correctly positioned?

We just encountered this on `iPhone XR`. We first determined that we were missing the proper resolution for `XR`. We then provided the `XR` resolution, but still failed to have our splashscreen displayed. Currently, the cordova splash screen is displayed instead.

I haven't tested this yet.

breautek:
This PR does not solve my problem and in fact causes breakage. When I apply this PR to my own fork the web view shrinks, not sure why that happens.

PR #635, opened 2019-06-25
Bump default minDeploymentTarget to 10.0 in Podfile

jcesarmobile:
We updated the templates and cordovaLib to use 10.0 as deployment target in 5.0, but was not changed here in the Podfile generation.


ramgrg:
Does this restrict minimum iOS version compatibility of the app?

jcesarmobile:
yes and no

yes because it sets the default minimum to iOS 10
but no because the default in other files was already iOS 10, so this only add consistency with the min used.

PR #644, opened 2019-07-14
ci(travis): add Xcode 10.2 & Xcode 11

brodybits:


janpio:
no need to test on Xcode 10.1 any more?

brodybits:
Now keeping Xcode 10.1 on Node.js 10

.travis.yml
```diff
@@ -3,15 +3,19 @@ sudo: false
 
 matrix:
   include:
-    # Run one test with Xcode 9.x, the rest with 10.1
+    # Run one test with Xcode 9.x, the rest with 10.2
```
janpio: should be adapted now as well

brodybits: should be fixed now

brodybits: should be fixed now

dpogue:
We're testing with Xcode 11 only now, closing due to conflicts.

brodybits:
My work is now wasted since people continued updating without even considering this proposal. I did request quite a few reviewers on GitHub, evidently ignored. And I recall raising some other PRs that never got reviewed.

One thing is that it would have been nice to include Xcode 10.2 in Travis CI, as long as it can still keep working, like I proposed here.

erisu:
@brodybits If you still want this merged in, change the target branch to `5.1.x`.

If there were to be any future minors releases for the `5.x` major, a new minor branch would need to be created based off `5.1.x`. Same also applies with future patch releases, PRs would need to be merged into `5.1.x` branch. Your changes in this PR can still be used to test those PRs that would be merged into that branch for any potential 5.x related releases.

brodybits:
Thanks @erisu, I think you are right since cordova-ios is rightfully dropping Xcode pre-11 in master (for upcoming major release). If we do apply this change to 5.x branch, I think we should try to ensure little things such as comments that I added are as consistent as possible with master. I really don't have so much extra time to pay attention to this right now. Some other thoughts for the record:

I think I did overreact. My work was not really such a huge amount, and I do strongly favor the review process we follow to keep our repos as clean as possible. My apologies.

Documentation needs to be updated as well, as I already reported in apache/cordova-docs#1056.

PR #646, opened 2019-07-14
tmp@0.1 update in devDependencies

brodybits:
Let’s see if Travis CI keeps working


janpio:
(FYI: The Draft option for PRs is perfect for such things where you don't want others to look at a PR before it is ready as it clearly communicates that the PR is not ready yet.)

timbru31:
Changelog: https://github.com/raszi/node-tmp/compare/v0.0.33...v0.1.0 and README hint:
```
Version 0.1.0
Since version 0.1.0, all support for node versions < 0.10.0 has been dropped.

Most importantly, any support for earlier versions of node-tmp was also dropped.

If you still require node versions < 0.10.0, then you must limit your node-tmp dependency to versions below 0.1.0.
```

IMHO okay.

dpogue:
cordova-ios already does not support node 0.10.0, so this shouldn't affect our versioning

brodybits:
Thanks guys for the quick feedback!

PR #662, opened 2019-08-30
Optional compile-time decision for WKWebView over UIWebView

### LINKED ISSUE #661: Remove all UIWebView code?
People has started receiving this warning when submitting apps that have references to UIWebView. 

> ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs.

Doesn't matter if they are using just WKWebView, as long as Apple sees "UIWebView" in the app code, they show the warning, and cordova-ios is full of "UIWebView" references. (Even react native, flutter and Capacitor users have received the warning)

**There is currently no solution to this problem for Cordova apps.**

So, instead of integrating WKWebView and making it possible to switch from UIWebView to WKWebView on runtime as it was planned, maybe we should just remove all UIWebView code on next major release?

~~There is no deadline, I've contacted an Apple developer that has been tweeting about moving to WKWebView to see if he can provide more information, will update if I get a response.~~

Apple finally announced the [deadline](https://developer.apple.com/news/?id=12232019b):

> The App Store will no longer accept new apps using UIWebView as of April 2020 and app updates using UIWebView as of December 2020.

fjaeger:
As Apple seems to become super picky about the usage of UIWebView APIs, this PR adds a preference (`WKWebViewOnly`) to config.xml, that makes sure that all traces of `UIWebView` are left behind on compile-time.

This PR should resolve #661


jcesarmobile:
I've tested the PR and doesn't work, the problem is the `WK_WEB_VIEW_ONLY` is added to both the CordovaLib and the user projects, but when setting the preference to true, the `WK_WEB_VIEW_ONLY` is only set to 1 in the user project, but not in CordovaLib, so always build with 0 value and have no effect.
Manually changing the value to 1 works.

fjaeger:
Please try to re-add the iOS platform. Then it should work. Forgot to mention that part ;)

jcesarmobile:
Yeah, removing and adding the platform back works if you have the preference, but I think the `WK_WEB_VIEW_ONLY` value of CordovaLib value should be updated on prepare command as the `WK_WEB_VIEW_ONLY` value of the user project does.

lcaprini:
Hi @fjaeger,
I used your cordova-ios fork but the warning still remains. I added in the config.xml the following preference:
`<preference name="WKWebViewOnly" value="true" />`
and all WK_WEB_VIEW_ONLY in XCode project was set to 1.
By the way Apple continues to send me the warning via email...

Did you tried to upload to the iTunes Connect one app with your cordova-ios platform?

Thanks,
--Luca

NiklasMerz:
> Hi @fjaeger,
> I used your cordova-ios fork but the warning still remains. I added in the config.xml the following preference:
> `<preference name="WKWebViewOnly" value="true" />`
> and all WK_WEB_VIEW_ONLY in XCode project was set to 1.
> By the way Apple continues to send me the warning via email...
> 
> Did you tried to upload to the iTunes Connect one app with your cordova-ios platform?
> 
> Thanks,
> --Luca

This is probably not related. Please check if you use any plugins that refrence UIWebView.

willnix86:
I also couldn't get this to work, sadly. If I search my Xcode codebase for the project, I still see a WHOLE bunch of UIWebView references (in Cordova itself, not just plugins) - as far as I'm aware, Apple just scans your code and sends the warning if it finds refs to UIWebView, right?

jcesarmobile:
This works if you have the preference before adding the platform, or if you remove and add it again. The compiler flags will prevent Apple from seeing the UIWebView references.

brodybits:
> This works if you have the preference before adding the platform, or if you remove and add it again.

This looks like it could be a major pitfall for people not super familiar with Cordova.

Assuming we document this option, I think a big bold warning statement would be needed.

Any way we could have Cordova build show a warning if someone changes the preference without removing and adding back cordova-ios?

jcesarmobile:
I think the PR should be changed so it updates the project without removing and re-adding the platform if possible. If not possible, then document it.

CordovaLib/CordovaLib.xcodeproj/project.pbxproj
```diff
@@ -659,6 +660,7 @@
 				PRODUCT_NAME = Cordova;
 				SDKROOT = iphoneos;
 				SKIP_INSTALL = YES;
+				WK_WEB_VIEW_ONLY = 0;
```
tom-hartz: This should reflect the config.xml preference, not be set to 0.

tom-hartz:
I am seeing the same issue as jcesarmobile, setting the preference 
`<preference name="WKWebViewOnly" value="true" />`
and then re-adding the platform.

It only sets the `WK_WEB_VIEW_ONLY=1` flag in the user project.  It then has to be manually changed in the CordovaLib project.

https://github.com/apache/cordova-ios/pull/662/files#r342301814

@fjaeger is there a way to read the config.xml preference instead of putting 0 here?

jcesarmobile:
I've modified your prepare script to make prepare edit the projects without the need to removing and adding the platform back.
https://github.com/apache/cordova-ios/pull/715

PR #693, opened 2019-10-20
(iOS) Allowed extra variation in version comparison

devleaf-matt:
This should allow more varients on the semver spec to pass.

### Platforms affected

iOS

### Motivation and Context

fixes https://github.com/apache/cordova/issues/165

### Description

Allows more of the semver spec to work with the version comparison function. Previously only X.X.X version would be allowed, however now variants like X.X.X-alphaX or X.X.X-betaX will be allowed.

### Testing

Ran across multiple variants of versions which match the above, is backwards compatible with the original method as if the regex isn't triggered then the system runs as previous. 

### Checklist

- [N/A] I've updated the documentation if necessary

### DIFF
--- a/bin/templates/scripts/cordova/lib/versions.js
+++ b/bin/templates/scripts/cordova/lib/versions.js
@@ -179,8 +179,8 @@ exports.compareVersions = function (version1, version2) {
             return parsed;
         });
     }
-    var parsedVer1 = parseVer(version1);
-    var parsedVer2 = parseVer(version2);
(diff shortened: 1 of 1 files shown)


breautek:
Thanks for your PR. I was imagining that would use `semver` npm package to test for semver. I think this would be better than reinventing the wheel.

devleaf-matt:
That’s actually a much better call. I’ll submit another PR using that instead tomorrow. In which case it would make this one redundant so I’ll close it for now. 👍👍

Issue #583, opened 2019-03-27
Simulator selection via --target is wonky


janpio:
# Bug Report

### What is expected to happen?

Correct simulator is selected and started.

### What does actually happen?

It selects the wrong simulator.

## Information

This is from work on cordova-paramedic, where the process usually selects a simulator that is available automatically and then executes the `run` command with that device as `--target` param. Since upgrading to Cordova iOS x, CLI 8 and Xcode 9 (from 7.x), this doesn't work any more.

### Command or Code

https://travis-ci.org/janpio/cordova-paramedic/jobs/512186335

Relevant snippet:

```
    cordova run --list --emulator --no-telemetry --no-update-notifier | grep ^iPhone | tail -n1
$ cordova run --list --emulator --no-telemetry --no-update-notifier | grep ^iPhone | tail -n1
iPhone-X, 11.0
running:
    instruments -s devices | grep ^iPhone
$ instruments -s devices | grep ^iPhone
iPhone 4s (8.1) [DCAF3602-A974-48F4-B6B6-8247F7BE4290] (Simulator)
iPhone 4s (8.2) [DCDC5358-FFBF-4989-8092-7F65ED8F02AF] (Simulator)
iPhone 4s (8.3) [2725F258-12B4-4EC0-BF92-2F6972815E5B] (Simulator)
iPhone 4s (8.4) [781CBE8C-F7A2-4F43-B860-D5FE770AB9CC] (Simulator)
iPhone 4s (9.0) [95CABF9C-189B-4EC0-AC51-2BE373A29E95] (Simulator)
iPhone 4s (9.1) [F34DC048-527F-4281-858D-E2E3DC5D32DA] (Simulator)
iPhone 4s (9.2) [75838F3E-BE3C-43E0-BC20-E725CC43435D] (Simulator)
iPhone 4s (9.3) [984A1890-994E-4381-ABDC-7A302426E9AF] (Simulator)
iPhone 5 (10.0) [3AA26DE0-C2F0-4CAF-936D-09F1ECCF9BDA] (Simulator)
iPhone 5 (10.1) [F5558AB4-0D25-451A-98E6-80A5661E3615] (Simulator)
iPhone 5 (10.2) [DEF54A5A-0DB7-4752-B2B5-836A924C1960] (Simulator)
iPhone 5 (10.3.1) [2523EF7F-852D-4268-A9A3-C55A4428EAE0] (Simulator)
[...]

PR #708, opened 2019-11-05
[MAJOR] Do not overwrite all targets with same id (minimal)

### LINKED ISSUE #538: Cordova overwrites bundleIdentifier for all existing targets

NiklasMerz:
### Platforms affected
iOS

### Motivation and Context

This is a simplified version of #545 with the minimal changes to fix #538 

### Description

See discussion in #545 

### Testing

Manual tests with complex project

### Checklist


bin/templates/scripts/cordova/lib/prepare.js
```diff
@@ -274,12 +274,54 @@ function handleOrientationSettings (platformConfig, infoPlist) {
...
+    }
+}
+
 function handleBuildSettings (platformConfig, locations, infoPlist) {
     var pkg = platformConfig.getAttribute('ios-CFBundleIdentifier') || platformConfig.packageName();
     var targetDevice = parseTargetDevicePreference(platformConfig.getPreference('target-device', 'ios'));
     var deploymentTarget = platformConfig.getPreference('deployment-target', 'ios');
     var needUpdatedBuildSettingsForLaunchStoryboard = checkIfBuildSettingsNeedUpdatedForLaunchStoryboard(platformConfig, infoPlist);
     var swiftVersion = platformConfig.getPreference('SwiftVersion', 'ios');
+    var displayName = platformConfig.name().replace(/"/g, '');
```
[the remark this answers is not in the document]
NiklasMerz: I did and quick searcg and I could not find a way to get `projectName` or `targetName`. But these lines use displayName and shortName. Should I change this to short name? 

https://github.com/GEDYSIntraWare/cordova-ios/blob/55a06f97bbc72769afa0d32b97053357ce5dba1d/bin/templates/scripts/cordova/lib/prepare.js#L190

Shortname falls back to the full name if not defined:
https://github.com/apache/cordova-common/blob/7ceeeb4ddf0930343588c77c09561a6b4aa89b4b/src/ConfigParser/ConfigParser.js#L121

And this needs to be encoded with `unorm.nfd()`. 

What do you suggest?

NiklasMerz:
I changed the the original PR that way that only the bundle id gets checked for the target. Other settings should be applied to other targets, too.

For me type extensions this necessary. @msari-ipe-ext-1 Does this work for watch extensions.

NiklasMerz:
Would be great to get this and this https://github.com/apache/cordova-node-xcode/pull/79 fixed for the next release. How can I help to get this ready?

brodybits:
Hi @NiklasMerz I may need some extra time to understand the history of issue #538, previous attempts such as #545, this proposal, and apache/cordova-node-xcode#79, before I am ready to answer your question. I would also like to point to the sustainability issue I raised in apache/cordova#163. Thanks for your understanding and thanks for your recent contributions!

bin/templates/scripts/cordova/lib/prepare.js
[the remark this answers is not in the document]
brodybits: I do not follow the arguments 100%. One thing is that one of the code references seems to point into a fork, please point into the code in our repos whenever possible.

@NiklasMerz can you please give a quick explanation of what this variable means?

brodybits:
I added [MAJOR] to the title and set the milestone to the next major release (6.0.0), due to the risk of this change breaking something else.

bin/templates/scripts/cordova/lib/prepare.js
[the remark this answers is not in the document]
NiklasMerz: Oof this was a while ago and I am having a hard time to understand my comments, too. Should be a sign that I am not clear enough.

So the line in our fork is the same in the original repo. I probably got confused with the repos:
https://github.com/apache/cordova-ios/blob/c1a6640758d1e95f7f7e0dbca980df8780a2adf4/bin/templates/scripts/cordova/lib/prepare.js#L191

I am just realising @dpogue probably suggest to change the name of variable and I thought about getting another variable. That's why it's so confusing. Thank you helping sort that out. I think that makes sense and will clear that up once I am back at my workstation.

NiklasMerz:
Thanks @brodybits for your review. Good points.

I am aware of the current maintainer situation. I will do my best to clear this code up and explain the open questions as soon as I find time at work with my dev environment on hand.

Tests are a good idea since this should be a fix for a regression as described in #583. To put that confusing discussion into short words: Cordova previously did not touch additional targets in Xcode projects (like Watch extension, call directroy extension ..). With Cordova 5 that changed and thos additinal targets get the name of the host app for example. This is undesirable for certain complex apps with additional targets.

This PR tries to bring back the old behavior basically. I think the whole replacing placeholder stuff in Xcode projects changed in Cordova 5. That's why https://github.com/apache/cordova-node-xcode/pull/79 is needed to fix the `PRODUCT_BUNDLE_IDENTIFIER` placeholder while adding targets with hooks. 

As said before I am going to work on the code and I am open to discussions. We are using to extensions to our Cordova app and this is really important. A new release with this fix is not urgent because using forks works fine for a while but I don't want to maintain these forks forever.

NiklasMerz:
I updated the `displayName` variable to make this clear.

I can agree the `updateBuildPropertyLocal` method is not that easy to understand but it is handling info.plist file wizardry I don't understand either. The method is written by @wolfmanfx in #545 and just put together in this PR.

@wolfmanfx What do you think of this PR? Do you have ideas how to improve this method?

brodybits:
Thanks @NiklasMerz for the recent fix, looks nice. And thanks for your help and understanding of our maintainer situation.

I completely understand that you are basically adapting a bug fix from someone else, with the obvious challenges of taking over code with missing test cases.

I would like to continue the discussion in #538 which I have just pinned in order to get a better understanding of the whole situation of this functionality.

NiklasMerz:
Yes the missing test cases part is crucial. That kind of app is not considered in the tests and I am having a hard time thinking of adding it right now.

But I must admit that using extensions with Cordova might not be a often used features except for plugins like [open with](https://github.com/j3k0/cordova-plugin-openwith) .

timbru31:
I've restarted the timed out Travis test, CI is now green

bin/templates/scripts/cordova/lib/prepare.js
```diff
@@ -275,12 +275,54 @@ function handleOrientationSettings (platformConfig, infoPlist) {
     }
 }
 
+// Make sure only update properties from our target project
+function updateBuildPropertyLocal (proj, targetName, prop, value, build) {
+    try {
+        // Check if we have a valid target - during prepare we do not have it
+        var target = proj.pbxTargetByName(targetName);
```
erisu:
```suggestion
        const target = proj.pbxTargetByName(targetName);
```

bin/templates/scripts/cordova/lib/prepare.js
```diff
@@ -275,12 +275,54 @@ function handleOrientationSettings (platformConfig, infoPlist) {
...
+        if (target == null || target.buildConfigurationList == null) {
```
erisu:
```suggestion
        if (!target || !target.buildConfigurationList) {
```
If you want to use `== null`, make sure you use triple equals `=== null`

bin/templates/scripts/cordova/lib/prepare.js
```diff
@@ -275,12 +275,54 @@ function handleOrientationSettings (platformConfig, infoPlist) {
...
+            proj.updateBuildProperty(prop, value, build);
+        } else {
+            var targetProjectBuildReference = target.buildConfigurationList;
+            // Collect the uuid's from the configuration of our target
+            var COMMENT_KEY = /_comment$/;
+            var validConfigs = [];
+            var configList = proj.pbxXCConfigurationList();
```
erisu: Try to use `const` and `let` here as well.. any anywhere else..

erisu: Just a nitpick but `targetProjectBuildReference` is used only once and also it is 2 characters less than `target.buildConfigurationList`, you might not actually need to create an extra variable.

bin/templates/scripts/cordova/lib/prepare.js
```diff
@@ -275,12 +275,54 @@ function handleOrientationSettings (platformConfig, infoPlist) {
...
+                        validConfigs.push(buildVariants[i].value);
+                    }
+                    break;
+                }
+            }
+            // Only update target props
+            var configs = proj.pbxXCBuildConfigurationSection();
+            for (configName in configs) {
+                if (!COMMENT_KEY.test(configName)) {
+                    if (validConfigs.indexOf(configName) === -1) {
```
erisu:
```suggestion
                    if (!validConfigs.includes(configName)) {
```

bin/templates/scripts/cordova/lib/prepare.js
NiklasMerz: I am working to much with typescript. I have forgotton all JS quirks :-)

bin/templates/scripts/cordova/lib/prepare.js
NiklasMerz: Done

NiklasMerz:
This code somehow works for my project, but is not correct. We need to find a better solution possibly involving a change in cordova-node-xcode.

I would love to contribute this but I think it is beyon my capabilities and understandings of Cordova. I tried to work on this a few times but I head always explodes. I appreciate if someone with more experience has the capacity to work on this and I will do my best to support that.

Issue #700, opened 2019-10-25
Cocoapods 1.8.x platform add fails


kjng:
# Bug Report

### What is expected to happen?

iOS platform is successfully added using Cocoapods 1.8.x and running `cordova platform add ios`

### What does actually happen?

An error is thrown from Cordova attempting to run pod install.

## Information

I am using cordova-plugin-firebasex which needs to install relevant pods.

The error message is:
> 'The CocoaPods repo has not been synced yet, this will take a long time (approximately 500MB as of Sept 2016). Please run `pod setup` first to sync the repo.'

It appears to originate from https://github.com/apache/cordova-ios/blob/f43812ab249a6fe7bca3ab2611de7d8a9dc080f2/bin/templates/scripts/cordova/lib/check_reqs.js#L132 and I think this is caused by changes in Cocoapods 1.8.x.

With Cocoapods 1.8.x, the master spec repo isn't required/included, so I think the check_reqs repo validation that Cordova does while trying to install pods fails.

### Command or Code

`cordova platform add ios`

### Environment, Platform, Device

I'm building using an Azure DevOps MacOS build agent:

MacOS 10.14
Cocoapods 1.8.3

### Version information

Cordova CLI: 9.0.0
cordova-ios: 5.0.1

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

DominicBoettger:
Same here. We solved it by downgrading cocoapods in our azure devops pipelines. Now our builds needs additional 10 minutes to be built :-( .

erisu:
#719 should resolve this.

mibzman:
This issue is only partially resolved.  #717 has the same behavior, except that it only occurs when the app name has a single quote.

brodybits:
> This issue is only partially resolved. #717 has the same behavior
> [...]

Due to limited resources we do not like to support discussions in closed issues. Do you see anything beyond issue #717 that needs to be fixed regarding the platform failure on Cocoapods 1.8.x?

If so, it would really help the user community if you would raise a new issue with the required information including the standard [minimal, reproducible example ([1])](https://stackoverflow.com/help/minimal-reproducible-example).

[1] https://stackoverflow.com/help/minimal-reproducible-example

PR #719, opened 2019-11-17
Add check for newer versions of cocoapods to avoid locking adding platforms on non synced pods repo

renanccastro:
Starting on 1.8.0, cocoapods now use CDN to avoid having to sync the master repo, so we don't need to check anymore if the repo is populated.

Even more, the "pod setup" command now doesn't do anything, and newer versions of cocoapods are actually incompatible with cordova-ios, as the repository filled check will kick in and make it impossible to use it.

Error example with meteor and cocoapods 1.8.4:

✗ CocoaPods: The CocoaPods repo has not been synced yet, this will take a long
  time (approximately 500MB as of Sept 2016). Please run `pod setup` first to
  sync the repo.

This also solves issue #700 

### Checklist

- [X] I added automated test coverage as appropriate for this change
- [X] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [X] I've updated the documentation if necessary

### DIFF
--- a/bin/templates/scripts/cordova/lib/check_reqs.js
+++ b/bin/templates/scripts/cordova/lib/check_reqs.js
@@ -23,6 +23,7 @@ const Q = require('q');
 const shell = require('shelljs');
 const util = require('util');
 const versions = require('./versions');
+const semver = require('semver');
 
 const SUPPORTED_OS_PLATFORMS = [ 'darwin' ];
 
@@ -121,15 +122,23 @@ module.exports.check_cocoapods = function (toolChecker) {
         // check whether the cocoapods repo has been synced through `pod repo` command
         // a value of '0 repos' means it hasn't been synced
         .then(function (toolOptions) {
+            let podVersion = shell.exec('pod --version', { silent: true });
+            if (podVersion && podVersion.output) {
+                const version = podVersion.output.trim();
+                // starting with 1.8.0 cocoapods now use cdn and we dont need to sync first
+                if ((semver.valid(version) && semver.gte(version, '1.8.0')) || toolOptions.ignore) {
+                    return toolOptions;
+                }
+            }
             let code = shell.exec('pod repo | grep -e "^0 repos"', { silent: true }).code;
             let repoIsSynced = (code !== 0);
 
             if (toolOptions.ignore || repoIsSynced) {
                 // return check_cocoapods_repo_size();
                 // we could check the repo size above, but it takes too long.
-                return Q.resolve(toolOptions);
+                return toolOptions;
             } else {
-                return Q.reject(COCOAPODS_NOT_SYNCED_MESSAGE);
+                return Promise.reject(COCOAPODS_NOT_SYNCED_MESSAGE);
             }
         });
 };
--- a/bin/templates/scripts/cordova/lib/check_reqs.js
+++ b/bin/templates/scripts/cordova/lib/check_reqs.js
@@ -23,7 +23,6 @@ const Q = require('q');
 const shell = require('shelljs');
 const util = require('util');
 const versions = require('./versions');
-const semver = require('semver');
 
 const SUPPORTED_OS_PLATFORMS = [ 'darwin' ];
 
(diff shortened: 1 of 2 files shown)


bin/templates/scripts/cordova/lib/check_reqs.js
```diff
@@ -121,6 +121,20 @@ module.exports.check_cocoapods = function (toolChecker) {
...
+            let podVersion = shell.exec('pod --version', { silent: true });
+            if (podVersion) {
+                const semver = podVersion.toString().split('.');
+                if (semver.length >= 3) {
+                    const major = parseInt(semver[0]);
+                    const minor = parseInt(semver[1]);
+                    const patch = parseInt(semver[2]);
+
+                    // starting with 1.8.0 cocoapods now use cdn and we dont need to sync first
+                    if ((major >= 1 && minor >= 8 && patch >= 0) || toolOptions.ignore) {
```
DAnn2012: Hi,

maybe it's better this way:

`if (((major > 1) || (major == 1 && minor >= 8 && patch >= 0)) || toolOptions.ignore) {`

do not you think?

DJmRek: It's maybe also sufficient to only check for major and minor version.

bin/templates/scripts/cordova/lib/check_reqs.js
```diff
@@ -121,6 +121,20 @@ module.exports.check_cocoapods = function (toolChecker) {
...
+                        return Q.resolve(toolOptions);
+                    }
```
erisu: I would recommend two thing.

1. Use the `semver` dependency that we already include with `cordova-ios`.
2. Return a native promise instead of using `Q`. We are removing the `Q` dependency in next major.

Example:

```js
const semver = require('semver');

if (semver.satisfies(`>=${podVersion.toString()}`) || toolOptions.ignore) {
    return Promise.resolve(toolOptions)
}

```

raphinesse: Regarding 2. (Q vs native Promise): Here we return inside of a `then` callback. Anything we return will be wrapped inside a Promise automatically. So no need to manually wrap it at all here.

renanccastro: Thanks, guys. I missed that this dependency was already on the tree. I've already changed the code and squashed the commit. I changed the other flow inside this method to return native promise also, to standardize this at least inside this method. Thanks!

erisu:
All that is missing now is resolving the AppVeyor test.

Since AppVeyor runs on Windows and there is no `pod` installation, you might want to skip the checking and return something to present that requirement was not met.

renanccastro:
> All that is missing now is resolving the AppVeyor test.
> 
> Since AppVeyor runs on Windows and there is no `pod` installation, you might want to skip the checking and return something to present that requirement was not met.

Done!

raphinesse:
@renanccastro When giving the code a thorough review, I noticed that we already get the cocoapods version passed in as part of the result from the `check_cocoapod_tool` call. I updated your code to use that and also used `versions.compareVersions` which internally uses `semver` but also gracefully handles non-semver versions.

It would be great to have a unit test that checks the behavior for pod versions starting with `1.8.0`. Other that, that this is good to go for me now. Hopefully we can include this in the upcoming release. In any case: thanks for your contribution and patience.

bin/templates/scripts/cordova/lib/check_reqs.js
```diff
@@ -121,15 +121,22 @@ module.exports.check_cocoapods = function (toolChecker) {
         // check whether the cocoapods repo has been synced through `pod repo` command
         // a value of '0 repos' means it hasn't been synced
         .then(function (toolOptions) {
+            if (toolOptions.ignore) return toolOptions;
```
timbru31: a bit nit-picky, but what is your code styleguide in regards to `{ }` after an `if` statement?

raphinesse: Our ESLint style is pretty liberal in this aspect, I think. My personal styleguide says "prefer return statement on same line & without braces if line does not get too long"

erisu: I also prefer `return`, `throw`, and sometimes even variable setting with out using `{ }` as long as the line is not too long.

Issue #679, opened 2019-09-30
Binary make use of malloc Function


oxr463:
# Bug Report

Cordova iOS Applications scanned with [MobSF](https://github.com/MobSF/Mobile-Security-Framework-MobSF) report use of unsafe `malloc` function, (See also: https://github.com/MobSF/Mobile-Security-Framework-MobSF/issues/372).

## Problem

See: [CWE-789: Uncontrolled Memory Allocation](https://cwe.mitre.org/data/definitions/789.html)

### What is expected to happen?

Memory should be allocated via `calloc`.

### What does actually happen?

Memory is allocated via unsafe `malloc`.

## Information

See below.

### Command or Code

Searching the repository leads me to the following line of code:

```
Byte* buffer = (Byte*)malloc((unsigned long)[assetRepresentation size]);
```

Source: https://github.com/apache/cordova-ios/blob/46da4bc34495ec7bfe13c1ec1b4074dd2e82641a/CordovaLib/Classes/Public/CDVURLProtocol.m#L65

### Environment, Platform, Device

iOS

### Version information

Cordova

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

jd048:
I am facing same issue with latest iOS cordova project, can any one give me solution for same 
Thanks

brodybits:
I am sorry but this really looks like a false positive to me. I am becoming increasingly suspicious of the many code scanning tools that assert this kind of message without sufficient analysis to back it up. See this reply on a SQLite thread: http://sqlite.1065341.n5.nabble.com/Coding-standard-tc109945.html#a109956

This kind of report diverts time and attention away from important issues that we **do** need to resolve.

I did take a look at how the code in PR #7741 is using ALAssetLibrary and would say we can expect it to go away once we drop UIWebView support: #551

oxr463:
> I am sorry but this really looks like a false positive to me. I am becoming increasingly suspicious of the many code scanning tools that assert this kind of message without sufficient analysis to back it up. 

In my experience, many of these tools are not sophisticated enough to determine how `malloc` is being used, just that it is present. However, after examining the code for myself this looks to be a trivial fix; whether or not it is a false positive, there is no detriment by switching to `calloc`, and this is better than trying to patch dozens of tools, some of which are proprietary in nature.

> This kind of report diverts time and attention away from important issues that we do need to resolve.

[...]

PR #741, opened 2019-12-12
WIP: Replace malloc with calloc

oxr463:
### Platforms affected

iOS

### Motivation and Context

Bug: https://github.com/apache/cordova-ios/issues/679

### Description

Replace `malloc`, with `calloc`.

### Testing

Code compiles and run successfully.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVURLProtocol.m
+++ b/CordovaLib/Classes/Public/CDVURLProtocol.m
@@ -62,7 +62,7 @@ NSString* const kCDVAssetsLibraryPrefixes = @"assets-library://";
                 // We have the asset!  Get the data and send it along.
                 ALAssetRepresentation* assetRepresentation = [asset defaultRepresentation];
                 NSString* MIMEType = (__bridge_transfer NSString*)UTTypeCopyPreferredTagWithClass((__bridge CFStringRef)[assetRepresentation UTI], kUTTagClassMIMEType);
-                Byte* buffer = (Byte*)malloc((unsigned long)[assetRepresentation size]);
+                Byte* buffer = (Byte*)calloc((unsigned long)[assetRepresentation size], sizeof(Byte));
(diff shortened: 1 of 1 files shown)


CordovaLib/Classes/Public/CDVURLProtocol.m
```diff
@@ -62,7 +62,7 @@ - (void)startLoading
                 // We have the asset!  Get the data and send it along.
                 ALAssetRepresentation* assetRepresentation = [asset defaultRepresentation];
                 NSString* MIMEType = (__bridge_transfer NSString*)UTTypeCopyPreferredTagWithClass((__bridge CFStringRef)[assetRepresentation UTI], kUTTagClassMIMEType);
-                Byte* buffer = (Byte*)malloc((unsigned long)[assetRepresentation size]);
+                Byte* buffer = (Byte*)calloc((unsigned long)[assetRepresentation size], sizeof(Byte));
```
brodybits: - I cannot see any real benefit to using calloc rather than malloc here, since we know the memory will be overwritten anyway
- While reviewing this change, I noticed a minor opportunity for the following code golf: store the result of `[assetRepresentation size]` which is invoked multiple (2) times in the code here

oxr463:
> 
> 
> Thanks for the contribution. I am personally not convinced there is any value to clearing allocated memory which we know will be overwritten anyway. I also noticed a very small opportunity for code golf which is in a comment on the code.

This is a low finding for mobile penetration tests. I have flagged countless applications for this, and they all seemed to be using cordova for iOS, hence in order to remediate the flaw; I took it upon myself to submit an issue here.

brodybits:
This sounds like a false positive to me, see my comments here: <https://github.com/apache/cordova-ios/issues/679#issuecomment-565102344>

dpogue:
CDVURLProtocol has been removed on master as part of dropping UIWebView support, so this PR no longer applies.

PR #747, opened 2019-12-17
refactor: use arrow functions

erisu:
### Motivation and Context

Modernize code

### Description

* Use `lebab` to apply `--transform arrow` and `--transform arrow-return`.
* Reverted some of the applied `arrow-return` for readability.

### Testing

- `npm t`

### DIFF
--- a/bin/templates/scripts/cordova/lib/check_reqs.js
+++ b/bin/templates/scripts/cordova/lib/check_reqs.js
@@ -61,8 +61,8 @@ module.exports.check_ios_deploy = () => {
     return checkTool('ios-deploy', IOS_DEPLOY_MIN_VERSION, IOS_DEPLOY_NOT_FOUND_MESSAGE);
 };
 
-// Build iOS apps available for OSX platform only, so we reject on others platforms
 module.exports.check_os = () => {
+    // Build iOS apps available for OSX platform only, so we reject on others platforms
     return os_platform_is_supported()
         ? Q.resolve(process.platform)
         : Q.reject('Cordova tooling for iOS requires Apple macOS');
--- a/bin/templates/scripts/cordova/lib/check_reqs.js
+++ b/bin/templates/scripts/cordova/lib/check_reqs.js
@@ -82,54 +82,58 @@ function check_cocoapod_tool (toolChecker) {
  * Checks if cocoapods repo size is what is expected
  * @return {Promise} Returns a promise either resolved or rejected
  */
-module.exports.check_cocoapods_repo_size = () => check_cocoapod_tool()
-    .then(toolOptions => {
-        // check size of ~/.cocoapods repo
-        const commandString = util.format('du -sh %s/.cocoapods', process.env.HOME);
-        const command = shell.exec(commandString, { silent: true });
-        // command.output is e.g "750M   path/to/.cocoapods", we just scan the number
-        const size = toolOptions.ignore ? 0 : parseFloat(command.output);
+module.exports.check_cocoapods_repo_size = () => {
+    return check_cocoapod_tool()
+        .then(toolOptions => {
+            // check size of ~/.cocoapods repo
+            const commandString = util.format('du -sh %s/.cocoapods', process.env.HOME);
+            const command = shell.exec(commandString, { silent: true });
+            // command.output is e.g "750M   path/to/.cocoapods", we just scan the number
+            const size = toolOptions.ignore ? 0 : parseFloat(command.output);
 
-        if (toolOptions.ignore || command.code === 0) { // success, parse output
-            return Q.resolve(size, toolOptions);
-        } else { // error, perhaps not found
-            return Q.reject(util.format('%s (%s)', COCOAPODS_REPO_NOT_FOUND_MESSAGE, command.output));
-        }
-    })
-    .then((repoSize, toolOptions) => {
-        if (toolOptions.ignore || COCOAPODS_SYNCED_MIN_SIZE <= repoSize) { // success, expected size
-            return Q.resolve(toolOptions);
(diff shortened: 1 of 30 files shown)


bin/templates/scripts/cordova/lib/clean.js
```diff
@@ -24,19 +24,26 @@ const superspawn = require('cordova-common').superspawn;
...
+        ['-project', projectName, '-configuration', 'Debug', '-alltargets', 'clean'],
+        { cwd: projectPath, printCommand: true, stdio: 'inherit' }
+    )
+        .then(
+            () => superspawn.spawn(
+                'xcodebuild',
+                ['-project', projectName, '-configuration', 'Release', '-alltargets', 'clean'],
+                { cwd: projectPath, printCommand: true, stdio: 'inherit' }
+            )
+        )
```
raphinesse: Ugh, can we please transform this to something like the following:
```js
    const xcodebuildClean = configName => {
        return superspawn.spawn(
            'xcodebuild',
            ['-project', projectName, '-configuration', configName, '-alltargets', 'clean'],
            { cwd: projectPath, printCommand: true, stdio: 'inherit' }
        );
    };

    return xcodebuildClean('Debug')
        .then(() => xcodebuildClean('Release'))
```

If it's too much out of scope for this PR that's OK, but I hope the new formatting isn't here to stay :sweat_smile:

tests/spec/unit/lib/check_reqs.spec.js
```diff
@@ -46,25 +46,21 @@ describe('check_reqs', function () {
             );
         });
 
-        it('should throw error because version is not following semver-notated.', () => {
-            return checkTool('node', 'a.b.c').then(
-                () => fail('Expected promise to be rejected'),
-                err => expect(err).toEqual(new TypeError('Invalid Version: a.b.c'))
-            );
-        });
+        it('should throw error because version is not following semver-notated.', () => checkTool('node', 'a.b.c').then(
```
raphinesse: That one wasn't caught by the rule I sent you. Here's a rule that requires tests to have block bodies:
```yaml
no-restricted-syntax: [2, 'CallExpression[callee.name=it] > .arguments:nth-child(2) > .body:not(BlockStatement)']
```

tests/spec/unit/lib/check_reqs.spec.js
```diff
@@ -46,25 +46,21 @@ describe('check_reqs', function () {
...
             });
         });
 
-        it('should reject because tool does not meet minimum requirement.', () => {
-            return checkTool('node', '1.0.1').then(
-                () => fail('Expected promise to be rejected'),
-                reason => expect(reason).toContain('version 1.0.1 or greater, you have version 1.0.0')
-            );
-        });
+        it('should reject because tool does not meet minimum requirement.', () => checkTool('node', '1.0.1').then(
```
raphinesse: this should also have a block body

bin/templates/scripts/cordova/lib/check_reqs.js
```diff
@@ -49,24 +49,18 @@ const COCOAPODS_REPO_NOT_FOUND_MESSAGE = 'The CocoaPods repo at ~/.cocoapods was
  * Checks if xcode util is available
  * @return {Promise} Returns a promise either resolved with xcode version or rejected
  */
-module.exports.run = module.exports.check_xcodebuild = function () {
-    return checkTool('xcodebuild', XCODEBUILD_MIN_VERSION, XCODEBUILD_NOT_FOUND_MESSAGE);
-};
+module.exports.run = module.exports.check_xcodebuild = () => checkTool('xcodebuild', XCODEBUILD_MIN_VERSION, XCODEBUILD_NOT_FOUND_MESSAGE);
```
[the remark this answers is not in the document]
breautek: I think I have an unpopular opinion here, but as much as I love arrow functions for it's automatic context binding, I really hate the syntax and find it confusing. In my own code bases, I always enforce parenthesis and curly braces for consistency. In my opinion, this makes the code base easier to read.

```javascript
() => {
    ...
}
```

Omitting the curly braces for the implicit return is, in my opinion confusing. Additionally it does not work if you're attempting to return an object literal. Therefore you are forced to switch up your syntax style depending on what the type of data you are returning and could be a source of bugs. Considering it doesn't even produce a syntax error, it can be a bug that may be hard to trace as it will simply just cause the function to return undefined. See below.

```
Welcome to Node.js v12.13.1.
Type ".help" for more information.
> let x = () => {test:true}
undefined
> x()
undefined
> x
[Function: x]
> let y = () => 123
undefined
> y()
123
> 
```

Not sure on the capabilities of `lebab`, but I would prefer configuring it to always use parenthesis for defining the arrow function argument section and to always use curly brace to define the function body, if possible.

bin/templates/scripts/cordova/lib/clean.js
erisu: def. out of scope.

bin/templates/scripts/cordova/lib/check_reqs.js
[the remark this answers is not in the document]
raphinesse: I wouldn't be a fan of always requiring the parens, but I agree that some expressions involving arrow functions can become hard to read. The code I flagged here was a good example for that.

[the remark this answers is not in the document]
breautek: The function body with curly braces is really my primary concern.

timbru31:
I'v took the liberty to resolve the minor merge conflicts via the GitHub UI.

erisu:
Thank you for fixing the merge conflicts @timbru31

PR #769, opened 2020-01-21
Add Python 3 compatibility to bin/cordova_plist_to_config_xml

cclauss:
These changes should be compatible with both Python 2 and Python 3.

### Checklist

### DIFF
--- a/bin/cordova_plist_to_config_xml
+++ b/bin/cordova_plist_to_config_xml
@@ -26,17 +26,13 @@ Usage:
 from __future__ import print_function
 
 import fileinput
+import io
 import os
 import plistlib
 import re
 import sys
 from xml.dom import minidom
 from xml.etree import ElementTree
-
-try:  # Python 2
-  from StringIO import StringIO
-except ImportError:  # Python 3
-  from io import StringIO
 
 
 def Usage():
@@ -93,10 +89,10 @@ def ConvertPlist(src_path, dst_path):
     root.append(ElementTree.Element('access', attrib={'origin':value}))
 
   tree = ElementTree.ElementTree(root)
-  s = StringIO()
-  tree.write(s, encoding='UTF-8')
-  mini_dom = minidom.parseString(s.getvalue())
-  with open(dst_path, 'w') as out:
+  with io.BytesIO() as s:
+    tree.write(s, encoding='UTF-8')
+    mini_dom = minidom.parseString(s.getvalue())
+  with open(dst_path, 'wb') as out:
     out.write(mini_dom.toprettyxml(encoding='UTF-8'))
 
 
--- a/bin/cordova_plist_to_config_xml
+++ b/bin/cordova_plist_to_config_xml
@@ -23,15 +23,20 @@ Converts a project's Cordova.plist file into a config.xml one. This conversion i
 Usage:
   plist2xml.py path/to/project
 """
+from __future__ import print_function
 
-import StringIO
 import fileinput
-import plistlib
 import os
+import plistlib
 import re
 import sys
 from xml.dom import minidom
 from xml.etree import ElementTree
+
+try:  # Python 2
+  from StringIO import StringIO
+except ImportError:  # Python 3
+  from io import StringIO
 
 
 def Usage():
@@ -88,7 +93,7 @@ def ConvertPlist(src_path, dst_path):
     root.append(ElementTree.Element('access', attrib={'origin':value}))
 
   tree = ElementTree.ElementTree(root)
-  s = StringIO.StringIO()
+  s = StringIO()
   tree.write(s, encoding='UTF-8')
   mini_dom = minidom.parseString(s.getvalue())
   with open(dst_path, 'w') as out:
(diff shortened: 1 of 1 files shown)


brodybits:
Thanks for the contribution, I just requested a review from some iOS experts. My apologies for the delay.

bin/cordova_plist_to_config_xml
```diff
@@ -101,7 +106,7 @@ def UpdateProjectFile(path):
     if 'Cordova.plist' in line:
       line = line.replace('Cordova.plist', 'config.xml')
       line = line.replace('lastKnownFileType = text.plist.xml', 'lastKnownFileType = text.xml')
-    print line,
+    print(line, end='')
```
DavidM42: Why is it important to overwrite the newline character at the end of print here? Does this need to print in one line? Else `end=''` could be left out

NiklasMerz:
> lgtm. Asked if `end=''` really is needed but that's non breaking

If have no idea about python, that's why I got someone to review it who has :-)

bin/cordova_plist_to_config_xml
cclauss: Did you forget this line?  `from __future__ import print_function`

DavidM42: Yeah, you're right. Never worked with that before

erisu:
Just as a side-note, this file seems to be used only for the Platform-Centered workflow.

There are multiple ways of going about creating a Platform-Centered project. One is a fresh project from Xcode. This means the Cordova's Xcode based project is not used and therefore does not have the `config.xml` file.

IMO, I think we do not have a lot of resources to focus on the Platform-Center workflow. I think it may have been discussed to consolidate our focus only on the CLI workflow and potentially dropping Platform-Center workflow.

If the changes work, we could just push it through and not spend too much time on this.

Just My Opinion...

brodybits:
Agreed @erisu, let's get this merged.

I just raised #791 to avoid losing track of improving the [platform-centered workflow ([1])](https://cordova.apache.org/docs/en/latest/guide/platforms/ios/#platform-centered-workflow) and hopefully alleviate the need for Python (someday). Definitely very unfortunate that we cannot afford the resources to make this kind of improvement right now.

[1] https://cordova.apache.org/docs/en/latest/guide/platforms/ios/#platform-centered-workflow

brodybits:
I think we want to drop this functionality in the near future, as I proposed in <https://github.com/apache/cordova/issues/198>. I wonder if we can get this fix into a minor release somehow. What do you think @erisu?

@erisu let's please get this merged in the next 12-24 hours if there are no objections, I do think it would be good to get this solution committed even if we do not want to support it too much longer.

bin/cordova_plist_to_config_xml
```diff
@@ -106,7 +106,7 @@ def UpdateProjectFile(path):
     if 'Cordova.plist' in line:
       line = line.replace('Cordova.plist', 'config.xml')
       line = line.replace('lastKnownFileType = text.plist.xml', 'lastKnownFileType = text.xml')
-    print(line, end='')
+    print(line, end=' ')
```
brodybits: Why add a space here?

cclauss: To put a space between each line.

brodybits: I don't really understand (did not work with Python for about 10 years), hope someone else can review it. We do appreciate your efforts to get this whole Python compatibility issue fixed.

cclauss: Which do we want?
```
>>> for char in "abcdefg":
...     print(char, end="")  # --> abcdefg
    -- or --
>>> for char in "abcdefg":
...     print(char, end="")  # --> a b c d e f g
```

knight9999:
I tried this PR with Python 3.6.4. But I got the following error.
```
TypeError: string argument expected, got 'bytes'
```

I feel it is better to use `BytesIO` with `ElementTree.write` in place of `StringIO` as follows.
```
diff --git a/bin/cordova_plist_to_config_xml b/bin/cordova_plist_to_config_xml
index f6c47505..94fe3371 100755
--- a/bin/cordova_plist_to_config_xml
+++ b/bin/cordova_plist_to_config_xml
@@ -38,6 +38,7 @@ try:  # Python 2
 except ImportError:  # Python 3
   from io import StringIO
 
+from io import BytesIO
 
 def Usage():
   sys.stderr.write(__doc__)
@@ -93,10 +94,11 @@ def ConvertPlist(src_path, dst_path):
     root.append(ElementTree.Element('access', attrib={'origin':value}))
 
   tree = ElementTree.ElementTree(root)
-  s = StringIO()
+  # s = StringIO()
+  s = BytesIO()
   tree.write(s, encoding='UTF-8')
   mini_dom = minidom.parseString(s.getvalue())
-  with open(dst_path, 'w') as out:
+  with open(dst_path, 'wb') as out:
     out.write(mini_dom.toprettyxml(encoding='UTF-8'))
```

PR #715, opened 2019-11-09
feat: add optional compile-time decision for disabling UIWebView

jcesarmobile:
Based on https://github.com/apache/cordova-ios/pull/662, updated the prepare script to also modify the CordovaLib project so no removing and re adding platform is needed when `<preference name="WKWebViewOnly" value="true" />` was added to the `config.xml` after the platform was added.
Also makes it possible to unset the value, the implementation on https://github.com/apache/cordova-ios/pull/662 didn't allow to change the value back to 0 without removing the platform. 

I added the modified platform and submitted to App store and got the warning, then added `<preference name="WKWebViewOnly" value="true" />`, ran `cordova prepare ios`, submitted again and didn't get the warning. 

closes: #662

PR #771, opened 2020-01-22
(iOS) Making CDVUIWebViewDelegate a project-scoped header

bhariharan:
### Platforms affected
`iOS`

### Motivation and Context
Generated pure `Swift` apps with `WK_WEB_VIEW_ONLY` set to `true` will throw compilation errors since the `Swift` compiler won't accept incomplete umbrella headers (the `Cordova` module is deemed incomplete).

### Description
- `CDVUIWebViewDelegate.h` was a public-scoped header. However, in recent times (in [this PR](https://github.com/apache/cordova-ios/pull/715/files)), this class could get compiled out if the preprocessor directive `WK_WEB_VIEW_ONLY` is set to `true` (which is always the case for our apps, since we no longer support `UIWebView`).
- As a result, the umbrella header generated for `CordovaLib` is incomplete since it won't include this header.
- For `Obj C` apps, this is simply a warning. However, is a pure `Swift` app, this will cause a compilation failure.
- The only workaround is to make this file a project-scoped header instead.

### Testing
We have a plugin [here](https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin), where I applied the changes and verified it works. I have also verified that it works on our internal projects.

### Checklist


jcesarmobile:
thanks for the PR, but CDVUIWebViewDelegate is being removed (see https://github.com/apache/cordova-ios/pull/773)

PR #663, opened 2019-09-02
Fix #661: Removed all UIWebView code.

bpresles:
[Build Status](https://travis-ci.org/bpresles/cordova-ios)

Removed all UIWebView related code.
Replaced default engine by CDVWKWebViewEngine.

### Motivation and Context

I've tested the pull request from https://github.com/apache/cordova-ios/pull/662, but the AppStore Connect still reported "ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs." with it (I tripled check that the WKWebViewOnly preference was correctly taken in account).

Also I don't think that doing a half way solution at compile time, instead of removing the UIWebView code altogether, is a good solution.
If the UIWebView is deprecated, than remove the code altogether as it'll be useless in the near future. People wanting to still use UIWebView can either use a UIWebView engine as a plugin or use older version of cordova-ios.

https://github.com/apache/cordova-ios/issues/661

### Description
Removed all UIWebView code and replaced default engine by CDVWKWebViewEngine from cordova-plugin-wkwebview-engine (making it useless now).

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [x] I added automated test coverage as appropriate for this change
- [ ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [ ] I've updated the documentation if necessary

PR #773, opened 2020-01-29
refactor: drop uiwebview & add wkwebview

erisu:
### Motivation, Context, Description

* Removes UIWebView
* Integrates `cordova-plugin-wkwebview-engine`
* Renames WKWebView engine plugin classes to not conflict with Ionic.
  - Prefixed w/ `CDVWebView`: 
  - Notice: Plugins should not use the follow prefix `CDV`. It is typically used for Cordova's internal class naming convention. If a plugin happens to uses the same class name, it will create conflicts. As well, if any changes were made on Cordova's end and happens to break the plugin, the plugin would need to make the corrections because the plugin decided to use and conflict with Cordova's internal naming convention.
* Removes `WK_WEB_VIEW_ONLY` preference options.
* fixes: #661
* closes: #663

### Testing

- `npm t`
- `cordova platform add`
- `cordova requirements ios`
- `cordova prepare ios`
- `cordova build ios --release --device`
  - Uploaded Manually Built IPA
  - Launch Manual Uploaded IPA
- `cordova run ios --release --emulator`
- IAB test
  - `cordova plugin add cordova-plugin-inappbrowser@3.2.0`
  - `cordova build ios --release --device`
    - Uploaded Manually Built IPA
    - Launch Manual Uploaded IPA

### Checklist


cordova-js-src/exec.js
```diff
@@ -259,3 +163,12 @@ execProxy.nativeCallback = function () {
 };
 
 module.exports = execProxy;
+
+if (window.webkit && window.webkit.messageHandlers && window.webkit.messageHandlers.cordova && window.webkit.messageHandlers.cordova.postMessage) {
+    // unregister the old bridge
+    cordova.define.remove('cordova/exec');
```
dpogue: We shouldn't need to do this hack if cordova/exec is already the right implementation for WKWebView

CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.h
```diff
@@ -20,11 +20,7 @@
 #import <Cordova/CDVPlugin.h>
 #import <Cordova/CDVWhitelist.h>
 
-#if WK_WEB_VIEW_ONLY
 #define CDVWebViewNavigationType int
```
dpogue: Does this get used anywhere by WKWebView or should it be removed entirely?

CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m
```diff
@@ -95,13 +95,8 @@ - (CDVIntentAndNavigationFilterValue) filterUrl:(NSURL*)url
     return [[self class] filterUrl:url intentsWhitelist:self.allowIntentsWhitelist navigationsWhitelist:self.allowNavigationsWhitelist];
 }
 
-#if WK_WEB_VIEW_ONLY
 #define CDVWebViewNavigationTypeLinkClicked 0
 #define CDVWebViewNavigationTypeOther 5
```
dpogue: Do these get used by our WKWebView implementation?

CordovaLib/Classes/Private/Plugins/CDVLocalStorage/CDVLocalStorage.m
```diff
@@ -23,19 +23,12 @@ Licensed to the Apache Software Foundation (ASF) under one
 @interface CDVLocalStorage ()
 
 @property (nonatomic, readwrite, strong) NSMutableArray* backupInfo;  // array of CDVBackupInfo objects
-#if !WK_WEB_VIEW_ONLY
-@property (nonatomic, readwrite, weak) id <UIWebViewDelegate> webviewDelegate;
-#endif
 
 @end
 
 @implementation CDVLocalStorage
```
dpogue: I'm pretty sure that all the paths in this file for LocalStorage are UIWebView-specific. Worth a bit more investigation to be sure, but maybe this can be removed entirely.

CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.h
erisu: `CDVWebViewNavigationType` is used here.

https://github.com/erisu/cordova-ios/blob/refactor/swap-webviews/CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m#L105-L132

https://github.com/erisu/cordova-ios/blob/refactor/swap-webviews/CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m#L134-L137

CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m
erisu:
```
#define CDVWebViewNavigationTypeOther 5
```
Is a UIWebView specific value. For WKWebView, the WKNavigationType Other value is `-1` and not `5`. I have updated this value to represent the correct value for WKWebView.

The result of this value is used in the `shouldOverrideLoadWithRequest` method.

It is what determines if a link should load when coming from something other than an anchor tag. For example using `location.href` in JS. It is apart of the `allow-intent` stuff.

bin/templates/scripts/cordova/lib/prepare.js
```diff
@@ -310,22 +308,6 @@ function handleBuildSettings (platformConfig, locations, infoPlist) {
...
     }
-    if (wkWebViewOnly) {
-        let wkwebviewValue = '1';
-        if (wkWebViewOnly === 'true') {
-            events.emit('verbose', 'Set WK_WEB_VIEW_ONLY.');
-        } else {
-            wkwebviewValue = '0';
-            events.emit('verbose', 'Unset WK_WEB_VIEW_ONLY.');
-        }
-        project.xcode.updateBuildProperty('WK_WEB_VIEW_ONLY', wkwebviewValue);
```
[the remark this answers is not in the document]
erisu: Flag removed from JS. Flag has been hardcoded, to true, to ensure that the IAB plugin is forced to use WKWebView in this release.

jpike88:
@erisu now that this has been merged when can we look forward to next release?

brodybits:
If you need this update now, I think your best bet at this point would be to install from GitHub, for example:

- `cordova platform add github:apache/cordova-ios`
- `cordova platform add github:apache/cordova-ios#bce99f94`

There should also be a nightly tag on npm. In general, we do not actively support unstable versions from master or from nightly.

My understanding is that there are some breaking changes still in progress before we are ready to publish the new release.

In case you have any questions about the upcoming release I would highly recommend you followup with us on Slack or on the mailing list, follow links in the footer of cordova.io or cordova.apache.org.

PR #780, opened 2020-02-10
breaking: upgrade xcode compatible to 11.0

erisu:
### Motivation and Context

Upgrade project, CordovaLib, and CordovaLib test to be Xcode 11.0 compatible.

Even though next major will support iOS 11 minimum, which was introduced in `Xcode 9.x`, the goal of this PR is to set the version of Xcode that Apple is enforcing all apps to be built with for app submission starting April 2020. 

Keeping it modern, or as modern as possible with our target release.

[WWDC 2019 Video](https://developer.apple.com/videos/play/wwdc2019/239/), as Apple says, around `6:15` mark:

> Keeping your project file modern is a critical way to make sure that Xcode can help you out and avoids the accumulation of issues.
> First, when you're updating to a new version of Xcode, you'll be offered the opportunity sometimes to have **Xcode update the project settings and update your project file to the latest format**.

### Description

* Change the `Project Format` to `Xcode 11.0-compatible`. 
* Rebuilt the SampleApp's `project.pbxproj` with the current PR's changes.
* Ensure that values in `project.pbxproj` that contains  `__PRODUCT_NAME__` has quotations.

### Testing

- `npm t`
- `cordova platform add`
- `cordova build ios`

### Checklist


jcesarmobile:
Why Xcode 9? Apple requires Xcode 10 for submissions, and soon (by April) they will require Xcode 11.
So maybe we can target Xcode 11 already and remove Xcode 10 and earlier from tests

erisu:
I forgot all about the new submission rules from Apple. I will go ahead and update to Xcode 11 to follow the submission rule which is described on this page.

https://developer.apple.com/app-store/submissions/

> Starting April, 2020, all apps submitted to the App Store will need to be built with Xcode 11. Xcode 11 requires macOS Mojave 10.14.3 or later.

BBosman:
I would have expected the “LastUpgradeCheck” to reference the XCode 11 level as well?

erisu:
The changes in this PR were made though the Xcode interface. Nothing was manually changed except to re-add back quotation wrappers around the product name for UTF-8 support. I believe  `LastUpgradeCheck` shouldn't be manually changed.

Xcode should have updated it if it was needed. It was not changed because it was not though an upgrade notification.

If we manually changed it, we would need to know the correct value.
* What would be the side-effect of changing it manually?
* If not everything was upgraded and it was manually changed it, would it be basically disabling potentially the only flag used to determining to display an upgrade or to make additional missing changes that are needed?

Anyways, I think it is safe to keep it as is and wait till next time Xcode automatically updates this for us.

erisu:
Last call, this PR will be merged with or without a review in 24 hours.

PR #782, opened 2020-02-12
fix (node): require failure with shebang interpreter

erisu:
### Motivation and Context

Fix failing tests.
Bin files that contains the shebang interpreter fails when being required.

Example Error:

> /Users/travis/build/apache/cordova-ios/bin/templates/scripts/cordova/lib/list-devices:1
(function (exports, require, module, __filename, __dirname) { #!/usr/bin/env node
                                                              ^
> SyntaxError: Invalid or unexpected token

https://travis-ci.org/apache/cordova-ios/jobs/649198385#L1474

### Description

* Keep the bin files with the shebang interpreter but pull out all the code into a separate JS file. The bin files require the newly created JS files and execute them as usual.

* Update everywhere else that use to require the bin but now require the newly created JS files which do not contain the shebang interpreter.

**Additional Notes:**

The version bin file, which contains the platform version, is updated by coho during release. Coho should now update the version in `Api.js` instead. It would be better if the version is pulled from `package.json` but the platform-centered workflow does not have this file. Additional investigation to improve this should be done but preferably outside of this PR.

### Testing

- `npm t`

### Checklist

### DIFF
--- a/bin/lib/create.js
+++ b/bin/lib/create.js
@@ -1,5 +1,3 @@
-#!/usr/bin/env node
-
 /*
     Licensed to the Apache Software Foundation (ASF) under one
     or more contributor license agreements. See the NOTICE file
--- a/bin/templates/scripts/cordova/Api.js
+++ b/bin/templates/scripts/cordova/Api.js
@@ -19,6 +19,15 @@
 
 /* jslint node: true */
 
+/**
+ * @todo update coho to update this line.
+ * @todo use `package.json` instead but first
+ *  figure out how this fit in with the platform-centered workflow structure.
+ *  This workflow would not have the `package.json` file.
+ */
+// Coho updates this line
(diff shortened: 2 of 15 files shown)


erisu:
Additional Feedback:

After more investigation the issue seems to be a combination of using the new node with the 2 year old `rewire` npm package.

I was able to recreate the issue here: https://github.com/erisu/shebang-test

erisu:
I went ahead and merged this PR to unblock other PRs from the failing tests. Because of `rewire` has not been maintained or updated in the past 2 years, it would be best to look in alternatives or how to test without rewire.

NiklasMerz:
> I went ahead and merged this PR to unblock other PRs from the failing tests. Because of `rewire` has not been maintained or updated in the past 2 years, it would be best to look in alternatives or how to test without rewire.

Should we create an issue for that?

PR #784, opened 2020-02-12
refactor: deprecate mediaPlaybackRequiresUserAction

erisu:
### Motivation and Context

Apple has deprecated the `mediaPlaybackRequiresUserAction` property in iOS 8.0–9.0.
This property should be replaced with `mediaTypesRequiringUserActionForPlayback`.

https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/1614727-mediaplaybackrequiresuseraction

### Description

This PR:
* Updates the native side to use the new property name.
* Updates to read the new preference name (`MediaTypesRequiringUserActionForPlayback`) from `config.xml`
* Warn users that the preference is being deprecated in the near future if it is still being used.
* Fallback on the the value from the old preference name if it still remains.

### Testing

- `npm t`
- `cordova platform add`
- `cordova build ios`

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -74,7 +74,15 @@
     }
 
     configuration.allowsInlineMediaPlayback = [settings cordovaBoolSettingForKey:@"AllowInlineMediaPlayback" defaultValue:NO];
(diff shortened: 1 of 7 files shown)


jcesarmobile:
Sorry for the late review, but mediaTypesRequiringUserActionForPlayback doesn’t accept true or false values, if true it should just do nothing, if false it should set the value to WKAudiovisualMediaTypeNone.
For false is not a real problem since false and WKAudiovisualMediaTypeNone have 0 value, but setting to true will be unexpected.

PR #785, opened 2020-02-12
refactor: deprecate mediaPlaybackAllowsAirPlay

erisu:
### Motivation and Context

Apple has deprecated the `mediaPlaybackAllowsAirPlay` property in iOS iOS 8.0-9.0.
This property should be replaced with `allowsAirPlayForMediaPlayback`.

https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/1614726-mediaplaybackallowsairplay

### Description

This PR:
* Updates the native side to use the new property name.
* Updates to read the new preference name (`AllowsAirPlayForMediaPlayback`) from `config.xml`
* Warn users that the preference is being deprecated in the near future if it is still being used.
* Fallback on the the value from the old preference name if it still remains.

### Testing

- `npm t`
- `cordova platform add`
- `cordova build ios`

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -91,7 +91,7 @@
      */
     BOOL allowsAirPlayForMediaPlayback = [settings cordovaBoolSettingForKey:@"MediaPlaybackAllowsAirPlay" defaultValue:YES];
     if([settings cordovaSettingForKey:@"AllowsAirPlayForMediaPlayback"] != nil) {
-        allowsAirPlayForMediaPlayback = [settings cordovaBoolSettingForKey:@"MediaPlaybackAllowsAirPlay" defaultValue:YES];
+        allowsAirPlayForMediaPlayback = [settings cordovaBoolSettingForKey:@"AllowsAirPlayForMediaPlayback" defaultValue:YES];
     }
     configuration.allowsAirPlayForMediaPlayback = allowsAirPlayForMediaPlayback;
 
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -84,7 +84,17 @@
     configuration.mediaTypesRequiringUserActionForPlayback = mediaTypesRequiringUserActionForPlayback;
 
     configuration.suppressesIncrementalRendering = [settings cordovaBoolSettingForKey:@"SuppressesIncrementalRendering" defaultValue:NO];
-    configuration.mediaPlaybackAllowsAirPlay = [settings cordovaBoolSettingForKey:@"MediaPlaybackAllowsAirPlay" defaultValue:YES];
+
+    /*
+     * If the old preference key "MediaPlaybackAllowsAirPlay" exists, use it or default to "YES".
+     * Check if the new preference key "AllowsAirPlayForMediaPlayback" exists and overwrite the "MediaPlaybackAllowsAirPlay" value.
+     */
+    BOOL allowsAirPlayForMediaPlayback = [settings cordovaBoolSettingForKey:@"MediaPlaybackAllowsAirPlay" defaultValue:YES];
+    if([settings cordovaSettingForKey:@"AllowsAirPlayForMediaPlayback"] != nil) {
+        allowsAirPlayForMediaPlayback = [settings cordovaBoolSettingForKey:@"MediaPlaybackAllowsAirPlay" defaultValue:YES];
+    }
+    configuration.allowsAirPlayForMediaPlayback = allowsAirPlayForMediaPlayback;
+
     return configuration;
 }
 
--- a/bin/templates/scripts/cordova/lib/prepare.js
+++ b/bin/templates/scripts/cordova/lib/prepare.js
@@ -522,7 +522,8 @@ function updateFileResources (cordovaProject, locations) {
 
 function alertDeprecatedPreference (configParser) {
     const deprecatedToNewPreferences = {
-        MediaPlaybackRequiresUserAction: 'MediaTypesRequiringUserActionForPlayback'
+        MediaPlaybackRequiresUserAction: 'MediaTypesRequiringUserActionForPlayback',
+        MediaPlaybackAllowsAirPlay: 'AllowsAirPlayForMediaPlayback'
     };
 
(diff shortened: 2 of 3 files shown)


CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
```diff
@@ -74,9 +74,25 @@ - (WKWebViewConfiguration*) createConfigurationFromSettings:(NSDictionary*)setti
...
-    configuration.mediaPlaybackRequiresUserAction = [settings cordovaBoolSettingForKey:@"MediaPlaybackRequiresUserAction" defaultValue:YES];
+
+    // Check for usage of the older preference key, alert, and use.
+    BOOL mediaTypesRequiringUserActionForPlayback = [settings cordovaBoolSettingForKey:@"MediaTypesRequiringUserActionForPlayback" defaultValue:YES];
+    NSString *mediaPlaybackRequiresUserActionKey = [settings cordovaSettingForKey:@"MediaPlaybackRequiresUserAction"];
+    if(mediaPlaybackRequiresUserActionKey != nil) {
+        mediaTypesRequiringUserActionForPlayback = [settings cordovaBoolSettingForKey:@"MediaPlaybackRequiresUserAction" defaultValue:YES];
+    }
+    configuration.mediaTypesRequiringUserActionForPlayback = mediaTypesRequiringUserActionForPlayback;
+
```
knight9999: These codes are for `mediaTypeRequiringUserActionForPlayback` which (already) included in other PR https://github.com/apache/cordova-ios/pull/784. 
Is this necessary for `allowsAirPlayForMediaPlayback` ?

bin/templates/project/__PROJECT_NAME__/config.xml
```diff
@@ -53,7 +53,7 @@
     <preference name="DisallowOverscroll" value="false" />
     <preference name="EnableViewportScale" value="false" />
     <preference name="KeyboardDisplayRequiresUserAction" value="true" />
-    <preference name="MediaPlaybackRequiresUserAction" value="false" />
+    <preference name="MediaTypesRequiringUserActionForPlayback" value="false" />
```
knight9999: This is for `mediaTypesRequiringUserActionForPlayback` which (already) included in other PR https://github.com/apache/cordova-ios/pull/784.
Or the `AllowsAirPlayForMediaPlayback` property here?

bin/templates/scripts/cordova/defaults.xml
```diff
@@ -27,7 +27,7 @@
     <preference name="DisallowOverscroll" value="false" />
     <preference name="EnableViewportScale" value="false" />
     <preference name="KeyboardDisplayRequiresUserAction" value="true" />
-    <preference name="MediaPlaybackRequiresUserAction" value="false" />
+    <preference name="MediaTypesRequiringUserActionForPlayback" value="false" />
```
knight9999: This is for mediaTypesRequiringUserActionForPlayback which (already) included in other PR #784.
Or the AllowsAirPlayForMediaPlayback property here?

bin/templates/project/__PROJECT_NAME__/config.xml
dpogue: @erisu FYI, `value="false"` will not work the way you think it will 🙃 

Essentially, `cordovaBoolSettingForKey` expects the value to be a 0 or 1, and turns anything other non-zero value truthy. I kinda consider this to be a bug in `cordovaBoolSettingForKey` tbh.

erisu: Read the last commit. First commit is from the last Pr. This was built off of the last Pr and should be merged first.

erisu: Thank you for bringing this up.

I will take a look at how to improve and fix this behavior so it works the way we expect.

Since my changes are just updating the preference name, for the most part, and that the behavior you are bringing up is applied to all boolean preferences, I will handle it in a separate PR.

bin/templates/scripts/cordova/defaults.xml
erisu: Same as I mentioned before.

jcesarmobile:
If both the old and the new preference are present the old one takes precedence , is that desired?

erisu:
@jcesarmobile When you have a chance, can you take a look at this PR again please. I flipped the order of priority.

Now, the old preference value should be set first, or defaults to "YES" when missing, but if the new preference name exists, it will take higest priority and overwrite whatever was set from the old preference name.

CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
```diff
@@ -84,7 +84,17 @@ - (WKWebViewConfiguration*) createConfigurationFromSettings:(NSDictionary*)setti
...
     configuration.suppressesIncrementalRendering = [settings cordovaBoolSettingForKey:@"SuppressesIncrementalRendering" defaultValue:NO];
-    configuration.mediaPlaybackAllowsAirPlay = [settings cordovaBoolSettingForKey:@"MediaPlaybackAllowsAirPlay" defaultValue:YES];
+
+    /*
+     * If the old preference key "MediaPlaybackAllowsAirPlay" exists, use it or default to "YES".
+     * Check if the new preference key "AllowsAirPlayForMediaPlayback" exists and overwrite the "MediaPlaybackAllowsAirPlay" value.
+     */
+    BOOL allowsAirPlayForMediaPlayback = [settings cordovaBoolSettingForKey:@"MediaPlaybackAllowsAirPlay" defaultValue:YES];
+    if([settings cordovaSettingForKey:@"AllowsAirPlayForMediaPlayback"] != nil) {
+        allowsAirPlayForMediaPlayback = [settings cordovaBoolSettingForKey:@"MediaPlaybackAllowsAirPlay" defaultValue:YES];
```
jcesarmobile: Should check AllowsAirPlayForMediaPlayback , right?

```suggestion
        allowsAirPlayForMediaPlayback = [settings cordovaBoolSettingForKey:@"AllowsAirPlayForMediaPlayback" defaultValue:YES];
```

erisu: No, this is correct..

It suppose to read the old preference option first and store its to be used if uses still sets the old option, not the new one. But it will though still set the new flag. If the new preference value is set when the old one is or is not set, the value stored is trash for the new preference options.

If this is too complicated, then I want to completely trash the old preference option and done with it. I don't care if the old preference option is completely removed with small step for migration. If the app stops working as intended, they should still be able to figure it out with the build warnings.

PR #794, opened 2020-02-18
style: remove extra space from Objective-C test modules

brodybits:


erisu:
LGTM.

Sorry I made some changes to the PR entry.

I don't find this as a bug as it does not break tests or built apps in anyway. It is more of a styling change and per available tags, `enhancement` IMO fits best.

Also for title, though this can be debated and as we don't really have a hard requirement, I changed it to `style`. It is shorter and fits the type of changes.

I been using recently these as a guide but slightly modified.

https://karma-runner.github.io/1.0/dev/git-commit-msg.html
https://www.conventionalcommits.org/en/v1.0.0/

brodybits:
Thanks @erisu, you should see that I just did a quick merge. I actually noticed the title update after the merge, cool 😎(I generally like to do this kind of change as fast as possible, I was in the middle of testing for something else.)

> I been using recently these as a guide but slightly modified.

+1, nice; I wouldn't mind if we would discuss these somewhere other than a closed PR :)

Issue #793, opened 2020-02-18
*-Info.plist & config.xml - general


brodycj:
__updated:__

There seem to be quite a few outstanding issues with *-Info.plist and config.xml, not sure if they are all the same or not:

- <https://github.com/apache/cordova-common/issues/88>
- #764
- #775
- #826 - testing: removing an internal *-Info.plist filter does not cause a test failure (BREAKING TEST PR #792 does not cause a failure)

I personally do not understand this mechanism so well, raising this issue with the **help wanted** label for now. In case this issue is a duplicate, we can close it.

brodybits:
I wonder if <https://github.com/apache/cordova-common/pull/148> may be related

PR #795, opened 2020-02-19
[minor] fix *-Info.plist & config.xml functionality

### LINKED ISSUE #764: could not find -Info.plist file, or config.xml file after adding extension, cordova build, with FIX
# Bug Report

I have included the fix on the bottom. I am pretty new at Cordova, someone else should take a look.

## Problem
Cannot find `*-info.plist or config.xml` error when adding extension in Xcode. In a new cordova repo. After an extension is installed, depends on how you name the extension in Xcode, it will trigger a build error when running `cordova build ios`

## How to reproduce

1. Start a fresh repo
2. Run all the cordova install, build commands
3. Open xcode, add extension, name it `OneTwoThree`
4. Run cordova build
5. Throws here

## How did we solve it

In `ios/cordova/lib/projectFile.js`, we found this.

```
var xcBuildConfiguration = xcodeproj.pbxXCBuildConfigurationSection();
  var plist_file_entry = _.find(xcBuildConfiguration, function(entry) {
    return entry.buildSettings && entry.buildSettings.INFOPLIST_FILE;
  });
```

Log `plist_file_entry` returns the path of the `OneTwoThree` extension, instead of the project path.

We fixed it by adding one more condition, it basically search through that path, and find your project name by finding a file name ending with `.xcworkspace`. Not sure how reliable is this.

```
var projectName = fs
    .readdirSync(project_dir)
    .find(d => d.includes(".xcworkspace"))
    .replace(".xcworkspace", "");

 var xcBuildConfiguration = xcodeproj.pbxXCBuildConfigurationSection();
 var plist_file_entry = _.find(xcBuildConfiguration, function(entry) {
    return (
      entry.buildSettings &&
      entry.buildSettings.INFOPLIST_FILE &&
      entry.buildSettings.INFOPLIST_FILE.includes(projectName)
    );
 });
```

IOS

### Version information
Latest

## Checklist

brodycj:
This is proposed to include and supersede PR #765, as a solution for #764, with some additional fixes and test updates that I think should be included at the same time.

- updates and improvements to unit testing of projectFile parse method, now with xcworkspace included in the "ios-config-xml" fixture
- include all changes from PR #765 so far, with possible solution to issue #764
- some followup fixes to PR #765
- add test cases for the code updated in PR #765
- some more general improvements

This proposal should be considered breaking since it now depends on existence of xcworkspace with the correct name. This should have been present in generated platforms/ios project for years.

I have discussed some other related issues with plists in #793, not sure whether or not this proposal would resolve any such related issues.

I would like to have multiple reviews before getting this proposal merged. I think it should be good to merge this with a squash merge, with all co-authored-by credits that I gave in 8dfeedc8f754f7638b96b6677a0f34a4d246c8c5, maybe with a rebase to cleanup some of the items that will go into the commit message.

closes #765 (PR #765)
resolves #764

/cc @leogoesger


erisu:
@brodybits Is this ticket fixing a bug and also considered to be breaking because you are searching for `xcworkspace`?

What if you change `xcworkspace` search to `xcodeproj`?

Would this still fix the issue and change the PR's type from breaking to patch or minor?

There is a patch release (`6.0.1`) being prepared. If this PR is resolving a bug and can be altered to be a patch instead of breaking, you might be able to get this PR in.

You could still create a second breaking PR after this is merged into master to change only the `xcodeproj` to `xcworkspace`...

brodybits:
The good news is that I would no longer consider this to be a breaking change. If I would remove the xcworkspace file and then build from the command line, it results in an error from xcodebuild. (I was able to remove the xcworkspace file and then build & run from Xcode, which I think is a side point.)

I do think this is too much change for a patch, though, should probably target a minor release.

I did try merging. The changes did not seem to work after merging, and some of the tests need rework due to removing shelljs. I think we should really rebase and rework. I will try to fit this in as soon as I can.

brodybits:
I now wonder if <apache/cordova-common/pull/148> may be related

Menardi:
@brodybits I can confirm that the PR you mentioned does not fix this issue, as I am still seeing this in `cordova-ios@6.1.1`. Though the cause of the issues is the same (having multiple `Info.plist`) files, their fixes are separate.

bin/templates/scripts/cordova/lib/projectFile.js
```diff
@@ -40,13 +40,43 @@ function parseProjectFile (locations) {
     const xcodeproj = xcode.project(pbxPath);
     xcodeproj.parseSync();
 
+    const workspaceName =
+        fs.readdirSync(project_dir).find(d => d.includes('.xcworkspace')) || '';
+
+    const projectName = workspaceName.replace('.xcworkspace', '');
```
Menardi, line 46: I've noticed we have a [`getIOSProjectname` function in `cordova-common`](https://github.com/apache/cordova-common/blob/87167bd45670291b68e6e02852a6b4faf63ed247/src/ConfigChanges/ConfigFile.js#L233), maybe you could use it here?

PR #783, opened 2020-02-12
breaking (CDVURLProtocol): delete

erisu:
### Motivation and Context

Remove UIWebView Specific Code

### Description

Delete `CDVURLProtocol`

### Testing

* `npm t`

### Checklist

- [x] I've run the tests to see all new and existing tests pass

PR #801, opened 2020-02-28
breaking (UserAgent): Drop CDVUserAgentUtil and Implement for WKWebView

erisu:
### Motivation and Context

* Drop `CDVUserAgentUtil` which handled managing the User Agent string for `UIWebView`.
* Add alternative implementation for `WKWebView`.

### Description

* Remove all existing UserAgent code that uses or associated with `CDVUserAgentUtil`.
* Add to the WKWebView Engine configuration setup to set the user agent string.

### Testing

- `npm t`
- `cordova platform add`
- `cordova build`
- Run in Simulator
- Tested below test cases in `config.xml`

Test Cases
- [x] `AppendUserAgent`
- [x] `AppendUserAgent` w/ `OverrideUserAgent `
- [x] `OverrideUserAgent `

### Checklist


erisu:
@jcesarmobile I re-added back the `wkWebView.customUserAgent` for the `OverrideUserAgent` preference.

* left the comment for `OverrideUserAgent` with the other configurations so they are gathered in one area. 
* updated the condition in the configuration setup method to still skip the append preference when override is set.
* add the `OverrideUserAgent` condition in the `pluginInitialize` method where the `wkWebView` instance is created and availabe.

CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -304,45 +301,12 @@ - (void)viewDidLoad
...
     }
 
-    // /////////////////
-    NSURL* appURL = [self appUrl];
-    __weak __typeof__(self) weakSelf = self;
-
-    [CDVUserAgentUtil acquireLock:^(NSInteger lockToken) {
-        // Fix the memory leak caused by the strong reference.
-        [weakSelf setLockToken:lockToken];
-        if (appURL) {
```
jcesarmobile: I don't think the logic inside the acquireLock block should be removed, it seems to handle the case where the appUrl is invalid and loads an error page instead

CVeniamin:
Went ahead and checked which plugins might be affected by this change. 
I found that both [cordova-plugin-file-transfer](https://github.com/apache/cordova-plugin-file-transfer/blob/dc8fbd78a234bd88450ebe01835c60371bc2cf8e/src/ios/CDVFileTransfer.m#L107) and [cordova-plugin-wkwebview-engine](https://github.com/apache/cordova-plugin-wkwebview-engine/blob/3be1b118f4b629f9a770821ebaa06c0129ea7590/src/ios/CDVWKWebViewEngine.m#L104) depend on the previous API. 

Although `cordova-plugin-wkwebview-engine` should now be "deprecated" and the change should not affect it as much since the WKWebview is the default engine.

jcesarmobile:
cordova-plugin-file-transfer is deprecated too

about `cordova-plugin-wkwebview-engine`, we were going to recommend using it for those who want to use file urls instead of the custom schemes, so maybe should be fixed to work on cordova-ios 6

CVeniamin:
@jcesarmobile, thanks for the feedback!
I know that there are some use cases of `cordova-plugin-file-transfer` being used to play tracks offline combined with `cordova-plugin-media` for music players, that is why I have mentioned it.

tryhardest:
Hopefully we can still use these plugins and upgrade  specifically file transfer plugin. Sure would love to see anyone with authority release a new version of Ionic WKwebview plugin.

vitto32:
> about `cordova-plugin-wkwebview-engine`, we were going to recommend using it for those who want to use file urls instead of the custom schemes, so maybe should be fixed to work on cordova-ios 6

I'm in that situation. #783 and this one are making things complicated.

I've spent a lot of time in the past replacing `cordova-plugin-file-transfer`, and I came up to a solution using `cordova-plugin-wkwebview-file-xhr`.

Now I'm trying to remove `wkwebview-engine` and `wkwebview-file-xhr` but I'm facing CORS issues when trying (in developement mode) to access file:// or app://  locations.

PR #810, opened 2020-03-19
Remove dead link to wiki from README

mbeijen:
The wiki is moved to Confluence, and there is no Cordova content there. I've also checked the Cordova main site and there are no links to a wiki from there.


NiklasMerz:
Thank you very much 👍

Issue #821, opened 2020-04-03
shouldAutorotateToInterfaceOrientation was deprecated on iOS 6


jcesarmobile:
we should remove remaining usages of shouldAutorotateToInterfaceOrientation as it was deprecated on iOS 6

PR #822, opened 2020-04-05
chore: remove deprecated orientation methods

jcesarmobile:
`- (NSUInteger)application:(UIApplication*)application supportedInterfaceOrientationsForWindow:(UIWindow*)window`, `- (NSUInteger)supportedInterfaceOrientations;` and  `- (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)interfaceOrientation` were deprecated on iOS 6, remove the code as is not being used.

I think it's not breaking since we have the new methods that replaced them.

closes https://github.com/apache/cordova-ios/issues/821

### DIFF
--- a/CordovaLib/Classes/Public/CDVAppDelegate.m
+++ b/CordovaLib/Classes/Public/CDVAppDelegate.m
@@ -86,11 +86,7 @@
     return YES;
 }
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED < 90000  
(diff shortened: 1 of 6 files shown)


jcesarmobile:
I've changed the title to "breaking" as it will break plugins using the orientation delegate, not sure how many are using it.
I updated InAppBrowser to remove the method there.

Since it's already merged and the breaking wasn't included in the commit I hope we don't forget about it.

PR #850, opened 2020-05-16
breaking: fix file's chmod

erisu:

Set files with correct chmod permissions: 
* Executable files: `755`
* Logic files: `644`

### Testing

- `npm t`
- `platform add`
- `build`

### Checklist


NiklasMerz:
Just out of curiosity. Why was this change necessary?

erisu:
So that the binaries are executable/usable.

This is apart of removing this from code:

https://github.com/apache/cordova-ios/blob/master/bin/lib/create.js#L103-L105

PR #851, opened 2020-05-16
breaking: replace shelljs with fs-extra

erisu:

- Remove `shelljs` dependency for `fs-extra`.

### Testing

- `npm t`

### Checklist


bin/lib/utils.js
```diff
@@ -0,0 +1,65 @@
...
+ * @param {RegExp} searchRegex The search regex
+ * @param {string} replacementString The string to replace the found occurences
+ * @returns {void}
+ */
+exports.replaceFileContents = function (file, searchRegex, replacementString) {
+    let contents;
+    try {
+        contents = fs.readFileSync(file).toString();
+    } catch (ex) {
+        console.log('TRYING TO READ: ', file);
```
breautek, line 42: This I made the comment on the wrong PR, but I think this should be using the cordova log system so this is only printed during `-v` verbose mode.

PR #860, opened 2020-05-20
breaking: drop q dependency

erisu:

* Remove direct requirement of `Q` dependency from `cordova-ios`
* Keep the `Q.progress` calls which is available from `superspawn`.
* Use Native Promises

Requires PR #851 & #859

### Testing

- `npm t`

### Checklist


breautek:
Also note that I ignored anything related to the prerequisite PRs.

PR #866, opened 2020-05-25
refactor: default to file scheme

erisu:
### Motivation and Context

Use `file` as default scheme to avoid break in current apps/

### Description

1. defaults to `file` if preference `scheme` is not set.
2. continue to uses `file` if preference scheme is set to `file`.
3. use what ever value the user defines in as preference `scheme` as a custom scheme handler, excluding `file` which was defined in the above case 2.
4. preference `hostname` is only used when `scheme` is not `file`.
5. `CDV_ASSETS_URL` is set to `<scheme>://<hostname>` 
  -> if `file`, `hostname = “”`.
  -> if `!file`, `hostname = "localhost"` or user define value for preference `hostname`.
6. Extension of above case 3, if `scheme` is set and is not `file`, it will also validate against `WKWebView handlesURLScheme`, if it is not valid, it will default to `app`
 
### Testing

- cordova build
- run in simulator

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVIntentAndNavigationFilter/CDVIntentAndNavigationFilter.m
@@ -45,10 +45,16 @@
 
 - (void)parserDidStartDocument:(NSXMLParser*)parser
 {
-    NSString* scheme = [NSString stringWithFormat:@"%@://",((CDVViewController*)self.viewController).appScheme];
     // file: url <allow-navigations> are added by default
     // navigation to the scheme used by the app is also allowed
-    self.allowNavigations = [[NSMutableArray alloc] initWithArray:@[ @"file://", scheme ]];
+    self.allowNavigations = [[NSMutableArray alloc] initWithArray:@[ @"file://"]];
+
+    // If the custom app scheme is defined, append it to the allow navigation as default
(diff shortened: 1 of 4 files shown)


rollinq:
I have a hard time understanding how to use custom scheme with cordova-ios 6, specifically the behaviour introduced here :

In the description of the behaviour by erisu, it says : 

> it will also validate against WKWebView handlesURLScheme, if it is not valid, it will default to app

But in the code, the actual condition is :

```
if(scheme == nil || [WKWebView handlesURLScheme:scheme]){
     scheme = @"app";
}
```

which, to my understanding of Objective-C, means that if it IS valid, it will default to app, cf. [WKWebView handlesURLScheme documentation here](https://developer.apple.com/documentation/webkit/wkwebview/2875370-handlesurlscheme?language=objc)

Could someone clarify this to me ?

jcesarmobile:
`handlesURLScheme` returns true if the webview already handles the scheme, if the webview already handles that scheme, it means we can't use that scheme for serving the app content, so if you set the scheme to http, https, file or a few others, `handlesURLScheme` is true and we default to `app`, if it returns false, it means whatever you configured is safe to use.

lovelyelfpop:
Cordova-ios@6.1.0 says 'refactor: default to file scheme', why the url of my app is still `myscheme://` not 'file://'?

breautek:
@lovelyelfpop

See https://github.com/apache/cordova-ios/pull/866#issuecomment-650176421

PR #878, opened 2020-05-30
Replaced references to Cocoapods repository

enricodente:
Updated with new self-hosted CDN instead of GitHub, due to GitHub rate limiting solves issue when running 'pod install'.

Official reference: https://blog.cocoapods.org/CocoaPods-1.7.2/
SO reference: https://stackoverflow.com/questions/23755974/cocoapods-pod-install-takes-forever

### Platforms affected
Apple, building for iOS

### Motivation and Context
Everytime 'pod install' was requested, the GitHub repository rate made the cloning too long. The team moved to a self-hosted CDN (which is also the main one now), that solves the issue with very fast cloning.

### Description
Replaced all occurrences of 'https://github.com/CocoaPods/Specs.git' with 'https://cdn.cocoapods.org/' (will take effect in the content of the generated Podfile).

### Testing
I had to build a project that won't compile if I could not run 'pod install'. After the changes I could.

### Checklist

- [x ] I've run the tests to see all new and existing tests pass
- [x ] I added automated test coverage as appropriate for this change
- [x ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [x ] I've updated the documentation if necessary

### DIFF
--- a/tests/spec/unit/Api.spec.js
+++ b/tests/spec/unit/Api.spec.js
@@ -143,7 +143,7 @@ describe('Platform Api', () => {
                     },
                     sources: {
                         'https://github.com/sample/SampleSpecs.git': { source: 'https://github.com/sample/SampleSpecs.git' },
-                        'https://github.com/CocoaPods/Specs.git': { source: 'https://github.com/CocoaPods/Specs.git' }
+                        'https://cdn.cocoapods.org/': { source: 'https://cdn.cocoapods.org/' }
                     },
                     libraries: {
                         AFNetworking: {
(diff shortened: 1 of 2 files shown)


erisu:
Just making a note that these changes are only applied to the test specs. This has no affect to the actual production code. The change only modifies mock code so pods is not actually being executed in the test suite. Fake calls.

> I had to build a project that won't compile if I could not run 'pod install'. After the changes I could.

The statement above is incorrect. The changes can not change the outcome of a compile.

enricodente:
Sorry for that, I thought I had replaced all the occurrences to the old repository, not only in unit tests.

As I am running the latest "production" cordova build from npm and encountered the issue, I thought that by replacing all the occurrences with a batch execution would be equivalent to what I did manually changing it on my Podfile, but apparently that's not the case. I am wondering now... were is that reference to the Github repository then?

erisu:
If you are using an older version of CocoaPods, it could be using GitHub by default. 

If you are using version 1.7.2+ and lower than 1.8.0, you can configure CocoaPods to use CDN.

In the latest Cordova iOS release 6.0.0, we bumped the minimum required version of CocoaPods to 1.8.0, which uses CDN by default.

Now, if the issue is not related to the tooling itself, for example if you are using 1.8.0+, it is possible that one of the plugins you are loading has set the source to GH.

For example, if you read our docs on how to configuring the `plugin.xml` for building a plugin.
https://cordova.apache.org/docs/en/latest/plugin_ref/spec.html#pod

A plugin developer could have written something like this.

```xml
    <podspec>
      <config>
        <source url="https://github.com/CocoaPods/Specs.git"/>
      </config>
      <pods use-frameworks="true">
        <pod name="AFNetworking" spec="~> 3.2" />
      </pods>
    </podspec>
```

In that example, they explicitly set the GH url as one of the sources to check. Added the source is not required if they are using the default repo/CDN.

PR #896, opened 2020-06-10
Set the background colour as an xcasset color set

### LINKED ISSUE #890: Problem with splashscreen and cordova-ios 6.0.0
# Bug Report

## Problem

The version 6.0.0 of cordova-ios integrates the  [splashscreen plugin](https://github.com/apache/cordova-plugin-splashscreen). Accordingly, the team of cordova-plugin-splashscreen removed the support of this version with [the latest version of the plugin](https://github.com/apache/cordova-plugin-splashscreen/blob/master/RELEASENOTES.md). 

With that in mind I decided to update cordova-ios to version 6.0.0, but the integrated splashscreen does not work the same way as the plugin one.

The splashscreen should be displayed the same way as before.

The splash screen is hidden very quickly, it does not use the settings of the config.xml file. It is hidden even if the _AutoHideSplashScreen_ option is set to false, and the _SplashScreenDelay_ option does not change the delay.

## Information

I looked at the [CDVViewController.m file](https://github.com/apache/cordova-ios/blob/master/CordovaLib/Classes/Public/CDVViewController.m) file with the Xcode debugger, and it seems that the options of the config.xml file are correctly interpreted, but it does not change nothing. 

Moreover when I call the `navigator.splashscreen.hide()` function, the splashscreen is not displayed and I cannot click on the screen anymore.

Some people from the splashscreen plugin team seem to have spotted [the same problem](https://github.com/apache/cordova-plugin-splashscreen/issues/268).

I use Xcode 11.4 and an IPhone 6S with iOS 13.5.

### Version information

I use cordova-ios 6.0.0.

## Checklist

dpogue:
# Platforms affected
iOS

### Motivation and Context
Resolves #890 where the splashscreen UIView defaults to a transparent background if no `BackgroundColor` preference is specified in config.xml.

### Description
Add a named color set to Images.xcassets named `BackgroundColor` which defaults to the system background colour (respecting dark mode). On prepare, it will be set to the value specified in config.xml. Set the LaunchScreen Storyboard to use that as the background colour, and also ensure the WebView and SplashScreen UIView also use it as the background colour.

This also hypothetically opens the door for people to (via `<resource-file>` tags) provide their own color set content with different values for different devices or different light/dark modes. Caveat that this is not in the realm of supported Cordova behaviour, and we do not make any guarantees about this working. 

If you're embedding the Cordova ViewController in another app or otherwise don't have a `BackgroundColor` color set in xcassets, it will fallback to white.

### Testing

* Manual testing against a simulator.
* Added spec tests for the `BackgroundColor` preference parsing in prepare.spec.js

### Checklist


dpogue:
@erisu Thanks for catching that. I've fixed the missing test fixture, and also added a second commit to address the SplashScreen being blank. It should work as intended now 😁

PR #973, opened 2020-08-23
fix: Append startURL to the initial url loaded

jcesarmobile:
When using a custom scheme (app in example), the initial url loaded is `app://localhost`

This PR makes the initial url loaded to append the startURL to the url, in example `app://localhost/index.html`

This makes possible to run the tests as they use `<content src="cdvtests/index.html" />`, so now the url loaded will be `app://localhost/cdvtests/index.html` instead of `app://localhost`.


erisu (approved):
👍 LGTM

Tested locally both with and without paramedic.

Waiting for @dpogue to review and merge since he will be preparing the upcoming patch release.

Issue #922, opened 2020-06-26
iOS 14 will enable Intelligent Tracking Prevention in WKWebView by default


NiklasMerz:
# Bug Report

## Problem

Apple just announced that iOS 14 will enable "Intelligent Tracking Prevention" (ITP) by default in WKWebView. This issue is open for discussion around that and possible ways to fix Cordova apps that break with these changes. Please leave a 👍 if you are having issues with this change.

## Possible solution -> Work in progress

* Add preference and features to allow users to disable ITP in settings
* Proxy HTTP(s) requests through WKUrlSchemehandler to handle cookies on the native layer.
* Get exceptions for `localhost` in WebKit?

## Code - Workarounds

Currently the best solution in my case seems to build some kind of HTTP proxy with an WKURLSchemeHandler. I did this in the WKWebView plugin a while ago to workaround any CORS restrictions. **Would this make sense to integrate into cordova-ios?**

So the app runs on `cordova://localhost` for example and you convert every external URL like `https://myserver.com/image.png` to `cordova://localhost/_httpproxy/myserver.com/image.png` and the WKURLSchemeHandler does the server request instead with native Code instead of letting the WebView doing it.

https://github.com/GEDYSIntraWare/cordova-plugin-ionic-webview/blob/proxyios/src/ios/IONAssetHandler.m#L38-L85

https://github.com/GEDYSIntraWare/cordova-plugin-ionic-webview/compare/master...GEDYSIntraWare:proxyios

## Information

[WKWebView session from WWDC 2020](https://developer.apple.com/videos/play/wwdc2020/10188/)

[Webkit bug report](https://bugs.webkit.org/show_bug.cgi?id=213510)

### Environment, Platform, Device

iOS 14 beta

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

jlsync:
It would also be very useful for these same cookies to be sent/supported on websocket connections from cordova-ios.

CGS38:
On iOS14, still stuck on this ITP problem... 

I tried to disable ITP controls by adding 
NSCrossWebsiteTrackingUsageDescription to the app’s Info.plist but no user prompt is requested at runtime. By going manually to the app authorisations it is possible in this case to deactivate this security and restore nominal operation for xhr calls with third party cookies. Not satisfactory...

Another unsuccessfully tested solution, 
addition of  

```
<plist version="1.0">
<dict>
[...]

Issue #605, opened 2019-04-18
WKWebView CORS support


gisongeorge:
Hi, I'm trying to load an HTML that is in the main Bundle with WKWebView this worked perfectly with UIWebView. Now I migrated UIWebView to WJKWebView and I have an error, the files in the HTML does not load send an error in the console: "XMLHttpRequest cannot load"

How I can resolve this problem?

vg0x01:
<plugin name="cordova-plugin-wkwebview-file-xhr" spec="^2.1.1" />

vg0x01:
cordova plugin install cordova-plugin-wkwebview-file-xhr

jcesarmobile:
That’s one of the known WKWebView issues.
Should work once we implement url scheme handlers
https://github.com/apache/cordova-ios/issues/415

In the meantime you can use ionic webview plugin that already have the url scheme handlers implemented or use a plugin like the proposed by vg0x01

PR #1004, opened 2020-10-05
Proxy requests through WKURLSchemeHandler to bypass Cookies and ITP restrictions

NiklasMerz:
### Platforms affected
iOS

### Motivation and Context

iOS became more restrictive for CORS and tracking prevention which makes it hard for some apps to communicate with their server and use cookies.

See #922 #605 and possibly many more

### Description

This PR implements a kind of "proxy" that takes requests like the normal webview and runs them through native code to make the request work without the CORS restrictions. Cookies are synced with the webview.

Originally taken from a modification in Ionics webview: https://github.com/ionic-team/cordova-plugin-ionic-webview/pull/448

**TODO**

⚠️ **The whitelist settings need to be part of this**

- [ ] Should this be part of the ios platform or better a (third) party plugin?
- [ ] Tests?
- [ ] Docs

### Testing

In a sample app for now. Use this code snippet to run code through the proxy:

`fetch(window.WkWebView.convertProxyUrl("https://cordova.apache.org"));`

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVURLSchemeHandler.h
+++ b/CordovaLib/Classes/Public/CDVURLSchemeHandler.h
@@ -25,6 +25,7 @@
 @interface CDVURLSchemeHandler : NSObject <WKURLSchemeHandler>
 
 @property (nonatomic, strong) CDVViewController* viewController;
+@property (nonatomic) Boolean isRunning;
 
 - (instancetype)initWithVC:(CDVViewController *)controller;
 
--- a/CordovaLib/Classes/Public/CDVURLSchemeHandler.m
+++ b/CordovaLib/Classes/Public/CDVURLSchemeHandler.m
@@ -94,7 +94,7 @@
                     };
                 }
 
-                // Do not use urlSchemeTask if it has been closed in stopURLSchemeTask
+                // Do not use urlSchemeTask if it has been closed in stopURLSchemeTask. Otherwise the app will crash.
                 if(self.isRunning) {
                     [urlSchemeTask didReceiveResponse:response];
                     [urlSchemeTask didReceiveData:data];
@@ -142,7 +142,7 @@
 
 - (void)webView:(nonnull WKWebView *)webView stopURLSchemeTask:(nonnull id<WKURLSchemeTask>)urlSchemeTask
 {
-
+    self.isRunning = false;
 }
 
 -(NSString *) getMimeType:(NSString *)fileExtension {
--- a/CordovaLib/Classes/Public/CDVURLSchemeHandler.m
+++ b/CordovaLib/Classes/Public/CDVURLSchemeHandler.m
@@ -37,8 +37,14 @@
 {
     NSString * startPath = [[NSBundle mainBundle] pathForResource:self.viewController.wwwFolderName ofType: nil];
     NSURL * url = urlSchemeTask.request.URL;
-    NSString * stringToLoad = url.path;
     NSString * scheme = url.scheme;
+    self.isRunning = true;
+    Boolean loadFile = true;
+    NSDictionary * header = urlSchemeTask.request.allHTTPHeaderFields;
+    NSMutableString * stringToLoad = [NSMutableString string];
+    [stringToLoad appendString:url.path];
+    NSString * method = urlSchemeTask.request.HTTPMethod;
+    NSData * body = urlSchemeTask.request.HTTPBody;
 
     if ([scheme isEqualToString:self.viewController.appScheme]) {
         if ([stringToLoad hasPrefix:@"/_app_file_"]) {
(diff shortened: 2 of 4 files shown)


adamdport:
Do you have any concerns with Apple rejecting apps that essentially "bypass" the WkWebView restrictions they've added?

NiklasMerz:
I don't have any concerns but I am not entirely sure. We are using this

in an app for over a year now without any problems. I think an Apple
engineer even suggested this as an possible solution while discussing
options how to solve it. Native apps or the xhr plugin just do it like

that, too and don't have the restrictions.

On October 7, 2020, GitHub <notifications@github.com> wrote:
> Do you have any concerns with Apple rejecting apps that essentially
> "bypass" the WkWebView restrictions they've added?
>

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/apache/cordova-ios/pull/1004#issuecomment-
705076254>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AA3LORFKKVEB6WBRU5WPNTDSJSOVBANCNFSM4SE7MS7A>.

PieterVanPoyer:
Hey Niklas

I think this is a great addition.
Not having to deal with CORS issue, will make building a Cordova app a lot nicer.
I cannot review the code as I am no ios developer, but I think the addition of code looks good structured and is rather small for such a great feature.

Will it also work with sending multipartform data to the backend? [ multipart/form-data ]
Will it also work for loading local assets like local images or translations or config jsons?

In my opinion this must be an addition to the ios platform.
This should be the default way for Http requests.

Maybe there should be a flag or setting to Opt out of the proxy. (If this is not yet foreseen) 

Kind regards
Pieter

NiklasMerz:
> Hey Niklas
> 
> I think this is a great addition.
> Not having to deal with CORS issue, will make building a Cordova app a lot nicer.
> I cannot review the code as I am no ios developer, but I think the addition of code looks good structured and is rather small for such a great feature.

Thank you for the feedback

> Will it also work with sending multipartform data to the backend? [ multipart/form-data ]
> Will it also work for loading local assets like local images or translations or config jsons?

This should work for form data too if you use the special URL. For local stuff you could use `window.WkWebView.convertFileSrc`. This has been part of the platform for a while now.

> In my opinion this must be an addition to the ios platform.
> This should be the default way for Http requests.
> 
> Maybe there should be a flag or setting to Opt out of the proxy. (If this is not yet foreseen)

They way it is built now the "proxy" will only work if you convert you URL like `https://mysite.com` with `convertProxyUrl` to `app://myscheme/_https_proxy_mysite.com`. If you use normal URLs the webview will work just like it used too.

> Another remark is the ability to inspect the HTTP traffic. We are used to inspect it with the network inspector of the browser, but this will not be possible anymore.
> Maybe there must be an addition in some documentation about how to inspect the network traffic with this setup.

This is no problem somehow. All requests done with the special URLs appear in the inspector as usual.

> Kind regards
> Pieter

razam217:
@NiklasMerz Can you please suggest me the steps I should follow to use this PR in my app?

NiklasMerz:
@razam217  Sure. But you need to be aware that this is work in progress and not part of any release yet. It would be nice if you could test it and give feedback. This PR may have bugs or change in the future.

At first you need to remove your iOS platform `cordova platform rm ios`. Then you can add the platform from the branch this PR is based of `cordova platform add https://github.com/GEDYSIntraWare/cordova-ios#proxy`. If you want to use the proxy now for an URL you need to call this helper function to convert the URL `window.WkWebView.convertProxyUrl(myURL)`. This converts the URL to a special URL (using a custom scheme) and the proxy will handle it if you use the URL in an XHR, fetch  tag etc. You can still inspect the network tab in the Safari console to check your requests.

Good luck!

vvsou862:
Hi @NiklasMerz ,  I have tried the above mentioned approach in our app and i am able to load the application using cookies.
 Thank you for the update. I have found two observations(issues). 
1. When i cancel the ongoing request and re issue again it is throwing below error. (using  ajax,abort)
    **Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'This task has already been stopped'** 
  
2. When i close the app and re open it is giving the blank page

It would be great if you can help me what i am missing here.

adamdport:
@NiklasMerz I tested this PR out to see if it would resolve my cookie issues. In the network tab, you can click a request. In the headers tab of that request, there are several sections: Summary, Request, Response, and Request Data. When using this branch, the Request Data section seems to be missing, and my endpoint is not receiving the data I've tried to attach to the request. If it helps, I'm using Angular 10.1.3's HttpClient to generate the request.

If you have any ideas or need more information from me please let me know. Thanks!

NiklasMerz:
From my experience request data is always there for post requests done with fetch. Any ideas what is different?

adamdport:
It looks like the issue is only present in iOS 12 or iOS11 (I hadn't realized I hadn't updated yet). I don't see it in iOS 13 or iOS 14. Not sure if you intend to support <13, but it looks like cordova-ios@6.0.0 supports 11 or later.

A couple other things that should be considered:

- Looks like it uses a default request timeout of 60s. I was able to get past this by setting my own `setTimeoutInterval` on the request, but this probably needs to be either disabled or made configurable from cordova
- Cookies still didn't appear in the network tab, but they did appear to be sent. Not sure if this is a safari issue (or "feature")
- When I first build, I get an error `Value for SWIFT_VERSION cannot be empty`. I can set it manually in XCode and rebuild without issue, but I haven't had to do that before (with cordova-ios@6.1.1)

Thanks!

adamdport:
I found one more thing: it doesn't look like this works for iframes. An iframe that's hosted by the same remote that I'm authenticated with does not seem to have access to authentication cookies.

Thanks again.

NiklasMerz:
New PR #1030 is the new path forward. cordova-ios will just allow plugins to plug into the WKURLSchemeHandler. The proxy is done in a plugin.

girish2711:
@NiklasMerz  Thanks for the WKURLSchemeHandler implementation. I took #1030 PR and integrated with your plugin https://github.com/GEDYSIntraWare/cordova-plugin-webview-proxy. Also changed the URL per your suggestion. All my XHR calls are now directed to webiewproxy.m and completes with proper response.

However, in my case my API responds with a few session cookies and I don't see them getting synced to my Cordova main web view. I see cookies in native cookie store but not in Cordova main web view. Is this expected ? Any solution to this?
 
I have a Single Sign On implementation which needs these session cookies to be sent along with http GET URL request (window.location = url to a subdomain) to a my own server. But the request is sent without cookies and hence it fails. 

I tried by configuring CDVWebViewEngine's webdatastore to a nonPersistant one and added cookies to the same store as in below, but that didn't sync cookie to Cordova Main webview

first added app bound domain in the info.plist

and then in CDWebViewEngine:77. 

    
if (@available(iOS 14.0, *)) {
         
     configuration.limitsNavigationsToAppBoundDomains = YES;    
     WKWebsiteDataStore *dStore =[[CDVWebViewProcessPoolFactory sharedFactory] sharedDataStore]; // by adding new method in sharedFactory

        configuration.websiteDataStore = dStore; 
    } else {
        // Fallback on earlier versions
    }


and in webviewproxy.m:40

    WKWebsiteDataStore* dataStore = [[CDVWebViewProcessPoolFactory sharedFactory] sharedDataStore];
        WKHTTPCookieStore* cookieStore = dataStore.httpCookieStore;


Any idea if cookies are not meant for syncing or am I missing something ? I am testing in iOS14 with XCode 12.3.

PR #1006, opened 2020-10-10
Hide splash sceen without `SplashScreenDelay` on .hide()

### LINKED ISSUE #991: SplashScreenDelay should not apply when calling splashscreen.show() / splashscreen.hide()
# Bug Report

## Problem

Calling `splashscreen.show()` or `splashscreen.hide()` do not function correctly when `SplashScreenDelay` is set to a value.

Manually calling `splashscreen.show()` or `splashscreen.hide()` should trigger the splash screen to show/hide immediately. The value of `SplashScreenDelay` is only meant to be taken into account when the app first loads and `AutoHideSplashScreen` is `true` - i.e. it's for controlling the initial delay when the app first boots.

(at least, this is how the documentation reads, and how it used to work on previous versions of the splash screen plugin)

One should be able to manually show/hide the splash screen immediately as required.

Calling `splashscreen.show()` or `splashscreen.hide()`, there is a delay before the action actually happens - the delay being the value of `SplashScreenDelay`.

iOS any version

### Version information
Cordova iOS 6.1.1

## Checklist

BendingBender:
Fixes #991

### Platforms affected
iOS

### Motivation and Context
The `cordova-plugin-splash-screen` allowed to hide the splash screen as a race between the timeout specified in the `SplashScreenDelay` preference and the call to `navigator.splashscreen.hide()` in JS code. This was very desirable, it allowed to show the splash screen for a long time if the application took a long time to boot but not too long so that users wouldn't think that the application is hanging.

Since the reimplementation of this functionality here, the behavior has changed. Now, the `SplashScreenDelay` preference applies to calls made by `navigator.splashscreen.hide()`, defeating the behavior described above.

### Description
Make hiding of splash screen again a race between the timeout specified in `SplashScreenDelay` setting and `navigator.splashscreen.hide()` call.

### Testing
Manual tests in simulator and real device (iPhone X).

### Checklist

- [ ] ~I added automated test coverage as appropriate for this change~ - no idea, how
- [ ] ~Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)~ - doesn't make sense, this is a cordova-ios repo, it can only be ios-related

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -767,7 +767,18 @@
     self.webView.hidden = NO;
 
     if ([self.settings cordovaBoolSettingForKey:@"AutoHideSplashScreen" defaultValue:YES]) {
-       [self showLaunchScreen:NO];
+        CGFloat splashScreenDelaySetting = [self.settings cordovaFloatSettingForKey:@"SplashScreenDelay" defaultValue:0];
+
+        if (splashScreenDelaySetting == 0) {
+            [self showLaunchScreen:NO];
+        } else {
+            // Divide by 1000 because config returns milliseconds and NSTimer takes seconds
+            CGFloat splashScreenDelay = splashScreenDelaySetting / 1000;
+
+            [NSTimer scheduledTimerWithTimeInterval:splashScreenDelay repeats:NO block:^(NSTimer * _Nonnull timer) {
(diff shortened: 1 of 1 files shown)


dpogue:
Thanks for the PR @BendingBender! 🎉 

This looks good with a quick skim over it, I'll take a more thorough look tomorrow.

dpogue (approved):

Issue #897, opened 2020-06-11
origin is null for requests sent with window.fetch on WKWebView engine


StaymanHou:
### Issue Type
- [ ] Bug Report
- [ ] Feature Request
- [x] Support Question

## Description
I'm attempting to upgrade cordova-ios from 5.1.1 to 6.0.0 for a project. At the same time, I'm also trying to migrate from uiwebview to wkwebview since that's the new default. And uiwebview is going to be deprecated anyways.

Following the release note, I've managed to get the standard xhr working by setting the `scheme` and `hostname` preferences in config.xml

However the requests sent with window.fetch still have the origin set to null despite the custom scheme configuration.

Is that the expected behavior, and I will have to migrate to some other plugin to replace the window.fetch functionality? Is there something else that I'm missing?

breautek:
> Following the release note, I've managed to get the standard xhr working by setting the scheme and hostname preferences in config.xml

It's my understanding that this should have done the trick. Can you maybe test the `XMLHttpRequest` method instead of `fetch`?

breautek:
This sounds like a wkwebview issue if it works with standard `XMLHttpRequest` but not with `fetch`.

breautek:
My wild guess is wkwebview is treating the request as a tainted origin, which I believe only does this on the `fetch` api. If the origin is considered tainted, then the origin becomes null, according to the Fetch spec.

https://fetch.spec.whatwg.org/#serializing-a-request-origin

StaymanHou:
Managed to workaround that by using https://github.com/aporat/cordova-plugin-fetch

Still would love to figure out the cause when get a chance, but the fetch plugin by @aporat would do it for now.

75341704:
Same problem when the axios is used.

jony89:
same problem here with axios as well.

```Failed to load resource: Origin null is not allowed by Access-Control-Allow-Origin.```

sams4465:
in `CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m`
after `WKWebViewConfiguration* configuration = [[WKWebViewConfiguration alloc] init];`
add `[configuration setValue:@"TRUE" forKey:@"allowUniversalAccessFromFileURLs"];`
can fix api fail

gaollard:
Same problem when the axios is used.

adipascu:
Here are some links that may be relevant for this issue:
[...]

PR #1012, opened 2020-10-27
(iOS)fix origin is null when I want to use axios to get an JSON from …

sams4465:
### Platforms affected
iOS

### Motivation and Context
#897 

### Description
add allowUniversalAccessFromFileURLs setting

allowUniversalAccessFromFileURLs
Boolean that sets whether JavaScript running in the context of a file scheme URL should be allowed to access content from any origin. Including accessing content from other file scheme URLs. The default value is false.

### Checklist


dpogue:
Thanks for the pull request.

Unfortunately, `allowUniversalAccessFromFileURLs` is considered private API by Apple, and has been [actively discouraged](https://bugs.webkit.org/show_bug.cgi?id=154916#c8) by WebKit developers due to potential for breaking in future iOS versions. Cordova will not adopt this because it introduces too much risk to apps built with Cordova.

PR #1014, opened 2020-10-28
Don't show splashscreen when using standalone CDVViewController

### LINKED ISSUE #929: Cannot disable splashscreen in standalone CDVViewController
# Bug Report

## Problem

According to the Cordova Splashscreen documentation, iOS apps should be able to disable the Cordova-provided launch splashscreen by setting SplashScreenDelay to 0 and FadeSplashScreenDuration to 0. As of 6.1.0, these settings no longer have that effect. `createLaunchView` in `CDVViewController.m` does not inspect these settings before instantiating the launch storyboard.

Reference:
https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-splashscreen/index.html#launch-storyboard-images

https://github.com/apache/cordova-ios/compare/5.1.x...6.1.x#diff-3d962b8f101cacedea7abce2ca246eac

## Checklist

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes #929 

### Description

When using CDVViewController via CordovaLib as a library, the app might already be loaded and not need to show the launch storyboard again.

Rather than unconditionally showing the splashscreen until the webview content has loaded, make it initially invisible. Then in the project template, make it visible so that it continues working as expected for apps using the Cordova CLI.

### Testing

Manual testing

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.h
+++ b/CordovaLib/Classes/Public/CDVViewController.h
@@ -37,6 +37,7 @@
 }
 
 @property (nonatomic, readonly, weak) IBOutlet UIView* webView;
+@property (nonatomic, readonly, strong) UIView* launchView;
 
 @property (nonatomic, readonly, strong) NSMutableDictionary* pluginObjects;
 @property (nonatomic, readonly, strong) NSDictionary* pluginsMap;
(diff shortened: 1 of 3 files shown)


NiklasMerz (commented):
I am still an ObjectiveC noob but trying to learn stuff. Why not use the hidden property here?

https://developer.apple.com/documentation/uikit/uiview/1622585-hidden

erisu (approved):
LGTM

PR #1015, opened 2020-10-29
(iOS) Project changes to allow consuming cordova as submodule

wmathurin:
### Platforms affected
iOS

### Motivation and Context
If your application consumes cordova as a submodule and depends on the cordova dynamic framework, it will fail to run properly because the web view engine classes are not in the framework.
As a result, CDVViewController fails to build the web view engine by reflection.
```obj-c
self.webViewEngine = [[NSClassFromString(defaultWebViewEngineClass) alloc] initWithFrame:bounds];
```
And the view controller is never unhidden.

### Description
The change simply adds the cordova framework as a target for the web view engine classes.

### Testing
In https://github.com/forcedotcom/salesforceMobileSDK-ios-hybrid, we consume cordova through a submodule.
We have several sample apps as well as suites of tests. The change allowed the apps to run properly and the tests to pass.

### Checklist


dpogue (changes requested):
I don't think we want to expose these as public API, because the CDVWebView classes are meant to be part of an internal plugin.

The better option is probably to add a fallback case to [CDVViewController](https://github.com/apache/cordova-ios/blob/master/CordovaLib/Classes/Public/CDVViewController.m#L500-L509) that explicitly uses CDVWebViewEngine if all else fails.
It is not totally clear to me why the `CordovaDefaultWebViewEngine` preference exists when `CordovaWebViewEngine` already allows overriding.

wmathurin:
The PR is not making them public but make them part of the framework target.

The fallback is here: https://github.com/apache/cordova-ios/blob/master/CordovaLib/Classes/Public/CDVViewController.m#L508

Because the class is not in the framework, it fails to find the class at runtime.

When you consume cordova through cocoapod, the class is available because we have
```
cordova.source_files = 'CordovaLib/Classes/**/*.{h,m}', 'CordovaLib/Cordova/Cordova.h'
```
in https://github.com/apache/cordova-ios/blob/master/Cordova.podspec#L41.

dpogue (approved):
Thanks for clarifying, and thanks for the pull request

Issue #900, opened 2020-06-16
CDVWebViewEngine needs a way to set the websiteDataStore of its configuration


msmtamburro:
# Feature Request

## Motivation Behind Feature

A websiteDataStore can both be configured (e.g., set to non-persistent) and re-used (e.g., in other web browser components).  
## Feature Description

Optionally exposing the configuration as typically done with the "Cordova settings" model may not suffice.  For example, if you only added a Cordova setting for persistent vs. non-persistent, this would still not give us a handle to the websiteDataStore for re-use elsewhere.  More specifically, allowing Cordova to use an existing websiteDataStore that was created and used prior to Cordova reaching its `createConfigurationFromSettings` method is important in my case.  You could accomplish this by providing a protocol that allows us to return a websiteDataStore.

For example, add the following lines to CDVWebViewEngine.h:

```
@protocol CDVWebViewEngineConfigurationDelegate <NSObject>
/// Provides a fully configured WKWebsiteDataStore; useful for customizing configuration
- (nonnull WKWebViewConfiguration *)configuration;
@end
```

and then expose this property:

```
@property (nullable, nonatomic, weak) CDVPlugin<CDVWebViewEngineConfigurationDelegate>* configurationDelegate;
```

And finally, in CDVWebViewEngine.m, inside of `createConfigurationFromSettings:` start with:

```
    WKWebViewConfiguration* configuration = [_configurationDelegate configuration];
    if (configuration) {
        return configuration;
    }
```

## Alternatives or Workarounds

When using the CDVWKWebViewEngine (the old plugin), it was possible to override createConfigurationFromSettings to customize portions of the configuration that were not exposed through Cordova settings.  With the migration to CDVWebViewEngine, its private status prevents this type of customization without resorting to swizzling.

msmtamburro:
Apologies: the code for this fix ended up in the (merged) PR to address 1157. I verified that everything looks good locally in cordova-ios@7.0.1.  (I can successfully utilize the CDVWebViewEngineConfigurationDelegate to update the configuration of the CDVWebViewEngine.)

PR #1030, opened 2020-11-25
Allow plugins to hook into the WKURLSchemeHandler

NiklasMerz:
### Platforms affected
iOS

### Motivation and Context

This offers plugin authors the possibility to plug into the WKURLSchemeHandler. Some options where discussed on the [list](https://lists.apache.org/thread.html/r2df0ee4db5746960d41935833a7bd50fa750ad7c30161b76eb5bf862%40%3Cdev.cordova.apache.org%3E)

I did this "hook" to implement a plugin which acts as a proxy to bypass CORS and cookie issues: https://github.com/GEDYSIntraWare/cordova-plugin-webview-proxy.

Supersedes #1004 

### Description

The implementation was based on the discussions on the list and similarly implemented like `shouldOverrideLoadWithRequest`. Thanks to @erisu for the hint in the right direction.

**TODO**

- [ ] Testing with more than one plugin
- [ ] Automated testing possible?
- [ ] Documentation in cordova-docs?

### Testing

* install ios platform from this PR
* install plugin https://github.com/GEDYSIntraWare/cordova-plugin-webview-proxy
* convert http/https URL to proxy URL with `window.WebviewProxy.convertProxyUrl(url)`
* check requests, console, cookie/CORS issues

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVURLSchemeHandler.m
+++ b/CordovaLib/Classes/Public/CDVURLSchemeHandler.m
@@ -52,7 +52,7 @@
 
     for (NSString* pluginName in vc.pluginObjects) {
         self.schemePlugin = [vc.pluginObjects objectForKey:pluginName];
-        SEL selector = NSSelectorFromString(@"handleSchemeURL:");
+        SEL selector = NSSelectorFromString(@"overrideSchemeTask:");
         if ([self.schemePlugin respondsToSelector:selector]) {
             handledRequest = (((BOOL (*)(id, SEL, id <WKURLSchemeTask>))objc_msgSend)(self.schemePlugin, selector, urlSchemeTask));
             if (handledRequest) {
--- a/CordovaLib/Classes/Public/CDVURLSchemeHandler.m
+++ b/CordovaLib/Classes/Public/CDVURLSchemeHandler.m
@@ -21,6 +21,8 @@
 #import "CDVURLSchemeHandler.h"
 #import <MobileCoreServices/MobileCoreServices.h>
 
+#import <objc/message.h>
+
 @implementation CDVURLSchemeHandler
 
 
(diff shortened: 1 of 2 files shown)


erisu (approved):

dpogue (approved):

jcesarmobile (approved):

Ichhakhemka:
Lll

jcesarmobile:
Just noticed it was added to 6.1.2 milestone. I think this is a feature, so should be added as 6.2.0

dpogue:
> Just noticed it was added to 6.1.2 milestone. I think this is a feature, so should be added as 6.2.0

:+1:
I've updated the 6.1.2 milestone to 6.next to better reflect "the next release in the 6.x version, before 7.0.0". The details of patch vs minor can be figured out when we're ready to tag (I agree this should probably be a minor bump).

CordovaLib/Classes/Public/CDVURLSchemeHandler.m
```diff
@@ -39,45 +41,70 @@ - (void)webView:(WKWebView *)webView startURLSchemeTask:(id <WKURLSchemeTask>)ur
...
+     */
+    BOOL anyPluginsResponded = NO;
+    BOOL handledRequest = NO;
+
+    for (NSString* pluginName in vc.pluginObjects) {
+        CDVPlugin* plugin = [vc.pluginObjects objectForKey:pluginName];
+        SEL selector = NSSelectorFromString(@"handleSchemeURL:");
+        if ([plugin respondsToSelector:selector]) {
+            handledRequest = (((BOOL (*)(id, SEL, id <WKURLSchemeTask>))objc_msgSend)(plugin, selector, urlSchemeTask));
+            NSLog(@"Handled: %d", handledRequest);
```
msmtamburro: Hi! Mind not adding NSLogs to code that can run in production environments?

CordovaLib/Classes/Public/CDVURLSchemeHandler.m
```diff
@@ -39,45 +41,70 @@ - (void)webView:(WKWebView *)webView startURLSchemeTask:(id <WKURLSchemeTask>)ur
...
+    
+    CDVViewController* vc = (CDVViewController*)self.viewController;
+
+    /*
+     * Give plugins the chance to handle the url
+     */
+    BOOL anyPluginsResponded = NO;
+    BOOL handledRequest = NO;
+
+    for (NSString* pluginName in vc.pluginObjects) {
```
msmtamburro, line 53: This for loop pattern, which we also see in CDVWebViewEngine.m, occasionally causes crashes in our production apps because `vc.pluginObjects` can be mutated on a different thread while you're looping.  Here's an example:

```
0   CoreFoundation                	0x1ac96a5ac __exceptionPreprocess + 220 (NSException.m:199)
1   libobjc.A.dylib               	0x1c0a5842c objc_exception_throw + 60 (objc-exception.mm:565)
2   CoreFoundation                	0x1ac969f04 __NSFastEnumerationMutationHandler + 124 (NSEnumerator.m:129)
3   Our App                	0x1022e13ec -[CDVWebViewEngine webView:decidePolicyForNavigationAction:decisionHandler:] + 336 (CDVWebViewEngine.m:543)
```

There are several right ways to deal with this:

1. The best would be to use a protocol, like my similar request here: https://github.com/apache/cordova-ios/issues/900
2. You could alternatively use DispatchQueue to safely read and write pluginObjects on separate threads
3. You could simply make a copy of pluginObjects and enumerate the copy instead

Since #3 is the easiest to write in a PR comment, that would look like the below, but I'd be happy to help you do #1, which would be a lot better.

```
    /*
     * Give plugins the chance to handle the url
     */
    __block BOOL anyPluginsResponded = NO;
    __block BOOL shouldAllowRequest = NO;
    
    NSDictionary *pluginObjects = [[vc pluginObjects] copy];
    SEL selector = NSSelectorFromString(@"shouldOverrideLoadWithRequest:navigationType:");
    [pluginObjects enumerateKeysAndObjectsUsingBlock:^(id  _Nonnull key, id  _Nonnull plugin, BOOL * _Nonnull stop) {
        if ([plugin respondsToSelector:selector]) {
            anyPluginsResponded = YES;
            // https://issues.apache.org/jira/browse/CB-12497
            int navType = (int)navigationAction.navigationType;
            shouldAllowRequest = (((BOOL (*)(id, SEL, id, int))objc_msgSend)(plugin, selector, navigationAction.request, navType));
            if (!shouldAllowRequest) {
                *stop = YES;
            }
        }
    }];

```

msmtamburro (changes requested):

CordovaLib/Classes/Public/CDVURLSchemeHandler.m
msmtamburro: I updated https://github.com/apache/cordova-ios/issues/900 with an example of a protocol that you could use here.

CordovaLib/Classes/Public/CDVURLSchemeHandler.m
NiklasMerz: This has been removed on a later commit

msmtamburro: Ah, thanks for that-- (sorry) I googled in to the wrong place to begin my review of your PR.  Happy to discuss my other comment (which still applies), if clarification is required.

NiklasMerz: Yeah the other comment ist still valid. We would need to fix more places though.

Would be nice if you could provide a pull request for the existing places. I am not am ObjC expert.

adamdport:
Migrating some comments from #1004 :

- This has a default request timeout of 60s. I was able to get past this by setting my own `setTimeoutInterval` on the request, but this probably needs to be either disabled or made configurable from cordova
- Cookies don't appear in the network tab, but they did appear to be sent
- Cookies aren't sent to iframes that load from the same domain
- Cookies aren't sent to InAppBrowser windows that load from the same domain

Thanks again.

NiklasMerz:
@adamdport Thanks for you tests and comments. Most of your comments apply to the "proxy" I built in #1004. This PR is a generic way for plugins to intercept web requests. I built the first plugin to use this API from the proxy code of the previous PR here: https://github.com/GEDYSIntraWare/cordova-plugin-webview-proxy

So this PR has nothing to do with CORS, cookies etc. I just happen to use for these things by building a plugin that depends on this PR.

adamdport:
@NiklasMerz my mistake. I'll open issues on that repo instead, thanks

CordovaLib/Classes/Public/CDVURLSchemeHandler.m
msmtamburro: Will do.  I am in the process of obtaining permission from my employer.

CordovaLib/Classes/Public/CDVURLSchemeHandler.m
NiklasMerz: Thanks for your feedback on this. As I am not an ObjectiveC expert I am not sure how to deal with this. I would probably use "#3" for this PR and add a commen to #900 to properly implement the protocol for this and the other affected loop.

girish2711:
Bringing comments added for #1004 


@NiklasMerz Thanks for the WKURLSchemeHandler implementation. I took #1030 PR and integrated with your plugin https://github.com/GEDYSIntraWare/cordova-plugin-webview-proxy. Also changed the URL per your suggestion. All my XHR calls are now directed to webiewproxy.m and completes with proper response.

However, in my case my API responds with a few session cookies and I don't see them getting synced to my Cordova main web view. I see cookies in native cookie store but not in Cordova main web view. Is this expected ? Any solution to this?

I have a Single Sign On implementation which needs these session cookies to be sent along with http GET URL request (window.location = url to a subdomain) to a my own server. But the request is sent without cookies and hence it fails.

I tried by configuring CDVWebViewEngine's webdatastore to a nonPersistant one and added cookies to the same store as in below, but that didn't sync cookie to Cordova Main webview

first added app bound domain in the info.plist

and then in CDWebViewEngine:77.

if (@available(iOS 14.0, *)) {

 configuration.limitsNavigationsToAppBoundDomains = YES;    
 WKWebsiteDataStore *dStore =[[CDVWebViewProcessPoolFactory sharedFactory] sharedDataStore]; // by adding new method in sharedFactory

    configuration.websiteDataStore = dStore; 
} else {
    // Fallback on earlier versions
}
and in webviewproxy.m:40

WKWebsiteDataStore* dataStore = [[CDVWebViewProcessPoolFactory sharedFactory] sharedDataStore];
    WKHTTPCookieStore* cookieStore = dataStore.httpCookieStore;
Any idea if cookies are not meant for syncing or am I missing something ? I am testing in iOS14 with XCode 12.3.

NiklasMerz:
@girish2711 Thanks for you feedback. Seems like you hit some edge cases with the cookie quirks in WebKit. We should move this discussion to the proxy plugin and not continue in this PR.

This PR just adds the ability for plugins to hook into the code in the platform. Again please don't discuss issues around the "proxy" here.

NiklasMerz:
If no one objects I will merge this in 24 hours.

msmtamburro:
Just a note: I was waiting for #1050 to be merged so that I could merge that in to #1051.  #1051 has the example you could follow in this PR (but it might be hard to read until #1050 is merged).

totalpave (approved):

breautek (approved):

msmtamburro:
> Just a note: I was waiting for #1050 to be merged so that I could merge that in to #1051. #1051 has the example you could follow in this PR (but it might be hard to read until #1050 is merged).

This newly merged PR copied the crashing pattern that I've addressed in #1050 and #1051.  You might want to open a new issue to follow the new pattern once those two PRs get merged.

NiklasMerz:
@msmtamburro Yes we should address that in the next version with #1050, #1051 and changes to this feature together. Feel free to open an issue.

mattdsteele:
@NiklasMerz Thank you (and everyone else) for putting this together! I'm excited to put this to use.

It's not clear to me how this could be integrated into a Cordova app, for folks looking to solve the original issue of bypassing CORS errors in their app.

Is updating to `cordova-ios@6.2.0` and loading https://github.com/GEDYSIntraWare/cordova-plugin-webview-proxy (following the "Testing" guide above) the near-term recommendation? Or is there another plugin folks should be tracking on?

NiklasMerz:
@mattdsteele As far as I know my plugin https://github.com/GEDYSIntraWare/cordova-plugin-webview-proxy is only one to use this feature so far. I built it to solve my CORS and cookie issues. Maybe others will use this hook to implementen different solutions.

Give the plugin a try and let me know in the plugins repo if it works for you or issues you find.

tudordumitriu:
@NiklasMerz  I have tried using the cordova ios 6.2.0 but unfortunately still getting the "error: Build input file cannot be found: '../mobile/platforms/ios/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewProcessPoolFactory.m' (in target 'CordovaLib' from project 'CordovaLib')" and if you check the code https://github.com/apache/cordova-ios/tree/master/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine the folder indeed doesn't seem to be containing CDVWebViewProcessPoolFactory.m
I did add the platform via cordova platform add ios@6.2.0

NiklasMerz:
@tudordumitriu This is not the right PR for this. Please open a new issue or contine in the commit.

tudordumitriu:
@NiklasMerz 
Sorry, will post it in the proper thread

PR #1031, opened 2020-11-25
Set CDVWebViewProcessPoolFactory to public

NiklasMerz:
### Platforms affected
iOS

### Motivation and Context

The InAppBrowser used to import the process pool from the wkwebview plugin. Now the processpool is in the iOS platform and got renamed. Because it's in the private part InAppBrowser can never import it and use the shared pool.

See this issue: https://github.com/apache/cordova-plugin-inappbrowser/issues/751 and PR https://github.com/apache/cordova-plugin-inappbrowser/pull/825

### Description

I just set the file to public in Xcode.

From my understanding this should be public for plugins to use (like IAB to share Cookies). I am not sure if this is the right way to make this public though. Are there any other problems for having this private?

### Checklist


dpogue:
The problem is that this is not part of our public API, it's part of a built-in internal plugin. It lives in `Classes/Private/Plugins` rather than `Classes/Public`.

CDVWebViewProcessPoolFactory is fairly small and self-contained though, so maybe it makes sense to move that one class into the public API?

NiklasMerz:
By setting the checkbox it becomes public and works. But I would agree that moving it to `Classes/Public` would make sense since it should be part of the public API.

NiklasMerz:
I did another commit and moved the class to the `Classes/Public` folder. From my understanding this could be public like it used to be in WKWebView engine plugin. Now the InAppBrowser can use the pool again.

Does anyone know reasons why this is private with the other plugins? Any known issues in the past?

erisu (approved):

timbru31 (approved):
LGTM

NiklasMerz:
Thanks for the reviews. If no one objects I will merge this in 24 hours together with https://github.com/apache/cordova-plugin-inappbrowser/pull/825.

dpogue (approved):

PR #1050, opened 2021-01-05
feat: add CDVWebViewEngineConfigurationDelegate

msmtamburro:
to fully expose WKWebView configuration (#900)

### Platforms affected

cordova-ios

### Motivation and Context

https://github.com/apache/cordova-ios/issues/900

### Description

Allows the consumer to extend CDVViewController as a CDVWebViewEngineConfigurationDelegate and provide a WKWebViewConfiguration.  This is useful for more complex configuration cases (i.e., ones where configuration through config.xml is insufficient).

### Testing

I tested both the existing behavior:
- letting cordova-ios create the WKWebViewConfiguration on your behalf

And the new behavior:
- optionally providing your own configuration, which is used instead 

### Checklist


msmtamburro:
Great, understanding this existing dependency is helpful.  I can bring back the existing init alongside this new one.

CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -505,9 +505,10 @@ - (UIView*)newCordovaViewWithFrame:(CGRect)bounds
             self.webViewEngine = [[NSClassFromString(defaultWebViewEngineClass) alloc] initWithFrame:bounds];
         }
     } else {
-        self.webViewEngine = [[NSClassFromString(defaultWebViewEngineClass) alloc] initWithFrame:bounds];
+        WKWebViewConfiguration *config = [self respondsToSelector:@selector(configuration)] ? [self configuration] : nil;
+        self.webViewEngine = [[NSClassFromString(defaultWebViewEngineClass) alloc] initWithFrame:bounds configuration:config];
```
dpogue, line 509: It looks like you're only applying this in the case when an engine is not specified.

What might be best is to make a private `initWebViewEngine:(NSClass)engineClass` method that takes in the engine class, and checks if it responds to `initWithFrame:configuration:` and calls that, falling back to `initWithFrame:`.

Then we'd call `initWebViewEngine:` on lines 502, 505, and 509 here

dpogue (commented):
Does the addition of `NS_ASSUME_NONNULL_BEGIN` and `nullable` impact the public API for plugins that implement subclasses of CDVViewController and CDVWebViewEngineProtocol?

We need to be careful with public API to ensure (as much as possible) that we don't break existing plugins, otherwise people will just stay on older versions and keep opening issues requesting fixes be backported.

msmtamburro:
> Does the addition of `NS_ASSUME_NONNULL_BEGIN` and `nullable` impact the public API for plugins that implement subclasses of CDVViewController and CDVWebViewEngineProtocol?
> 
> We need to be careful with public API to ensure (as much as possible) that we don't break existing plugins, otherwise people will just stay on older versions and keep opening issues requesting fixes be backported.

Short answer, no impact. 

Longer answer: Before the changes, the original code could actually return nil, but the consumer had no way to understand this without nullability specifiers.  The consumers likely assumed that these methods never return nil.  Adding the `NS_ASSUME_NONNULL_BEGIN` is the equivalent of making no change to the nullability specifiers, however incorrect they may be.   Adding `nullable` to the new method tells the consumer to expect that this method can return null.  One could go through and update all of the previous methods with appropriate nullability specifiers at some point in the future.

dpogue:
> Adding `nullable` to the new method tells the consumer to expect that this method can return null. One could go through and update all of the previous methods with appropriate nullability specifiers at some point in the future.

You've changed `initWithFrame:` to have `nullable` too:
```objectivec
- (nullable instancetype)initWithFrame:(CGRect)frame;
```

Will existing plugins that implement it without `nullable` run into compiler errors?
i.e., https://github.com/ionic-team/cordova-plugin-ionic-webview/blob/master/src/ios/CDVWKWebViewEngine.m#L117

CordovaLib/Classes/Public/CDVViewController.m
msmtamburro: This is a great suggestion... I didn't imagine custom web engines using this, but of course, that makes sense.  I'll tidy this up.

msmtamburro: Since the change was more extensive than expected, I raised it in another branch on my fork: https://github.com/msmtamburro/cordova-ios/pull/1/files

If it makes sense, I'll merge it in to this one.

msmtamburro:
> > Adding `nullable` to the new method tells the consumer to expect that this method can return null. One could go through and update all of the previous methods with appropriate nullability specifiers at some point in the future.
> 
> You've changed `initWithFrame:` to have `nullable` too:
> 
> ```objc
> - (nullable instancetype)initWithFrame:(CGRect)frame;
> ```
> 
> Will existing plugins that implement it without `nullable` run into compiler errors?
> i.e., https://github.com/ionic-team/cordova-plugin-ionic-webview/blob/master/src/ios/CDVWKWebViewEngine.m#L117

They may possibly see a warning (depending on project setup, and language used), but it is a good warning regarding unchanged code.  The existing code could have always returned nil here.  It should not stop anyone from compiling, even if they are treating warnings as errors in their own project.

dpogue (approved):
This looks good to me.

@NiklasMerz @jcesarmobile Any thoughts/concerns here?

NiklasMerz:
No concerns. This looks like a really good change for plugin authors. I am just not confident enough in ObjectiveC to review the code/find issues.

msmtamburro:
> > > Adding `nullable` to the new method tells the consumer to expect that this method can return null. One could go through and update all of the previous methods with appropriate nullability specifiers at some point in the future.
> > 
> > 
> > You've changed `initWithFrame:` to have `nullable` too:
> > ```objc
> > - (nullable instancetype)initWithFrame:(CGRect)frame;
> > ```
> > 
> > 
> > Will existing plugins that implement it without `nullable` run into compiler errors?
> > i.e., https://github.com/ionic-team/cordova-plugin-ionic-webview/blob/master/src/ios/CDVWKWebViewEngine.m#L117
> 
> They may possibly see a warning (depending on project setup, and language used), but it is a good warning regarding unchanged code. The existing code could have always returned nil here. It should not stop anyone from compiling, even if they are treating warnings as errors in their own project.

I took a moment to go through existing (unchanged) methods that fall under the assumed "nonnull" but can actually return nil, and explicitly marked them as "nullable," so there's less ambiguity.

erisu (approved):

PR #1105, opened 2021-05-06
(iOS) add undeclared dependency `underscore`

merceyz:
### Platforms affected

iOS

### Motivation and Context

`cordova-ios` depends on `underscore` but doesn't declare it as a dependency
https://github.com/apache/cordova-ios/blob/048c1cace886d950860236adb893a860dfbd3503/bin/templates/scripts/cordova/lib/projectFile.js#L22

### Description

Added `underscore` as a dependency

### Testing

Ran `cordova prepare ios` under a strict dependency environment

### Checklist

- [ ] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [x] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [x] I've updated the documentation if necessary

PR #1116, opened 2021-06-06
refactor(projectFile): drop dependency on underscore

raphinesse:
### Motivation and Context

Drop dependency on underscore. Ref: #1105

### Description

Rewrote underscore usage in plain JS.

### Testing

Existing tests pass.

### DIFF
--- a/bin/templates/scripts/cordova/lib/projectFile.js
+++ b/bin/templates/scripts/cordova/lib/projectFile.js
@@ -19,7 +19,6 @@
 
 const xcode = require('xcode');
 const plist = require('plist');
-const _ = require('underscore');
 const path = require('path');
 const fs = require('fs-extra');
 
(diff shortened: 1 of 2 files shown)


bin/templates/scripts/cordova/lib/projectFile.js
```diff
@@ -40,7 +39,7 @@ function parseProjectFile (locations) {
     xcodeproj.parseSync();
 
     const xcBuildConfiguration = xcodeproj.pbxXCBuildConfigurationSection();
-    const plist_file_entry = _.find(xcBuildConfiguration, entry => entry.buildSettings && entry.buildSettings.INFOPLIST_FILE);
+    const plist_file_entry = Object.values(xcBuildConfiguration || {}).find(entry => entry.buildSettings && entry.buildSettings.INFOPLIST_FILE);
```
raphinesse, line 42: We could probably reduce this to the following:

```suggestion
    const plist_file_entry = Object.values(xcBuildConfiguration).find(entry => entry.buildSettings && entry.buildSettings.INFOPLIST_FILE);
```

Since in the next line, `plist_file_entry` is assumed to be truthy, which cannot be the case if `xcBuildConfiguration` is falsy.

breautek (approved):
Looks good to me.

As for your suggestion, I'm indifferent. Either approach looks like it will do the job and I don't think either is necessary better than the other.

timbru31 (approved):
Nice, thanks!

Issue #1120, opened 2021-06-23
Presenting with JS Alert()  dead lock


Luo-Kuang:
# Bug Report

### What is expected to happen?
alert normaly

### What does actually happen?
When I call JS alert in the present controller, the interface will be stuck

## Information

[Presentation] Attempt to present <UIAlertController: 0x15201c200> on <MainViewController: 0x151e09ab0> (from <MainViewController: 0x151e09ab0>) whose view is not in the window hierarchy.

### Command or Code

```js
index.html
function success(msg){
    console.log(msg)
    alert(msg)
}
function failed(){}
function pl() {
    cordova.exec(success,failed,"testPlugin","method",["msg"]);
}
```

```oc
testPlugin
- (void)method:(CDVInvokedUrlCommand *)command {
    ViewController *tempC = [[ViewController alloc] init];
    [[TestPlugin topViewController] presentViewController:tempC animated:YES completion:nil];
}
```

when I back to MainViewController, the interface will be stuck

### Environment, Platform, Device

iOS

### Version information

Apache Cordova native platform version 6.2.0 

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

spoxies:
This is indeed quite an issue as Lou-Kuang says, and I'd like to chip in with some extra context to replicate the issue (and show its urgency).

So in schort if webKit or the underlaying iOS presents a UIAlertController and whilst that is open, a `alert(..), confirm(...)` is called, the app will be stuck.

This commonly happens if a user is prompted for some native permission, or shake the device to trigger "Shake to Undo". As far as the javascript goes nothing is halted, so some event might trigger an alert

So an alternative/easy way to replicate this issue is execute:

```javascript
setTimeout(function(){
        alert('This will mess up');
    }, 5000);
```
and shake the device (e.g. like a Polaroid picture), to trigger native Undo dialog....app wil be dead

PR #1121, opened 2021-06-23
fix bug alert() iOS

Luo-Kuang:
### Platforms affected

iOS

### Motivation and Context

issues #1120

### Checklist

- [x ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [x ] I've updated the documentation if necessary

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewUIDelegate.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewUIDelegate.m
@@ -49,9 +49,8 @@
 
     [alert addAction:ok];
 
-    UIViewController* rootController = [UIApplication sharedApplication].delegate.window.rootViewController;
 
-    [rootController presentViewController:alert animated:YES completion:nil];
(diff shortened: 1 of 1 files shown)


CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewUIDelegate.m
```diff
@@ -160,4 +159,29 @@ - (void)webViewDidClose:(WKWebView*)webView
 }
 
 
++ (UIViewController *)topViewController {
```
dpogue, line 162: Is there a reason for this to be a static method on `CDVWebViewUIDelegate` rather than an instance method used with `[self topViewController]`? It looks like all uses of it are within the `CDVWebViewUIDelegate` class

dpogue (approved):
This looks okay to me. I'm not sure it needs to be a class method rather than an instance method, but the idea seems good

PR #1154, opened 2021-09-30
Swift Package Manager support for CordovaLib

dpogue:
### Platforms affected
iOS

### Motivation and Context
This adds a `Package.swift` file to allow installing the CordovaLib library through Swift Package Manager.

This also cleans up our Cocoapods spec file to remove the need to explicitly list each header.

### Description
Swift Package Manager requires all the public headers to be in an `include` folder, so all the headers from `Classes/Public` were moved to `include/Cordova`.

Also got rid of the precompiled header because SwiftPM doesn't know how to do that, and we were only precompiling Foundation and UIKit.

Xcode 13 also adds a new warning when framework headers are used with relative imports `"myfile.h"` rather than framework imports `<myfile.h>`. I've cleaned up the imports for all the public headers. (Note: some of the plugins use private headers, which must retain the relative import syntax)

### Testing
* Tested creating a new Swift project in Xcode, adding CordovaLib using SwiftPM and compiling successfully
* Tested creating a new Objective C project in Xcode, adding CordovaLib using Cocoapods as a dynamic library and compiling successfully
* Tested creating a new Objective C project in Xcode, adding CordovaLib using Cocoapods as a static library and compiling successfully
* Existing test suite passes

### Checklist


erisu (approved):
Code changes LGTM

I tested with default Cordova app.

* Create Project
* Platform Add
* Prepare
* Build
* Run
  * `Device is Ready` passed

```
Cordova Packages:

    cli: 10.0.0
        common: 4.0.2
        create: 3.0.0
        lib: 10.0.0
            common: 4.0.2
            fetch: 3.0.1
            serve: 4.0.0

Project Installed Platforms:

    ios: 6.2.0

Environment:

    OS: macOS 11.6 (20G165) (darwin 20.6.0) x64
    Node: v14.18.0
    npm: 6.14.15

ios Environment:

    xcodebuild:
Xcode 13.0
Build version 13A233

```

Issue #699, opened 2019-10-23
Catalyst for ios platform ?


jhjulkun:
Hi 
Has anyone tried building an ios project using 'Catalyst' for macOS Catalina ?

If I try enabling that build, I get:
 
CDVURLProtocol.m:20:9: AssetsLibrary is deprecated and is not available when building for Mac Catalyst. Consider migrating to Photos instead, or use `#if !TARGET_OS_MACCATALYST` to conditionally import this framework, first importing TargetConditionals.h if necessary.

Seems like the AssetsLibrary has been deprecated, and has been fixed in ionic, are you guys planning on removing it too ?
https://github.com/ionic-team/capacitor/pull/1825

Going forward, should Catalyst be used or the 'osx' platform ?

Thanks

jcesarmobile:
moved to cordova-ios repo

dpogue:
> Going forward, should Catalyst be used or the 'osx' platform ?

Once it works, I think we'd want to encourage using Catalyst for macOS to avoid needing to maintain a similar-but-separate cordova-osx platform.

See also https://github.com/apache/cordova-ios/issues/551.

We'd definitely like to address this, but it's not clear if that code is actually being used (it probably is, by someone, somewhere) and whether removing it (like Ionic did) is the right step.

jcesarmobile:
It was removed in Capacitor, was not really being used there.

Since it's in CDVURLProtocol.m and that won't work on WKWebView, I think it's safe to remove once we remove UIWebView.

brodybits:
Unfortunately the maintainers are overloaded due to lack of resources, see apache/cordova#163.

erisu:
`CDVURLProtocol` was removed in #783 and will be in next major release.

jcesarmobile:
Have you tried to run the app in Catalyst after that change?
If so, maybe we should enable it by default on the project template or add a way of easily adding it

dpogue:
Catalyst is only supported in Xcode 11 on macOS Catalina, so I'm not sure we'd want to enable it by default.

erisu:
From `master`, I believe it was decided to only support Xcode 11+ as all apps starting April 2020 requires to be built with Xcode 11.

I am not sure what we say about macOS requirements.

jcesarmobile:
What happens if you add macOS target and then you open the project on Mojave? I updated both of my computers, so I can’t test anymore.
If not by default, then maybe a command that adds catalyst instead of cordova-osx

[...]

PR #1158, opened 2021-10-08
refactor: replace copy-www-build-step script with build phase

NiklasMerz:
### Platforms affected
iOS and Mac (Catalyst)

### Motivation and Context

We could move the build step of copying the www directory and config.xml to the Xcode "Copy files" commands to avoid issues from the manual script like changed paths or different paths between iOS and macOS.

See #699

### Description

**Up for discussion.**

Some links: https://github.com/apache/cordova-ios/issues/699#issuecomment-645119408

This is probably the most important reason for the script: https://issues.apache.org/jira/browse/CB-5397?focusedCommentId=13859057&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13859057

### Testing

* Run iOS simulator from CLI
* Run iOS simulator from Xcode
* Run Mac (designed for iPad) from Xcode
* Enable and run Mac natively (M1) from Xcode 

### Checklist


NiklasMerz:
Fixed now. Good to know why they are there.

I should have checked the diff again. These Xcode files are tricky.

NiklasMerz:
@dpogue How did you test that? What's the best way to reproduce this?

I did quick tests and this www folder and config.xml look like the right one. I changed them to **__Project_NAME/..** but then plugins won't work anymore because cordovas js files are missing.

My test:

- Change index.html, build ios, run app on mac and ios sim -> see changes
- Change index.html in config.xml, build ios, run mac and ios sim from Xcode -> page cannot load

To me this looks like the right files get copied into the app.

dpogue:
hmm, I was testing this as part of https://github.com/apache/cordova-ios/pull/1162 (which used this as a starting point) and getting "Connecting to device..." in the simulator, and log messages about the Console plugin not being found.

This should work to reproduce:
```
npx cordova@nightly create xcodeTestApp
cd xcodeTestApp
npx cordova@nightly platform add github:apache/cordova-ios#xcode-copy
open platforms/ios/HelloCordova.xcworkspace
# Run the app in a simulator from Xcode
```

NiklasMerz:
Interesting side fact. I just tried to run the osx platform on the M1 mac with this PR https://github.com/apache/cordova-osx/pull/105 and got the same error:

`ERROR: Plugin 'Device' not found, or is not a CDVPlugin. Check your plugin mapping in config.xml.
`

dpogue (approved):
I've tested this (including with Catalyst) and it is working as expected. Also tested with cordova-plugin-device added, and was able to print out device information.

NiklasMerz (commented):
Thanks @erisu for completing this and your good explanation. It's working as espected for me as well with all Mac configurations and an iOS simulator.

PR #1162, opened 2021-10-19
WIP: Modern app template

dpogue:
Depends on https://github.com/apache/cordova-ios/pull/1158

### Platforms affected
iOS

### Motivation and Context
Make the app template project match a modern Xcode-created iOS app project.

### Description
* Restructures the Xcode project to more closely match what gets generated when you create a new Objective C iOS app with Xcode.
* Uses a storyboard for the main ViewController rather than a xib file.
* Renames Images.xcassets to Assets.xcassets.
* Uses CordovaLib from Swift Package Manager rather than a subproject.
* Enables Mac Catalyst support.
* Adopts SceneDelegate API on iOS 13+.

### Testing
Created an app with `npx cordova platform add ~/Coding/cordova-ios`.
Ran the app successfully on iOS 11, iOS 12, iOS 13, iOS 14, and iOS 15. Also ran the app successfully on macOS with Catalyst.

Unit tests pass.

This need **lots** of testing to make sure the SceneDelegate stuff isn't interfering and blocking stuff that used to fire on the AppDelegate. Might need to introduce a CDVSceneDelegate in CordovaLib to account for this.


dpogue:
I'm breaking this up into smaller pull requests for easier review

brunolau:
was this ever merged? or is there a simple way how to use cordova-ios 7.0.1 with Scene lifecycle? 

I'm trying to implement CarPlay and can't get my head around merging the AppDelegate and SceneDelegate

dpogue:
The template app for Cordova iOS 7.x does not use a Scene Delegate, but we're looking at adding that as part of Cordova iOS 8.0.0 (next major).

brunolau:
Thank you very much for quick reply. Is there (even a rough) ETA when 8.0.0 might be shipped so that I can report to the customer? Thanks a lot in advance

dpogue:
No ETA. Cordova is a volunteer-developed project, so we are not able to make any commitments as far as release timelines.

I'm hoping to have the time to start work on this particular bit later in August, but there are some larger things on the roadmap for 8.0 as well.

brunolau:
I've seen you've already implemented this in https://github.com/apache/cordova-ios/pull/1457 . Sorry to bother with another question, but is there a way we could use the nightly 8.0.0 for our project? (haven't been able to find the guide how to achieve this anywhere)

PR #1161, opened 2021-10-18
chore!: Breaking API cleanups

dpogue:
### Platforms affected
iOS

### Motivation and Context
Breaking API changes for v7.0.0

### Description
* Removes deprecated `colorFromColorString` API from CDVViewController
* Makes CDVCommandDelegateImpl private API since it's an implementation detail

### Testing
Unit tests pass

PR #1172, opened 2021-11-02
Simple build not working on CI

raphinesse:
This is more of a bug showcase than a pull request. This PR basically reduces the CI test to running the following commands:

    ./bin/create app
    ./app/cordova/build --emulator

The build step fails. Maybe someone can help me figure out what's going on here. Seems like an include path issue, but I'm absolutely not familiar with xcode. So any help would be appreciated.


dpogue:
That appears to be my fault, as a result of https://github.com/apache/cordova-ios/pull/1161. I'll open a PR shortly to fix it.

PR #1173, opened 2021-11-02
chore: Remove unused imports from template

dpogue:
### Platforms affected
iOS

### Motivation and Context
This work was missed in https://github.com/apache/cordova-ios/pull/1161 and is now causing errors.

### Description
Removed unused headers from the project template.

### Testing
Unit tests pass (although they also didn't catch this being an error last time)

### Checklist

### DIFF
--- a/bin/templates/project/__PROJECT_NAME__/Classes/MainViewController.h
+++ b/bin/templates/project/__PROJECT_NAME__/Classes/MainViewController.h
@@ -26,8 +26,6 @@
 //
 
 #import <Cordova/CDVViewController.h>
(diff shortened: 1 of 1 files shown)


raphinesse (commented):
I have no clue of the ongoing changes. But I guess you know what you're doing here 😁

The reason that failing builds are not flagged by our unit tests should be fixed as soon as #1171 is ready and merged.

Does the command sequence described in #1172 work for you with this change?

dpogue:
> Does the command sequence described in #1172 work for you with this change?

Just tried it, and yes it does

raphinesse (approved):

PR #1177, opened 2021-11-03
test(create): waste less time on name variants

raphinesse:
This should speed up the create tests by a few minutes while still allowing us to notice if any part of the build had issues regarding special characters.


raphinesse:
Well, that did not make the difference I expected...

erisu (approved):

PR #1180, opened 2021-11-04
refactor!: drop platform binaries

erisu:
### Motivation and Context

This PR:

* Removes all platform binaries that are not required for compatibility to Cordova CLI.
* Refactored the update process to remove excessive hops/function calls.
* Removed some unneeded comments from `version` binary.
* Updated `create.spec.js` to use JS methods for creating and building.

### Description

- [x] Drop `create` binaries
  - [x] Update `create.spec.js`
- [x] Drop `build` binaries
  - [x] Update `create.spec.js`
- [x] Drop `update` binaries
- [x] Drop `test` binaries
  - This binary has never been used since 2012. It was referencing a binary that was renamed in 2012. If we want to add mobile-spec, it should be done in a separate PR and done with GH Actions Workflows.
- [x] Drop `check_req` binaries
- [x] Drop `clean` binaries
- [x] Drop `log` binaries
- [x] Drop `run` binaries
- [x] Cleanup `version` binaries
- [x] Drop `uncrustify` binary & config
  - This binary referenced a npm module that has not been in our repo for a long time.
  - It was not used for over 2~3+ years.
  - The npm module has not been updated over 7 years. 
- [x] Drop `list-emulator-build-targets` binary
- [x] Drop `list-started-emulators` binary
- [x] Drop `start-emulator` binary

### Testing

- `npm t`
- GitHub Actions
- Cordova CLI `10.0.0 (cordova-lib@10.1.0)`
  - `cordova create`
  - `cordova platform add ../cordova-ios-7.0.0-dev.tgz`
  - `cordova platform update ../cordova-ios-7.0.0-dev.tgz`
  - `cordova prepare ios`
  - `cordova build ios`
  - `cordova info`
  - `cordova requirements`
  - `cordova platform`
  - `cordova run ios`
  - `cordova plugin add cordova-plugin-device`

### Checklist


erisu:
Binaries & scripts that needs reviewing...

- apple_ios_version
- apple_osx_version
- apple_xcode_version
- cordova_plist_to_config_xml
- list-devices
- list-emulator-build-targets
- list-emulator-images
- list-started-emulators
- start-emulator

I believe some of these could be removed and/or merged into some JS file(s) and exported.

tests/spec/create.spec.js
```diff
@@ -17,52 +17,98 @@
...
 const fs = require('fs-extra');
-const { superspawn } = require('cordova-common');
+const os = require('os');
+const path = require('path');
+const xcode = require('xcode');
+const create = require('../../bin/lib/create');
 
-const cordova_bin = path.join(spec, '../..', 'bin');
-const tmp = require('tmp').dirSync().name;
+const makeTempDir = () => path.join(os.tmpdir(), `cordova-ios-create-test-${Date.now()}`);
```
raphinesse, line 26: IIRC there are occasional issues with `os.tmpdir()` being a symlink on macOS. The form we use in lib and android is:
```js
function makeTempDir () {
    const tmpDirTemplate = path.join(os.tmpdir(), 'cordova-android-test-');
    return fs.realpathSync(fs.mkdtempSync(tmpDirTemplate));
}


```

tests/spec/create.spec.js
```diff
@@ -17,52 +17,98 @@
...
-        fs.removeSync(projectTempDir);
-    });
+    return new Api()
+        .build({ emulator: true })
+        .then(
+            () => {
+                expect(true).toBe(true);
+            },
+            err => fail(`Project Build has failed and is not expected.: ${err}`)
+        );
```
raphinesse, line 73: I'd just delete that and name the function `buildProject` or something like that. If the promise gets rejected, jasmine will fail the test. No reason to jump through hoops.

tests/spec/create.spec.js
```diff
@@ -17,52 +17,98 @@
...
+/**
+ * Runs various create and buid checks.
+ *
+ * @param {String} tmpDir
+ * @param {String} packageName
+ * @param {String} projectName
+ * @returns
+ */
+function verifyCreateAndBuild (tmpDir, packageName, projectName) {
+    verifyProjectFiles(tmpDir, projectName);
```
raphinesse, line 85: I'd include `await create.createProject(tmpDir, packageName, projectName, {}, undefined);` as the first line here and remove it in the tests below

raphinesse (commented):
Great to see this happening. I left a few comments for now. Full review pending!

tests/spec/create.spec.js
erisu: I know this was used in Android as this is where my script came from, but I removed part of it because it also created the folder which should not happen. The create script requires the folder to not exist.

I could remove `fs.mkdtempSync` and write it like this:

```javascript
const makeTempDir = () => path.join(
    fs.realpathSync(os.tmpdir()),
    `cordova-ios-create-test-${Date.now()}`
);
```

tests/spec/create.spec.js
erisu: Check now

tests/spec/create.spec.js
erisu: Removed the error method.

tests/spec/create.spec.js
raphinesse: I see. Didn't get the must not exist part. Good solution!

raphinesse (commented):
We should also remove [`hooks/`](https://github.com/erisu/cordova-ios/tree/feat/drop-binaries/hooks) if we remove uncrustify.

tests/spec/create.spec.js
```diff
@@ -17,52 +17,97 @@
...
-        () => fail('Project create has failed and is not expected.')
-    ).finally(() => {
-        // Delete Temp Project
-        fs.removeSync(projectTempDir);
-    });
+    return new Api()
+        .build({ emulator: true })
+        .then(() => {
+            expect(true).toBe(true);
+        });
```
raphinesse, line 73: How about
```suggestion
    return expectAsync(new Api().build({ emulator: true }))
        .toBeResolved();
```

raphinesse:
> Other binaries & scripts that will need reviewing... Maybe another PR?
> ...
> I believe some of these could be removed and/or merged into some JS file(s) and exported.

The following bins are used outside of `cordova-ios` and should be left alone for now:
```
apple_ios_version
./lib/src/plugman/util/default-engines.js
43:            { platform: 'ios', scriptSrc: path.join(project_dir, 'cordova', 'apple_ios_version') },

apple_osx_version
./lib/src/plugman/util/default-engines.js
45:            { platform: 'ios', scriptSrc: path.join(project_dir, 'cordova', 'apple_osx_version') },

apple_xcode_version
./lib/src/plugman/util/default-engines.js
41:            { platform: 'ios', scriptSrc: path.join(project_dir, 'cordova', 'apple_xcode_version') },

list-devices
./lib/src/cordova/targets.js
35:    const caller = { script: 'list-devices' };
37:    const cmd = path.join(projectRoot, 'platforms', platform, 'cordova', 'lib', 'list-devices');

list-emulator-images
./lib/src/cordova/targets.js
42:    const caller = { script: 'list-emulator-images' };
44:    const cmd = path.join(projectRoot, 'platforms', platform, 'cordova', 'lib', 'list-emulator-images');
```
These are not used outside of cordova-ios and could be removed in this PR:

- cordova_plist_to_config_xml
- list-emulator-build-targets
- list-started-emulators
- start-emulator

tests/spec/create.spec.js
erisu: Yes, this will work. I should read up on Jasmine's APIs. I still have not read it.

erisu:
Yes the pre-commit hook can be removed. I didn't notice it and it does not get runned.

erisu:
@raphinesse I will excluded `cordova_plist_to_config_xml` from the dropping for now as it is written in a log output.

https://github.com/apache/cordova-ios/blob/master/CordovaLib/Classes/Public/CDVViewController.m#L155

Alternative message will need to be decided and maybe we could keep the script in some folder for handy scripts but not release it with the packages.

raphinesse:
> @raphinesse I will excluded `cordova_plist_to_config_xml` from the dropping for now as it is written in a log output.
> 
> https://github.com/apache/cordova-ios/blob/master/CordovaLib/Classes/Public/CDVViewController.m#L155
> 
> Alternative message will need to be decided and maybe we could keep the script in some folder for handy scripts but not release it with the packages.

I think the script and log message can both be removed without replacement. It's a python script that is supposed to convert Projects to the Cordova 2.3 format:
https://github.com/apache/cordova-ios/blob/a6044dd4a586cac602ab4b0a87d925b1dd60fbfd/bin/cordova_plist_to_config_xml#L21

raphinesse (approved):
Great stuff! Always happy when we can delete a few kLoCs :muscle: 

Don't forget to check my last two comments, but feel free to merge as-is if you want.

erisu:
Since we already removed it from the Android platform and released it, maybe it is safe to remove it from here as well?

I wonder if anyone has any concerns, even though it was already removed in Android... If we remove it and later it is seen as an issue, we can easily add back as a patch release.

raphinesse:
I'd remove the bat files. That's my 2¢

erisu:
OK, the three last bat files were removed:

* `bin/apple_ios_version.bat`
* `bin/apple_osx_version.bat`
* `bin/apple_xcode_version.bat`

PR #1182, opened 2021-11-06
dep!(dev): bump eslint w/ corrections

erisu:
### Motivation and Context

Update Eslint Config with Fixes

### Description

- `@cordova/eslint-config@^4.0.0`
- applied auto corrections
- applied manual corrections
  - prefer-regex-literals

### Testing

- `npm t`

### Checklist

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -24,3 +24,7 @@ overrides:
 
   - files: [cordova-js-src/**/*.js]
     extends: '@cordova/eslint-config/browser'
+
+    # TODO add this to @cordova/eslint-config/browser
+    rules:
+      no-var: 0
--- a/bin/apple_ios_version
+++ b/bin/apple_ios_version
@@ -19,6 +19,6 @@
     under the License.
 */
 
-var versions = require('./templates/scripts/cordova/lib/versions.js');
+const versions = require('./templates/scripts/cordova/lib/versions.js');
 
 versions.printOrDie('apple_ios');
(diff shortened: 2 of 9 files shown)


.eslintrc.yml
```diff
@@ -24,3 +24,5 @@ overrides:
 
   - files: [cordova-js-src/**/*.js]
     extends: '@cordova/eslint-config/browser'
+    rules:
+      no-var: 0
```
erisu, line 28: This rule should be applied directly to `@cordova/eslint-config/browser` but in the mean time applied here. Will hope to reoslve this in a later patch release of  `@cordova/eslint-config`.

raphinesse (approved):
I did some commit squashing and added a TODO about the rule exception. Good to merge from my side.

raphinesse (changes requested):
Sorry, I messed up the lockfile :(
I'll fix that!

raphinesse (approved):
OK, all good now

raphinesse:
I just remembered that we only enforce ES5 for browser code that lives in `cordova-js`. We could actually migrate the `cordova-js-src` that lives in this repo to ES6 (or beyond) **if** we can guarantee that it is supported by the iOS WebView versions we support. Is that the case?

So I think we should not disable the `no-var` rule in `@cordova/eslint-config`, since a platform should be able to decide whether it needs to support ES5-only environments or not.

Nevertheless this PR is still good to go for me (maybe remove my TODO if you agree with the above).

PR #1186, opened 2021-11-07
feat(versions): print error messages, not error objects

raphinesse:
### Description

Only log error messages instead of the whole error objects in case of failure. This improves output especially for child process errors.

### Testing

Manual testing

### DIFF
--- a/bin/templates/scripts/cordova/lib/versions.js
+++ b/bin/templates/scripts/cordova/lib/versions.js
@@ -124,7 +124,7 @@ exports.printOrDie = versionName =>
             console.log(version);
         },
         err => {
-            console.error(err);
+            console.error(err.message);
             process.exit(2);
         }
     );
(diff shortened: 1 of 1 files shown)


erisu (approved):

bin/templates/scripts/cordova/lib/versions.js
```diff
@@ -49,9 +49,9 @@ exports.get_apple_xcode_version = () => {
     return spawn('xcodebuild', ['-version'])
         .then(output => {
             const versionMatch = /Xcode (.*)/.exec(output);
-
-            if (!versionMatch) return Promise.reject(output);
-
+            if (!versionMatch) {
+                throw new CordovaError('Could not determine Xcode version');
```
erisu, line 53: I am OK with the simplified message. Could the `output` contain additional information that would help users debug why it failed?

If it could, maybe only include the output with the `--verbose` flag?

But I also don't know if the flag is visible here.

Anyways, I am good with these changes as is.

raphinesse: Good Point! I included the output in the error message.

PR #1187, opened 2021-11-07
test(version): merge component tests into unit tests

raphinesse:
I noticed that the only component tests we had were for the `versions` module and very similar to the corresponding unit tests. To keep things simple, this PR re-labels the tests as unit tests.


erisu:
I believe these are not unit tests and they are also not component tests. Unit and component testing are the same things. 

> **Component testing** is defined as a software testing type, in which the testing is performed on each component separately without integrating with other components. It’s also referred to as Module Testing when it is viewed from an architecture perspective. Component Testing is also referred to as **Unit Testing**, Program Testing or Module Testing.

I might have created the *component* test but picked the wrong name.

These are supposed to be integration tests as it requires other components that are not apart of the library.

If we mocked the results and not actually call the externals for the versions, and only test what our methods does, then it is a unit test.

Anyways, since component and unit testing are the same thing, this PR is OK...

But to follow the terminology we should make sure it mocks the results, or pull them out as SI test.

erisu (approved):
PR is OK but also read the previous comment...

raphinesse:
@erisu I agree that these tests are actually integration tests, and I think it's good that they are, i.e. I would not mock the external calls.

While I think it would be great to accurately categorize all tests, the test suites of most of our repos fail to do so. And I'm afraid that even if an effort would be made to improve that, it would be hard to make sure that future contributions maintain that new order.

That being said, we could change this PR to combine the version tests that actually call binaries and the create test into a new collection of integration tests. How do you feel about that?

erisu:
I think the new collection of integration tests is a good idea but let's keep it as a separate PR and in the backlog.

I think we have been trying to push a lot of the major changes for the upcoming release (no ETA) and there is a lot of changes that have more priorities.

Some for example:

* Keeping Cordova node logic in `node_modules`
* The new `cordova-js` build system.

I agree about the
> hard to make sure that future contributions maintain that new order.

As for this PR, I already had approved the PR and am OK with the changes. We can think about better testing practices later.

raphinesse:
100% agreed. Thanks for sharing your thoughts on this :+1:

PR #1188, opened 2021-11-07
refactor: hide implementations of build & run

raphinesse:
This makes build and run modules more closed by avoiding to expose
internals via the resolved promise value. Both modules' main exports
previously returned a promise that resolved to a subrocess spawn result.
Now these promises resolve to undefined.

The public API methods that return these promises did not document a
specific promise return value, just when the promise would resolve or
reject. Thus, this change is not breaking.

### DIFF
--- a/bin/templates/scripts/cordova/lib/build.js
+++ b/bin/templates/scripts/cordova/lib/build.js
@@ -87,6 +87,7 @@ function getDefaultSimulatorTarget () {
         });
 }
 
+/** @returns {Promise<void>} */
 module.exports.run = buildOpts => {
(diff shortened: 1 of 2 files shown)


erisu (approved):
I think this should be ok... I dont know if anyone wrote wrapping tools that used it...

raphinesse:
> I dont know if anyone wrote wrapping tools that used it...

I'd argue they relied on implementation details. But I'd be totally OK with labeling this change as breaking, since we are targeting a new major anyways.

erisu:
I am OK with any label, but this should be good to merge in.

PR #1190, opened 2021-11-10
refactor: replace superspawn with execa

raphinesse:
### Motivation and Context

Progresses https://github.com/apache/cordova/issues/16 and https://github.com/apache/cordova/issues/7

### Description

Use `execa` for any subprocess spawning.

#### A few notes
`execa` has no equivalent of `superspawn`'s `printCommand` option (if set to `true`, it emits a verbose log of the command being executed). In the current state of this PR, those verbose logs are removed without a substitute.

`superspawn` promises resolve to the captured stdout, while `execa` resolves to an object with more properties. In many of the cases where we spawn subprocesses, this does not matter, since we we only care about promise fulfillment.

`superspawn` allowed to subscribe to a running process' output by means of Q's `progress` method. This basically directly translates to subscribing to `data` events on the backing streams. So that is what we used to replace these calls. However, the existing code seems to assume that the data emitted are whole lines of output only. This is by no means guaranteed, so we added `FIXME` comments to the code.

### Testing

- Existing tests pass
- It would be great if someone could do some manual tests with the platform (I own no Apple computer)

### DIFF
--- a/bin/templates/scripts/cordova/lib/Podfile.js
+++ b/bin/templates/scripts/cordova/lib/Podfile.js
@@ -21,11 +21,8 @@
 const fs = require('fs-extra');
 const path = require('path');
 const util = require('util');
-const {
-    CordovaError,
-    events,
-    superspawn: { spawn }
-} = require('cordova-common');
+const execa = require('execa');
+const { CordovaError, events } = require('cordova-common');
 
 Podfile.FILENAME = 'Podfile';
 Podfile.declarationRegexpMap = {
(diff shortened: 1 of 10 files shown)


erisu (approved):
Code changes LGTM.

I also tested:

- [x] `create`
- [x] `platform add ios`
- [x] `prepare ios`
- [x] `build ios`
- [x] `clean ios`
  Failed but not related to the `spawn`/`execa` migration.
  > note: Using new build system
  > error: Could not delete `/Users/cordova/.test/cordovaTest/platforms/ios/build` because it was not created by the build system.
  > note: Build preparation complete
  > 
  > ** CLEAN FAILED **
- [x] `run ios --list`
- [x] `run ios`

raphinesse:
Thanks so much for the test!

PR #1191, opened 2021-11-11
refactor!: cleanup create script

erisu:
### Motivation, Context & Description

This PR refactors the `create` script.

- Remove exported `updateSubprojectHelp`
- Remove exported `update_cordova_subproject`
- Renamed internal function `update_cordova_subproject` to `updateCordovaSubproject`
  - Refactored internal function `updateCordovaSubproject` to take in two static arguments:
    - `projectXcodePath` which is the path to project's xcodeproj
    - `cordovaLibXcodePath` which is the path to the CordovaLib's xcodeproj. (Note: it maybe a symbolic link when `--link` flag is added when installing platform)
  - Improved comments
  - Removed excessive supportive internal methods
    - `AbsParentPath`
    - `AbsProjectPath`
    - `relpath`
    - `updateSubprojectHelp`
- Cleanup `copyScript`
  - Begin consolidating items into the new template location for a single line `copySync`
  - Removed string replace line for files `apple-*`. No longer support the platform-centric so the files can be hardcoded for CLI.
- Rename and refactored `copyTemplateFiles` to `copyNativeTemplateFiles`
  - Move native related files to this method.
-  Reorg directory structure:
  - Moved `bin/templates/project` to `templates/project`

### Testing

- `npm t`
- standard `platform add`
- CordovaLib symlinked `platform add`

### Checklist

- [x] I've run the tests to see all new and existing tests pass

PR #1193, opened 2021-11-13
refactor: cleanup create script

raphinesse:
### Motivation and Context

This takes the changes from the first four commits of #1191 and adds a few further simplifications in the affected code areas.

I pulled this out to reduce the scope of said PR and since I already reviewed this part of the changes.

### Description

From #1191:
- Remove exported `updateSubprojectHelp`
- Remove exported `update_cordova_subproject`
- Renamed internal function `update_cordova_subproject` to `updateCordovaSubproject`
  - Refactored internal function `updateCordovaSubproject` to take in two static arguments:
    - `projectXcodePath` which is the path to project's xcodeproj
    - `cordovaLibXcodePath` which is the path to the CordovaLib's xcodeproj. (Note: it maybe a symbolic link when `--link` flag is added when installing platform)
  - Improved comments
  - Removed excessive supportive internal methods
    - `AbsParentPath`
    - `AbsProjectPath`
    - `relpath`
    - `updateSubprojectHelp`

New: 
- Cleanup `copyJsAndCordovaLib`

### Testing

`npm t`

### DIFF
--- a/bin/lib/create.js
+++ b/bin/lib/create.js
@@ -277,6 +277,3 @@ function update_cordova_subproject (argv) {
 
     utils.replaceFileContents(projectPbxprojPath, line, newLine);
 }
-
-exports.updateSubprojectHelp = updateSubprojectHelp;
-exports.update_cordova_subproject = update_cordova_subproject;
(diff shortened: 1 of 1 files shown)


raphinesse:
I would actually argue that this change is not actually breaking, since I would not consider the removed module exports public API.

erisu (approved):

PR #1197, opened 2021-11-16
build: build cordova.js during npm prepare

erisu:
### Motivation and Context

This will build `cordova.js` from source using npm's `prepare` hook.

### Description

- build `cordova.js` from source using npm's `prepare` hook
- copy prepared `cordova.js` before running Xcode tests
- delete and ignore the prebuilt `cordova.js` files we had
- do not bundle `cordova-js-src` anymore, since support for browserify builds has been discontinued for a while now

### Testing

- `npm t`
- `cordova platform add`
- `cordova build ios`
- `cordova run ios`


bin/lib/create.js
```diff
@@ -25,9 +25,10 @@ const { CordovaError, events } = require('cordova-common');
 const utils = require('./utils');
 
 function copyJsAndCordovaLib (projectPath, projectName, use_shared) {
-    fs.copySync(path.join(ROOT, 'CordovaLib', 'cordova.js'), path.join(projectPath, 'www/cordova.js'));
-    fs.copySync(path.join(ROOT, 'cordova-js-src'), path.join(projectPath, 'platform_www/cordova-js-src'));
-    fs.copySync(path.join(ROOT, 'CordovaLib', 'cordova.js'), path.join(projectPath, 'platform_www/cordova.js'));
+    const srcCordovaJsPath = path.join(projectPath, 'www/cordova.js');
+    const platformWwwDir = path.join(projectPath, 'platform_www');
+    fs.ensureDirSync(platformWwwDir);
+    fs.copySync(srcCordovaJsPath, path.join(platformWwwDir, 'cordova.js'));
```
raphinesse, line 31: Since `fs.copySync ` creates missing target directories, the following should suffice:
```suggestion
    fs.copySync(
        path.join(projectPath, 'www/cordova.js'),
        path.join(projectPath, 'platform_www/cordova.js')
    );
```

package.json
```diff
@@ -39,6 +40,7 @@
   },
   "dependencies": {
     "cordova-common": "^4.0.2",
+    "cordova-js": "^6.1.0",
```
raphinesse, line 43: Shouldn't this be a `devDependency`?

raphinesse (commented):
I cannot test this, but it looks good to me (my two comments aside)

package.json
erisu: Yeah, I guess I didnt reinstall it correctly on this branch.

raphinesse (approved):

PR #1198, opened 2021-11-17
refactor(create): cleanup createProject

raphinesse:
### Motivation and Context

This further cleans up the create module, this time focusing on the module entry function, `createProject`.

### Description

There should only be minor behavioral changes:
- method is 100% `async` by marking it as such
- log output slightly simplified
- does not fail anymore if parent of target dir does not exist

### Testing

`npm t`


breautek (approved):
lgtm

PR #1199, opened 2021-11-17
refactor(create): further cleanup

raphinesse:
### Motivation and Context

Further simplification of the create module. This should not change behavior at all.

### Description

Highlights:

- move towards SLA
- simpler CordovaLib path configuration
- integrate the `utils` module into the `create` module

### Testing

`npm t`


erisu (approved):
Other then some of the variables going back to snake_case, everything else looks OK.

PR #1201, opened 2021-11-18
refactor(create): convert main implementation to class

raphinesse:
### Motivation and Context

Last PR of @erisu's and my `create` cleanup effort.

Note: Ignore whitespace when viewing the diff.

### Description

This converts the create implementation to a class, so we can avoid excessive argument passing.

### Testing

`npm t`


erisu (approved):
LGTM

PR #1202, opened 2021-11-19
refactor: do not infer project root from script location

erisu:
### Motivation and Context

The goal of this PR is to eliminate all areas where our JS library expects to be located inside the project directory.

This is to prepare for a change that will keep the library in `node_modules` and not copy it to the platform project.

### Description

Basically this passes the project root that has been passed to `Api` to all places that need it but do not yet have it.

~~This also extends the E2E tests to check that we do not have to require the Api from the platform project folder.~~

### Testing

- `npm t`

### Checklist

### DIFF
--- a/bin/templates/scripts/cordova/Api.js
+++ b/bin/templates/scripts/cordova/Api.js
@@ -73,7 +73,7 @@ class Api {
     constructor (platform, platformRootDir, events) {
         // 'platform' property is required as per PlatformApi spec
         this.platform = platform || 'ios';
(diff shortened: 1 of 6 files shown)


raphinesse (approved):
I pushed a few commits with _suggestions_. Explanations are in the respective commit messages. Otherwise these changes look good to me

raphinesse:
The added test mentioned in the PR comment is not here, but that's OK. I take it that you just copied the text from my corresponding android PR? :grin:

PR #1203, opened 2021-11-22
refactor!: do not copy JS lib to platform project

erisu:
### Motivation and Context

The goal of this PR is to remove the pratice of copying the entire JS lib of `cordova-ios` into each generated platform project.

### Description

- create: do not copy our JS lib (or `node_modules`) to platform project
- Move files around
  - `bin/lib/create.js` to `lib/`
  - `bin/templates/scripts/cordova/lib` to `lib/`
  - `bin/templates/scripts/cordova/Api.js` to `lib/`

After these changes, the `cordova/` folder of a generated project only contains the bare minimum to maintain compatibility with existing cordova CLI versions (plus `defaults.xml` which IMHO should be moved in another PR). All of the remaining JS files in that folder require their dependencies via the `cordova-ios` module, so that needs to be resolvable somehow (which is the case for a platform project generated by the regular CLI workflow).

### Testing

- existing tests pass
- I successfully created and built a cordova-ios project using cordova@10


raphinesse (approved):
:tada:

yoldar:
@erisu Hi! Could you please tell me how to change from old method of getting /cordova/lib/projectFile.js to new one?
Here is code:
`// try cordova 7.0 structure`
`const IosPlatformApi = require( path.join( iosPlatformPath(), "/cordova/Api" ) );`
`const projectFileApi = require( path.join( iosPlatformPath(), "/cordova/lib/projectFile.js" ) );`
`const locations      = ( new IosPlatformApi() ).locations;`
`projectFile = projectFileApi.parse( locations );`

breautek:
> @erisu Hi! Could you please tell me how to change from old method of getting /cordova/lib/projectFile.js to new one? Here is code: `// try cordova 7.0 structure` `const IosPlatformApi = require( path.join( iosPlatformPath(), "/cordova/Api" ) );` `const projectFileApi = require( path.join( iosPlatformPath(), "/cordova/lib/projectFile.js" ) );` `const locations = ( new IosPlatformApi() ).locations;` `projectFile = projectFileApi.parse( locations );`

the lib files remain inside `node_modules` now, instead of being copied into the ios xcode project.

`projectFile.js` is not exposed in a public way, so if you need it, it would be great to learn why and maybe in the future we can add a public interface into it.

A hacky way getting access to `projectFile.js` is to simply require it: `require('cordova-ios/lib/projectFile.js')`.

Note the old way that you posted might still be required if you intend to support cordova-ios 6 and earlier. [Here](https://github.com/kelvinhokk/cordova-plugin-localization-strings/pull/85/files) is an example of how this plugin does it to support both versions.

PR #1210, opened 2021-12-07
feat: support extending export options

### LINKED ISSUE #1209: Support extending export options
# Feature Request

## Motivation Behind Feature

When building the ios app I want to set some additional [`exportOptions`](https://github.com/apache/cordova-ios/blob/07383c17d8cb30c6e36fdccec52c90b4c6ba64e2/lib/build.js#L218), for example set `uploadSymbols` to `false`.
*(I need this to not include symbols in my build when uploading it to the app store)*

## Feature Description

Basically I would want something like this
*(Depending on if it is useful to overwrite defaults or not)*
```js
const exportOptions = { ...buildOpts.exportOptions, compileBitcode: false, method: 'development' };
```

Right now I need to run `xcodebuild` manually instead of using the build-in functionality.

dertieran:
### Motivation and Context

Support extending the `exportOptions` through the `buildOpts`.

Resolves #1209 

### Checklist

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -215,7 +215,7 @@ module.exports.run = function (buildOpts) {
 
             const project = createProjectObject(projectPath, projectName);
             const bundleIdentifier = project.getPackageName();
-            const exportOptions = { compileBitcode: false, method: 'development' };
+            const exportOptions = { ...buildOpts.exportOptions, compileBitcode: false, method: 'development' };
 
             if (buildOpts.packageType) {
                 exportOptions.method = buildOpts.packageType;


erisu (approved):
Looks good to me

erisu:
Code wise everything looks good.

Can you provide me your example build command on how you tested it. How you set a single or multiple export options.

dertieran:
Not quite sure if this is what you mean, but I'm not using cordova through the cli.
So I didn't provide a way to set the `exportOptions` through the cli, but I guess it could be added as an option to the `build.json`.

I tested it through my build script that looks something like this
```js
const build = async function ({ release, platforms }) {
    const exportOptions = { uploadSymbols: !release };
    const options = { release, device: release, exportOptions };
    await cordova.build({ verbose: !release, platforms, options });
};

build({ release: true, platforms: ['ios'] });
```

This will then not include the `.symbols` in the produced `.ipa` file.
This command should not print anything in that case *(by default it does)*. 
```sh
unzip -l platforms/ios/build/device/<app>.ipa | grep Symbols
```

Also as mentioned in the issue, right now the code does not allow to overwrite the defaults.
Not sure if that would be a good idea or if it would break the whole build. 😅

I hope this helps, let me know if I can help with anything else.

dpogue (approved):

expcapitaldev:
hi guys, you forgot to add here 
```
const buildConfigProperties = [
    'codeSignIdentity',
    'provisioningProfile',
    'developmentTeam',
    'packageType',
    'buildFlag',
    'iCloudContainerEnvironment',
    'automaticProvisioning',
    'authenticationKeyPath',
    'authenticationKeyID',
    'authenticationKeyIssuerID',
	**'exportOptions'**
];
```

hooliapps:
Hello, with this would it possible to auto-fill signing team key please ?

PR #1245, opened 2022-08-10
(iOS) Avoid repetitive permission prompts

### LINKED ISSUE #1166: [enhancement] add requestMediaCapturePermissionForOrigin support to avoid multiple permission request on getUserMedia call
# Feature Request

## Motivation Behind Feature
If use native WebRTC from WKWebView inside Cordova, every getUserMedia leads to new permission  request to user. And this makes native WebRTC almost not suitable for use in production. 

## Feature Description
The simplest solution is add 
```
- (void)webView:(WKWebView *)webView
  requestMediaCapturePermissionForOrigin:(WKSecurityOrigin *)origin
  initiatedByFrame:(WKFrameInfo *)frame
  type:(WKMediaCaptureType)type
  decisionHandler:(void (^)(WKPermissionDecision decision))decisionHandler
  API_AVAILABLE(ios(15.0))
{
  decisionHandler(WKPermissionDecisionGrant);
}
```
to https://github.com/apache/cordova-ios/blob/master/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewUIDelegate.m
but it's definitely insecure solution, and application should have ability to respond to this request, or provide list of safe origins.

Didn't find any

jancabadaj:
### Platforms affected
iOS

### Motivation and Context

resolves #1166 

### Description

If the app uses WebRTC, the user is initially prompted for permission twice, and then prompted again each time the app is restarted.

The solution is inspired by react:
https://github.com/react-native-webview/react-native-webview/pull/2257

### Testing

I tested the changes in a project that is using WebRTC camera streaming.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -189,6 +189,7 @@
 
     CDVWebViewUIDelegate* uiDelegate = [[CDVWebViewUIDelegate alloc] initWithViewController:self.viewController];
     uiDelegate.title = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleDisplayName"];
+    uiDelegate.mediaPermissionGrantType = [self parsePermissionGrantType:[settings cordovaSettingForKey:@"MediaPermissionGrantType"]];
     uiDelegate.allowNewWindows = [settings cordovaBoolSettingForKey:@"AllowNewWindows" defaultValue:NO];
     self.uiDelegate = uiDelegate;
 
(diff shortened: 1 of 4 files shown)


vbraun:
Works for me, for what its worth!

I would consider setting the default to "grant" because nobody wants the repeated permission request.

agm:
Please merge to avoid all related workarounds

erikramalho:
please merge it!

jancabadaj:
I fixed merge issues and changed default to 'grant' as suggested

erikramalho:
Please, merge it guys!

dpogue:
Thanks for doing the work to implement this!

With the "grant by default" behaviour, does that mean things like 3rd party iframes on a page would automatically be able to initiate webcam and microphone access? If so, I don't think we can have it that permissive by default.

A reasonable middle ground would probably be the "grantIfSameHostElsePrompt" behaviour so that the app's same-origin content can access those APIs but any 3rd party origins would result in a prompt.

jancabadaj:
Yes, that is correct. There is still an initial permission request, but once the app has permission, any site can access the webcam and microphone with "grant". 

In our use case that is exactly what we need, since we are taking a photo from a website within iframe, but I agree that "grantIfSameHostElsePrompt" might be better as default. I will change it.

dpogue (approved):

erikramalho:
Do we have an estimated release date for version 8.0.0?

dpogue:
> Do we have an estimated release date for version 8.0.0?

@erikramalho No estimated release date because it's entirely dependent on volunteer time to investigate and address the remaining bugs and new features.

erikramalho:
Thanks for the answer @dpogue.  I'll ask my team to look and see if we can help with any remaining bugs and new features.

PR #1249, opened 2022-08-26
feat: Add LimitsNavigationsToAppBoundDomains configuration key

gazben:
### Platforms affected

IOS

### Motivation and Context

Currently there is no option to set the limitsNavigationsToAppBoundDomains attribute from the `config.xml`.  If you want to use cookie authentication or browser APIs you want to set this to `YES`.

Official documentation: https://webkit.org/blog/10882/app-bound-domains/

### Description

This changeset is copied from this tutorial: https://www.ryseapp.io/blog/engineering/step-by-step-guide-to-setup-and-configure-cordova-for-ios-and-android-to-load-a-hosted-react-app

### Testing

Tested on multiple simulator instances and and physical hardware. Without it the cookie authentication for a single domain will not work. After I add this change the authentication works.

### Checklist

### DIFF
--- a/CordovaLib/cordova.js
+++ b/CordovaLib/cordova.js
@@ -1,5 +1,5 @@
 // Platform: ios
-// 538a985db128858c0a0eb4dd40fb9c8e5433fc94
+// cordova-js rel/6.0.0-10-g07379820
 /*
  Licensed to the Apache Software Foundation (ASF) under one
  or more contributor license agreements.  See the NOTICE file
--- a/CordovaLib/cordova.js
+++ b/CordovaLib/cordova.js
@@ -1,5 +1,5 @@
 // Platform: ios
-// cordova-js rel/6.0.0-10-g07379820
+// 538a985db128858c0a0eb4dd40fb9c8e5433fc94
 /*
  Licensed to the Apache Software Foundation (ASF) under one
  or more contributor license agreements.  See the NOTICE file
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -145,6 +145,10 @@
         
     }
 
+    if (@available(iOS 14.0, *)) {
+        configuration.limitsNavigationsToAppBoundDomains = [settings cordovaBoolSettingForKey:@"LimitsNavigationsToAppBoundDomains" defaultValue:NO];
+    }
+
     return configuration;
 }
 


gazben:
I don't know how can I write a test for this honestly.

gazben:
You can only list 10 `WKAppBoundDomains` in your config file. I think if we automatically merge the `hostname`, `allow-navigation` etc... We can hit this limit pretty quickly, and I don't know what should we do if we hot this limit.

You are right navigating to a not set domain will trigger an error:

> Setting this flag indicates to WebKit that a WKWebView will only navigate to app-bound domains. Once set, any attempt to navigate away from an app-bound domain will fail with the error: “App-bound domain failure.”

I can create a separete merge request for the `master` but I think it would be better to just provide this option to users so they don't have to resort to manual pathing or forking the project. This flag is an optional thing even for native projects.

bryan-hunter:
> ### Platforms affected
> IOS
> 
> ### Motivation and Context
> Currently there is no option to set the limitsNavigationsToAppBoundDomains attribute from the `config.xml`. If you want to use cookie authentication or browser APIs you want to set this to `YES`.
> 
> Official documentation: https://webkit.org/blog/10882/app-bound-domains/
> 
> ### Description
> This changeset is copied from this tutorial: https://www.ryseapp.io/blog/engineering/step-by-step-guide-to-setup-and-configure-cordova-for-ios-and-android-to-load-a-hosted-react-app
> 
> ### Testing
> Tested on multiple simulator instances and and physical hardware. Without it the cookie authentication for a single domain will not work. After I add this change the authentication works.
> 
> ### Checklist
> * [x]  I've run the tests to see all new and existing tests pass
> * [ ]  I added automated test coverage as appropriate for this change
> * [x]  Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
> * [ ]  If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
> * [ ]  I've updated the documentation if necessary

Thanks for adding this @gazben ! I'm not sure why I didn't contribute this back here and just put it in the blog post, but you're being a better open source contributor than me :)

ochakov (approved):

CordovaLib/cordova.js
```diff
@@ -1,5 +1,5 @@
 // Platform: ios
-// 538a985db128858c0a0eb4dd40fb9c8e5433fc94
+// cordova-js rel/6.0.0-10-g07379820
```
erisu, line 2: This should be reverted.
```suggestion
// 538a985db128858c0a0eb4dd40fb9c8e5433fc94
```

erisu (approved):

dpogue (approved):
This looks good to me as a stepping stone to exploring the full potential app-bound domains.

For now, this will allow people to turn on app-bound domains, and they can use `edit-config` in config.xml to add domains to their `Info.plist` file. In the future we can explore whether it makes sense to automatically populate app-bound domains from `access`/`allow-navigation`/`allow-intent` tags, but I don't think we have enough information about the exact behaviour of app-bound domains to make that call yet.

gazben:
As an FYI I've been using this change in production for half a year now (~2k users) no problems reported yet. My comment still stands from earlier, that I don't think that the automatic population is feasible because of the heavy size limitation.

bryan-hunter:
> As an FYI I've been using this change in production for half a year now (~2k users) no problems reported yet. My comment still stands from earlier, that I don't think that the automatic population is feasible because of the heavy size limitation.

Same here 👍

PR #1258, opened 2022-09-14
(ios) fix: workaround for DisallowOverscroll on iOS 16

### LINKED ISSUE #1244: DisallowOverscroll does not work on iOS 16
# Bug Report

## Problem
The same code that worked as expected on iOS 15 seems to now be broken on the latest iOS 16 public beta. DisallowOverscroll used to successfully prevent the full app from bouncing, but on iOS 16 seems to have no effect.

The full app should not bounce on scroll.

The app bounces on scroll

## Information
We have investigated the issue and it seems like [setting the scrollView bounces to NO](https://github.com/apache/cordova-ios/blob/f12abcc9973fd0618ddc8999df3fa2e96468ecf1/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m#L392) doesn't fulfill its purpose anymore. 

This would imply a change on iOS 16 side, but it is unclear if this is a bug or the new intended behaviour and thus a necessary change for cordova-ios.

Note:
We could fix the issue in our application by additionally setting `alwaysBounceVertical` to `NO` but [this doesn't align with Apple's documentation](https://developer.apple.com/documentation/uikit/uiscrollview/1619383-alwaysbouncevertical?language=objc). The fix did however not work on a clean cordova-ios project so there must be some other factor to this that we could not yet determine.

### Command or Code
The issue is reproducible with the base Cordova iOS app, with the "DisallowOverscroll" config set to TRUE: the same application behaves differently on iOS 15 than on iOS 16 when scrolling (at least for vertical scrolls).

iOS 16, iPad

### Version information
cordova 11.0
cordova-ios 6.2.0 

## Checklist

- [X] I searched for existing GitHub issues
- [X] I updated all Cordova tooling to most recent version
- [X] I included all the necessary information above

vbraun:
### Platforms affected

iOS 16

### Motivation and Context

iOS 16 breaks DisallowOverscroll setting, this PR adds workaround discussed in https://github.com/apache/cordova-ios/issues/1244

### Description

Disabling both `webView.scrollView.bounce` and `webView.scrollView.alwaysBounceVertical`

### Testing

Manually tested on iOS 15 and iOS 16

### Checklist

- [X] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))

Closes #1244


dpogue (commented):
I feel like for consistency, we do also want to set `scrollView.alwaysBounceHorizontal = NO;` otherwise there's a risk of unwanted side-to-side rubberbanding.

dpogue (approved):

u01jmg3:
@dpogue when do you intend to release 6.2.1?

u01jmg3:
This issue looks to have been fixed in iPadOS 16.2 released yesterday (Dec 13th).

PR #1272, opened 2022-11-22
fix: navigation delegation loops

msmtamburro:
note: works around, rather than addresses existing architectural issues (#1030)
note: should be reviewed after merging this related PR https://github.com/apache/cordova-ios/pull/1050

### Platforms affected

cordova-ios

### Motivation and Context

This is related to https://github.com/apache/cordova-ios/pull/1030, in that I noticed the existing looping/reflection pattern in CDVWebViewEngine was copied, and I wanted to provide an example of how to allow consumers to explicitly be delegates, so that pattern, which crashes, does not make its way in to other pieces of code.

### Description

The existing behavior, that CDVViewController can be a WKNavigationDelegate, was not obvious, so I documented it.  While it would be more explicit to set the navigation delegate in the designated initializer (like is done with the WKWebViewConfiguration), I did not want to drastically change what's being done here.   The existing behavior also loops through the plugins and uses reflection to see if any handle specific WKNavigationDelegate methods.  The way this is done can crash (is crashing in my production app), as the list of plugins can change while this loop iterates.

What I did change:
- When a consumer has bothered to extend CDVViewController to be a WKNavigationDelegate, I've disabled the looping
- I cleaned up the looping to make it less likely to crash

### Testing

I've tested the existing behavior:
- Overriding does not disable looping
- Looping does not appear to crash

I've tested the new behavior:
- Looping is disabled when overriding (making CDVViewController a WKNavigationDelegate)

### Checklist


msmtamburro:
Thanks for your patience; I've re-raised this PR now that I've gotten access at work again.

msmtamburro:
Issue with the codecov.io setup?
{'detail': ErrorDetail(string='Missing "owner" argument. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}

breautek:
I'll rerun the test, I think that happens intermediately sometimes.

msmtamburro:
Before I lose access again (through work), this one seems like it is good to go.

dpogue:
Handled the loop issue slightly differently in https://github.com/apache/cordova-ios/pull/1479 but with the same idea of creating a copy of the map before iterating over the plugins. Also added synchronization to help ensure that we're not adding plugins while generating the copy of the values.

Thanks for clearly explaining the problem!

I intend to review the navigation policy stuff (in the context of additions like https://github.com/apache/cordova-ios/pull/1333) and will try to bring those changes in.

PR #1274, opened 2022-12-05
feat: add preference to disable MacOS target

tiagoappereira:
Xcode 14 considers that iOS apps should support by default not only iPhone and iPad, but also Mac (with Apple Silicon processors). With the introduction of “Designed for iPad“, IPAs can be installed on these devices, and Apple wants to make this the default behaviour/experience.

Also, when apps with the Mac target are submitted to the App Store they are available for Mac by default as well. They can be found in a specific “iPhone & iPad Apps“ tab.

Therefore, with the suggested preference in this PR, developers can disable the aforementioned default behaviour.

### Description

To disable Mac target, the following lines need to be added to the .pbxproj:
```
SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
SUPPORTS_MACCATALYST = NO;
SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = NO;
```

The Mac support removal (or not) is controlled by a preference called `SupportMac`:
| SupportMac  | Mac target  |
| ------------- | ----------- |
| N/A                 |  enabled     |
| true                |  enabled     |
| false               |  disabled    |

### Testing

Created a local iOS Cordova project based on this cordova-ios branch and the .pbxproj was:

- NOT updated when the preference SupportMac is not set
- NOT updated when the preference SupportMac is set with the value true
- updated when the preference SupportMac is set with the value false

### Checklist

### DIFF
--- a/tests/spec/unit/prepare.spec.js
+++ b/tests/spec/unit/prepare.spec.js
@@ -649,6 +649,60 @@ describe('prepare', () => {
             });
         });
 
+        it('should support Mac apps by default', () => {
+            cfg2.name = () => 'SampleApp'; // new config does *not* have a name change
+            writeFileSyncSpy.and.callThrough();
+            return updateProject(cfg2, p.locations).then(() => {
+                const xcode = require('xcode');
+                const proj = new xcode.project(p.locations.pbxproj); /* eslint new-cap : 0 */
+                proj.parseSync();
+                const supportedPlatforms = proj.getBuildProperty('SUPPORTED_PLATFORMS');
+                expect(supportedPlatforms).toBeUndefined();
+                const supportsMacCatalyst = proj.getBuildProperty('SUPPORTS_MACCATALYST');
+                expect(supportsMacCatalyst).toBeUndefined();
+                const supportsMacIPhoneIPad = proj.getBuildProperty('SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD');
+                expect(supportsMacIPhoneIPad).toBeUndefined();
+            });
+        });
+
+        it('should support Mac apps if SupportMac preference is set to true', () => {
+            cfg3.name = () => 'SampleApp'; // new config does *not* have a name change
+            const pref = cfg3.doc.findall('platform[@name=\'ios\']/preference')
+                .filter(elem => elem.attrib.name.toLowerCase() === 'supportmac')[0];
+            pref.attrib.value = 'true';
+            writeFileSyncSpy.and.callThrough();
+            return updateProject(cfg3, p.locations).then(() => {
+                const xcode = require('xcode');
+                const proj = new xcode.project(p.locations.pbxproj); /* eslint new-cap : 0 */
+                proj.parseSync();
+                const supportedPlatforms = proj.getBuildProperty('SUPPORTED_PLATFORMS');
+                expect(supportedPlatforms).toBeUndefined();
+                const supportsMacCatalyst = proj.getBuildProperty('SUPPORTS_MACCATALYST');
+                expect(supportsMacCatalyst).toBeUndefined();
+                const supportsMacIPhoneIPad = proj.getBuildProperty('SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD');
+                expect(supportsMacIPhoneIPad).toBeUndefined();
+            });
+        });
+
+        it('should not support Mac apps if SupportMac preference is set to false', () => {
+            cfg3.name = () => 'SampleApp'; // new config does *not* have a name change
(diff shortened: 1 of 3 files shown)


tests/spec/unit/prepare.spec.js
```diff
@@ -649,6 +649,60 @@ describe('prepare', () => {
             });
         });
 
+        it('should support Mac apps by default', () => {
+            cfg2.name = () => 'SampleApp'; // new config does *not* have a name change
+            writeFileSyncSpy.and.callThrough();
+            return updateProject(cfg2, p.locations).then(() => {
+                const xcode = require('xcode');
+                const proj = new xcode.project(p.locations.pbxproj); /* eslint new-cap : 0 */
```
breautek, line 657: `xcode.project` shouldn't be called with `new`. Additionally, the test file imports it as

```javascript
const xcode = require('xcode');
const XcodeProject = xcode.project;
```

So you can use `XcodeProject` for consistency.

```suggestion
                const proj = XcodeProject(p.locations.pbxproj);
```

I think this will fix:

```
1) prepare updateProject method should support Mac apps by default
  Message:
    TypeError: target is not a constructor
  Stack:
    TypeError: target is not a constructor
        at <Jasmine>
        at /Users/runner/work/cordova-ios/cordova-ios/tests/spec/unit/prepare.spec.js:657:30
```

tests/spec/unit/prepare.spec.js
```diff
@@ -649,6 +649,60 @@ describe('prepare', () => {
...
+
+        it('should support Mac apps if SupportMac preference is set to true', () => {
+            cfg3.name = () => 'SampleApp'; // new config does *not* have a name change
+            const pref = cfg3.doc.findall('platform[@name=\'ios\']/preference')
+                .filter(elem => elem.attrib.name.toLowerCase() === 'supportmac')[0];
+            pref.attrib.value = 'true';
+            writeFileSyncSpy.and.callThrough();
+            return updateProject(cfg3, p.locations).then(() => {
+                const xcode = require('xcode');
+                const proj = new xcode.project(p.locations.pbxproj); /* eslint new-cap : 0 */
```
breautek, line 676: See above

```suggestion
                const proj = XcodeProject(p.locations.pbxproj);
```

tests/spec/unit/prepare.spec.js
```diff
@@ -649,6 +649,60 @@ describe('prepare', () => {
...
+
+        it('should not support Mac apps if SupportMac preference is set to false', () => {
+            cfg3.name = () => 'SampleApp'; // new config does *not* have a name change
+            const pref = cfg3.doc.findall('platform[@name=\'ios\']/preference')
+                .filter(elem => elem.attrib.name.toLowerCase() === 'supportmac')[0];
+            pref.attrib.value = 'false';
+            writeFileSyncSpy.and.callThrough();
+            return updateProject(cfg3, p.locations).then(() => {
+                const xcode = require('xcode');
+                const proj = new xcode.project(p.locations.pbxproj); /* eslint new-cap : 0 */
```
breautek, line 695: See above

```suggestion
                const proj = XcodeProject(p.locations.pbxproj);
```

dpogue:
We have an existing `target-device` preference that supports `handset`, `tablet`, and `universal` as values, and sets the `TARGETED_DEVICE_FAMILY`.

Maybe we can find a way to add this functionality to that existing preference?

tiagoappereira:
> We have an existing `target-device` preference that supports `handset`, `tablet`, and `universal` as values, and sets the `TARGETED_DEVICE_FAMILY`.
> 
> Maybe we can find a way to add this functionality to that existing preference?

If the idea is to expand the `target-device` supported values, doing so would change what the preference represents. From the [official documentation](https://cordova.apache.org/docs/en/11.x/config_ref/index.html#:~:text=Deployment%20Target%20Settings-,target%2Ddevice,-(string)%0A%3D%3DiOS)

> This property maps directly to TARGETED_DEVICE_FAMILY in the xcode project.

which would no longer be true.

dpogue:
The documentation can be updated to reflect what the preference actually does, so I wouldn't worry too much about that aspect.

I don't have strong feelings about it, but I do wonder if it's easier or more consistent to have a single preference that controls the output platforms (similar to the 3 checkboxes in Xcode) vs a combination of multiple preferences that might interact in unusual ways.

dpogue:
My current thinking around this is to disable the iPad-on-macOS thing by default, and to allow explicitly building a macOS app from the same codebase targeting Catalyst. That means there would be an extra step for those who want to distribute a macOS app, but it should be a better result (in terms of user experience and platform fit) than the iPad-on-macOS option.

PR #1276, opened 2022-12-24
feat: Add support for Apple Cloud Distribution signing

dpogue:
### Platforms affected
iOS

### Motivation and Context

This enables automatic provisioning with Apple Cloud Distribution certificates (which is really useful for cloud CI systems).

### Description

This adds 3 new properties that can be specified on the command-line or via build.json, matching the 3 properties that need to be passed to xcodebuild to allow automatic cloud signing.

I have some notes about how this works and how I set up automatic signing with GitHub Actions for a non-Cordova project here: https://gist.github.com/dpogue/d11df53cca6ba0e23d14d674ee5f9762

### Testing

Unit tests updated, still needs some further testing.

### Checklist

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -28,6 +28,20 @@ const util = require('util');
 const check_reqs = require('./check_reqs');
 const projectFile = require('./projectFile');
 
+const buildConfigProperties = [
+    'codeSignIdentity',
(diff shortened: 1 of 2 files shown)


erisu (approved):
The code changes LGTM

PR #1284, opened 2023-02-07
Patch 1

xiaosiZhang:
### Checklist


dpogue:
`_allowUniversalAccessFromFileURLs` is private API and isn't allowed by Apple.

The correct approach is to use a custom scheme (such as `app://`) to serve content rather than `file:///` URLs. This is supported since [Cordova iOS 6.0.0](https://cordova.apache.org/announcements/2020/06/01/cordova-ios-release-6.0.0.html).

PR #1295, opened 2023-03-03
fix(CDVPlugin): Swift init

### LINKED ISSUE #1288: convenience init behavior within `CDVPlugin.m` for conformance to Swift init and `let`
# Bug Report
Currently CDVPlugin.m uses a `self = [super init]` pattern within the `initWithWebViewEngine:` convenience initializer.

## Problem
The designated initializer pattern is used in place of a convenience initializer pattern.
🧶👷 `convenience` initializers should invoke required `designated initializers`
           using `self`, as in `self = [self init]` ｢not `self = [super init]`｣.
Specifically, when building a Swift based CDVPlugin subclass its `init` never gets invoked.

Additionally, on iOS, this has a negative impact on plugin design because `pluginInitialize` is
NOT an `init` method and it is only invoked (later) once the plugin has been initialized and registered.

Which leads to ⟨confusion in design patterns⟩ many temporal configuration times within the plugin architecture.
1) CDVPlugin subtype inst `self` being constructed and initialized. ｢designated init｣
    which DOES NOT happen now, but needs to for proper Swift `init` designs.
2) `pluginInitialize` when plugin is registered-as-ready
3) JavaScript **plugins** module/source has been loaded and configured the JS wrapper
4) `deviceReady` when JavaScript layer has been notified that Cordova itself is all configured.

The Objective-C/Swift `convenience` initializer pattern for `initWithWebViewEngine` invoking
`self = [self init]` ｢not `self = [super init]`｣.

Currently the convenience-init `initWithWebViewEngine` is written as if it were a designated-init calling a convenience-init.

It is using  `self = [super init]`, which results in a Swift subclass never having its `init` member
called and thus it never gets properly initialized if it has an `override init` or any memberwise initializers.
Which, in turn, means it may not compile if it used `let` consts members or it may simply crash due
to being uninitialized.

## Information
explained in other parts of this submission...

### Command or Code
Try the following changes with the swift plugin example below.
`self = [super init]` vs `self = [self init]` within `CDVPlugin.m` method  `initWithWebViewEngine:`.

```swift
class example : CDVPlugin {
  let a: String = "never actually set";
  let b: String; // `let` can only be set in an `init` func
  override init() {self.b = "never actually set either";}
}
```
### Version information
- cordova platform version "ios 6.2.0"
- OSX version "Ventura 13.2.1"

## Checklist
- [ x] I searched for existing GitHub issues
- [ x] I updated all Cordova tooling to most recent version
- [ x] I included all the necessary information above

dpogue:
### Platforms affected
iOS

### Motivation and Context

Fixes #1288 

### Description

This ensures that the default `init` method is called for CDVPlugin subclasses, to allow Swift initializers to run as expected.

Aside: I still feel like `NS_DESIGNATED_INITIALIZER` would be the theoretically correct way to solve this, but that doesn't work when our designated initializer with the webview is part of private implementation details (your designated initializer must be part of your public API surface).
The risk is that someone could (in their own code) do `[[MyCordovaPlugin alloc] init]` and try to construct a plugin instance without providing the webview engine, and lead to all sorts of weird issues.

However, that problem already existed and people don't seem to be shooting themselves in the foot with it (or at least, not complaining here when they do), so it's probably better to fix the issue that is impacting people even if the solution isn't theoretically perfect.

### Testing

Added a unit test which ensures that an init-time property in a Swift plugin is correctly set after the plugin has been initialized. This test case fails without the changes to CDVPlugin.m.

### Checklist

### DIFF
--- a/tests/CordovaLibTests/CordovaLibTests.xcodeproj/project.pbxproj
+++ b/tests/CordovaLibTests/CordovaLibTests.xcodeproj/project.pbxproj
@@ -22,6 +22,10 @@
 		686357B5141002F200DF4CF2 /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = 686357B3141002F200DF4CF2 /* InfoPlist.strings */; };
 		686357BA141002F200DF4CF2 /* CDVPluginResultJSONSerializationTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 686357B9141002F200DF4CF2 /* CDVPluginResultJSONSerializationTests.m */; };
 		7EF33BD71911ABA20048544E /* Default-568h@2x.png in Resources */ = {isa = PBXBuildFile; fileRef = 7EF33BD61911ABA20048544E /* Default-568h@2x.png */; };
+		902530FB29A6DC53004AF1CF /* CDVPluginInitTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 902530FA29A6DC53004AF1CF /* CDVPluginInitTests.m */; };
+		902530FC29A6DC53004AF1CF /* CDVPluginInitTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 902530FA29A6DC53004AF1CF /* CDVPluginInitTests.m */; };
+		908630C829A6D003002963FA /* SwiftInitPlugin.swift in Sources */ = {isa = PBXBuildFile; fileRef = 908630C729A6D003002963FA /* SwiftInitPlugin.swift */; };
+		908630C929A6D003002963FA /* SwiftInitPlugin.swift in Sources */ = {isa = PBXBuildFile; fileRef = 908630C729A6D003002963FA /* SwiftInitPlugin.swift */; };
 		C0FA7C9B1E4BB6420077B045 /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 303A4073152124BB00182201 /* main.m */; };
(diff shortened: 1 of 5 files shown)


tests/CordovaLibTests/CordovaLibTests.xcodeproj/project.pbxproj
```diff
@@ -551,7 +563,7 @@
 				PUBLIC_HEADERS_FOLDER_PATH = include/Cordova;
 				SDKROOT = iphoneos;
 				SKIP_INSTALL = YES;
-				SWIFT_OBJC_BRIDGING_HEADER = "__PROJECT_NAME__/Bridging-Header.h";
+				SWIFT_OBJC_BRIDGING_HEADER = "CordovaLibApp/Bridging-Header.h";
```
erisu, line 566: Can you confirm if this is expected?

tests/CordovaLibTests/CordovaLibTests.xcodeproj/project.pbxproj
```diff
@@ -604,7 +616,7 @@
...
```
erisu, line 619: Can you confirm if this is expected?

erisu (changes requested):
Can you confirm if these two changes are expected?

The rest appears OK.

Will also need a rebase if possible.

tests/CordovaLibTests/CordovaLibTests.xcodeproj/project.pbxproj
dpogue: yes, in the test project, the folder name is `CordovaLibApp`. Previously this didn't cause problems because we had no Swift code in the tests, but now it needs to be accurate.

erisu (approved):
Everything looks good to me. 👍 thanks!

Issue #1290, opened 2023-02-24
access domains not handled properly on NodeJS 18 


breautek:
# Bug Report

## Problem

Unit test fails on NodeJS 18, because the values returned by `URL.parse` has changed in NodeJS 18.

https://github.com/apache/cordova-ios/blame/master/lib/prepare.js#L1029

As of NodeJS 18, the parsed URL now contains ` .hostname` property that is a non-empty string, whereas older Node versions returned `hosname` as an empty string, when given a URL that contains a wildcard such as `https://*.test.com`.

Here is a summary of how the parsed URL has changed going into Node 18.
```
v16.14.0 -> v18.14.1

var URL = require('url')
var href = 'http://*.test';
Url {
 protocol: 'http:',
 slashes: true,
 auth: null,
 host: '',----------------> '*.test'
 port: null,
 hostname: '',------------> '*.test'
 hash: null,
 search: null,
 query: null,
 pathname: '/*.test',-----> '/'
 path: '/*.test',---------> '/'
 href: 'http:///*.test'---> 'http://*.test/'
}
```

As a result, the final object when testing https://github.com/apache/cordova-ios/blame/master/tests/spec/unit/prepare.spec.js#L864

fails because the key is `*.server01.com` instead of the expected `server01.com`

### Version information
cordova-ios master
NodeJS 18

TylerBreau:
The code mentions URL.parse being deprecated.
```
function parseAllowlistUrlForATS (url, options) {
    // @todo 'url.parse' was deprecated since v11.0.0. Use 'url.URL' constructor instead.
    const href = URL.parse(url); // eslint-disable-line
```

Here's a summary of the differences between URL.parse and URL.URL constructor.
```
v16.14.0 (URL.parse) -> v18.14.1 (new URL.URL)

var URL = require('url')
var href = 'http://*.test';
URL.parse -> new URL.URL(href)
Url {
 protocol: 'http:',
 slashes: true,-----------> undefined
 auth: null,--------------> undefined
 host: '',----------------> '*.test'
 port: null,
 hostname: '',------------> '*.test'
 hash: null,
 search: null,
 query: null,-------------> undefined
 pathname: '/*.test',-----> '/'
 path: '/*.test',---------> undefined
 href: 'http:///*.test'---> 'http://*.test/'


 v18 introduces the following properties
 origin: 'http://*.test',
 username: '',
 password: '',
 searchParams: URLSearchParams {},
}
```

[...]

PR #1300, opened 2023-03-21
Set webView.inspectable to true for debug builds on iOS >= 16.4

BBosman:
### Platforms affected

iOS

### Motivation and Context

With the introduction of iOS 16.4 WKWebView instances will no longer be inspectable by default.

Details: [Enabling the Inspection of Web Content in Apps](https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/)

### Description

This PR honors that change for Release builds, but explicitly sets the inspectability to `YES` for Debug builds by default.
It also introduce an override option, so consumers can influence this decision in their own build.

### Testing

Tested the code in one of our own apps.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -218,6 +218,20 @@
     WKWebView* wkWebView = [[WKWebView alloc] initWithFrame:self.engineWebView.frame configuration:configuration];
     wkWebView.UIDelegate = self.uiDelegate;
 
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 160400
+    // With the introduction of iOS 16.4 the webview is no longer inspectable by default.
+    // We'll honor that change for release builds, but will still allow inspection on debug builds by default.
+    // We also introduce an override option, so consumers can influence this decision in their own build.
+    if (@available(iOS 16.4, *)) {
+#ifdef DEBUG
+        BOOL allowWebviewInspectionDefault = YES;
+#else
+        BOOL allowWebviewInspectionDefault = NO;
+#endif
+        wkWebView.inspectable = [settings cordovaBoolSettingForKey:@"InspectableWebview" defaultValue:allowWebviewInspectionDefault];
+    }
+#endif
+
     /*
      * This is where the "OverrideUserAgent" is handled. This will replace the entire UserAgent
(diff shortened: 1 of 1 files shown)


dpogue (commented):
The master branch is already targeting the next major version, so I think we want to default this to NO in release builds (and YES in debug builds).

Thanks for adding the preference too! We'll want to add that to the cordova-docs repo (and maybe file an issue on cordova-android to obey the preference for inspecting there too)

BBosman:
@dpogue I've updated the code and PR description based on your feedback.

I had to remove the default value for the setting from `defaults.xml`, as that would always override the build based default setting in the code.

dpogue (approved):
LGTM

breautek:
> I can't explain the CI failure, but it looks unrelated to my changes?

That issue is caused by https://github.com/apache/cordova-ios/issues/1290 for reference.

breautek (approved):

breautek (approved):
Disclaimer: Non-lawyer analysis.

I revoked my previous approval out of concerns that allowing webview to be inspectable may be in a clear violation of [Section 4.7.1](https://developer.apple.com/app-store/review/guidelines/#third-party-software) of the App Guidelines.

However I believe [Section 4.7.1](https://developer.apple.com/app-store/review/guidelines/#third-party-software) only applies to code that isn't bundled with the App binary. And I didn't find anything else (In the Apple Developer Agreement and XCode terms of use) that says device APIs cannot be exposed to users outside the developer platform.

Therefore, at the core, I think Cordova is in the clear, for both the core platform & use in the in app browser plugin (should the plugin also obey this kind of flag, which I know there is a PR to introduce that functionality).

If the developer does import code from remote sources then they may be breaching the agreement with Apple, however that isn't the concern of Apache Cordova, it will be up to the end-users to ensure they comply with Apple's terms.

So I'm returning my approval.

NiklasMerz (approved):
Tested with an iPad and code looks good to me. I will also merge this into a path release

daviesdoclc:
This is not working for me on iOS 16.4.1, MacOS 13.3.1, Safari 16.4.

I'm running as follows:
```cordova run ios --device --debug```

I also tried setting InspectableWebview to true in config.xml.

breautek:
> This is not working for me on iOS 16.4.1, MacOS 13.3.1, Safari 16.4.
> 
> I'm running as follows: `cordova run ios --device --debug`
> 
> I also tried setting InspectableWebview to true in config.xml.

Let's create a new issue, using the issue template so that we can try to drill down and isolate the cause.

PR #1302, opened 2023-04-09
Fix URL parsing for ATS in node 18

### LINKED ISSUE #1290: access domains not handled properly on NodeJS 18

dpogue:
### Platforms affected
iOS, with node18

### Motivation and Context

Fixes #1290: currently parsing of `access`, `allow-navigation`, and `allow-intent` directives in config.xml for Application Transport Security is broken when using node 18 due to changes to how node handles URLs.

### Description

Move URL parsing to the (non-deprecated) WHATWG URL API in node, and workaround the issues that cause URL parsing to fail or return unexpected results with that API.

### Testing

All unit tests pass, but I would feel more confident if people tried throwing some weirder corner cases at this to ensure it's actually behaving consistently with previous versions.

### Checklist

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -997,12 +997,8 @@ function processAccessAndAllowNavigationEntries (config) {
     null is returned if the URL cannot be parsed, or is to be skipped for ATS.
 */
 function parseAllowlistUrlForATS (url, options) {
-    // @todo 'url.parse' was deprecated since v11.0.0. Use 'url.URL' constructor instead.
-    const href = URL.parse(url); // eslint-disable-line
-    const retObj = {};
-    retObj.Hostname = href.hostname;
-
     // Guiding principle: we only set values in retObj if they are NOT the default
+    const retObj = {};
 
     if (url === '*') {
         retObj.Hostname = '*';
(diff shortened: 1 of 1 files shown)


erisu (approved):

NiklasMerz (approved):
Logic looks good to me. Cannot test it right now.

quyenvsp:
Just note for people who has problem Invalid url
```
    <allow-navigation href="*youtube*" />
```
should change to
```
    <allow-navigation href="http:*youtube*" />
    <allow-navigation href="https:*youtube*" />
```

Issue #1019, opened 2020-11-07
Remove placeholder icons from Images.xcassets


calrsom:
# Feature Request

## Motivation Behind Feature

Our app was rejected from the app store because it included placeholder Cordova app icons.  I have to remove Cordova icons for sizes I don't provide by hand from the Xcode project.

## Feature Description

I'd like Cordova to not include or delete all placeholder icons from the `Images.xcassets` as long as at least one `<icon />` is defined in `config.xml`.  This gives developers direct control over which icon sizes are included in their application.

As a bonus this covers the more specific feature request #658 to avoid supplying watch icons.

## Alternatives or Workarounds

Currently, the only workaround is to include every single possible size icon in order to overwrite all placeholders.  This is excessive and unnecessary.  Our app only targets iPad and iOS 11+, so I'd prefer not to have to include icons for iPhone, Apple Watch, and iOS 5 & 6.

I was hoping setting `<icon />` without height and width would generate icons for appropriate sizes, but instead it does not resize the source image at all, generating warnings in Xcode.

calrsom:
I've moved on from the project I worked on that used Cordova, but I'm glad to hear this is implemented. 🎉 
Thanks!

PR #1309, opened 2023-04-14
feat!: Use single 1024x1024 icon

jcesarmobile:
Xcode 14 allows to use a single 1024x1024 icon instead of having all the icon sizes.

It's breaking because it requires Xcode 14, but since Xcode 14.1 is going to be required for app store submissions starting the 25 of April, I think we are good.

Removed the icon sizes link because it was broken


dpogue:
We should make sure this works properly if someone just puts `<icon src="res/ios/icon.png" />` in their config.xml (with no sizes specified).

There's a part of me that thinks it would be good to have some support for customizing the notification and watch icons, but that would involve dynamically adding them to the .xcassets Contents.json only if they were specified, which feels like it would be error-prone if people are mixing manual changes and automatic ones 😞

jcesarmobile:
If using single icon, Xcode only allows a single icon, doesn’t allow to add a different icon notifications. 
It allows a different icon for watchOS, but at the moment it was using the same as for iOS

dpogue:
Ahh okay

dpogue:
Note: Closes #1019

breautek:
It's also breaking in the sense it requires iOS 12+


Screenshot 2023-05-03 at 9 48 57 AM

So I think we also need to bump CordovaLib min target to iOS 12, which doesn't have to be part of this PR, but it will need to be  set for this feature.

breautek (approved):

breautek:
FWIW, we've gone back in our apps and just supplied the icons that were missing (so that we aren't relying on a feature that may not even make it into release).

This appears to cover all current icons, at least for the iPhone models.

erisu:
I believe it was decided to staying on iOS 11, so I think this PR can be closed.

PR #1310, opened 2023-04-15
feat!: remove default build dir overrides

dpogue:
### Platforms affected
iOS

### Motivation and Context

 Specifying `CONFIGURATION_BUILD_DIR` can cause issues with Cocoapods.

Closes #617
Closes #659
Closes #671

### Description

Once upon a time, Xcode's default directory values had issues when there were spaces in project names, so we override the `CONFIGURATION_BUILD_DIR` and `SHARED_PRECOMPS_DIR` variables with our own paths.

However, this can interfere with Cocoapods, and Xcode should be able to handle things sensibly nowadays.

Unfortunately, that results in our build artifacts living somewhere in a randomly-named Xcode DerivedData directory, and then we can't do things like run the app in a simulator or on a device because we don't know the path to it. 🤦🏼‍♂️ 

Setting `SYMROOT` allows us to control the output directory of the built products, with the caveat that Xcode always creates a subdirectory named with the current configuration.

So instead of `build/emulator`, we'll have `build/Debug-iphonesimulator` and instead of `build/device`, we'll have `build/Release-iphoneos`.  
Hypothetically this is better because now we are sure that debug and release files never get mixed up in the same output directory.

The downside is that this is a **breaking change** because it alters the path for the output .ipa files.

### Testing

- [x] Unit tests pass
- [x] Created test project with spaces in the project name and confirmed that it builds successfully
- [x] Created a test project with Cocoapods that previously failed and confirmed that it builds successfully
- [x] Run a test project in the simulator (debug configuration)
- [x] Run a test project in the simulator (release configuration)
- [x] Run a test project on a device (debug configuration)
- [x] Run a test project on a device (release configuration)

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [x] I added automated test coverage as appropriate for this change
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))

PR #1313, opened 2023-04-16
feat!: Better Catalyst build support

### LINKED ISSUE #420: Why is ios-deploy required for plain builds (without deployment)?
Why am I seeing this complaint when I'm not even trying to deploy? This happens on `cordova build`:
```shell
CordovaError: Promise rejected with non-error: 'ios-deploy was not found. Please download, build and install version 1.9.2 or greater from https://github.com/phonegap/ios-deploy into your path, or do \'npm install -g ios-deploy\''
```

### LINKED ISSUE #677: cordova build command tries to run app when a device is connected
When I have a iPhone device connected to my desktop and I run `cordova build ios` command, it tries to run the app on device while it should just build the application.

So I have an error message because `ios-deploy` is not installed on my desktop.

```
[Error] Error: Command failed: cordova build --buildFlag='-UseModernBuildSystem=0' ios
CordovaError: Promise rejected with non-error: 'ios-deploy was not found. Please download, build and install version 1.9.2 or greater from https://github.com/ios-control/ios-deploy into your path, or do \'npm install -g ios-deploy\''
    at cli.catch.err (/Users/othierry/.nvm/versions/node/v8.12.0/lib/node_modules/cordova/bin/cordova:29:15)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
```

I use Cordova 9 with Cordova iOS platform 5.0.1.

I had not this problem with previous Cordova releases.

### LINKED ISSUE #1417: Add preferences to config.xml to control which are the "Supported Destinations"
# Feature Request

## Motivation Behind Feature
The only way so far to control to which devices my iOS app gets built is via `target-device`, which only accepts iPhone and/or iPad as parameters, exactly how Xcode does it at Build Settings > Deployment > Targeted Device Families.

However, there's no way to avoid an app to be available for "designed for iPad" destinations, such as Mac or Apple Vision. This is removable by going to General > Supported Destinations, but it obviously goes away once you reset your platform folder, e.g. new clone.

image

This probably ties directly with the "Compatibility" section on the App Store:
image

## Feature Description
There should be a new preference tag to control those destinations.

As mentioned, some people confuse this request with the `target-device` preference, but that doesn't allow removing those "designed for iPad" destinations, in case you do want to support actual iPads.

#1313 was suggested as a similar topic, but that's related to Catalyst, which is indeed another option on that destinations list, but to my understanding the problem is not which version to build, but instead which versions are allowed to be built - or in which device types the app is allowed to be used, in case of iPad adaptations.

dpogue:
### Platforms affected
iOS, macOS

### Motivation and Context

With the deprecation of cordova-osx and the availability of Catalyst to deploy iOS apps on macOS as their own app bundles, we should support targeting Catalyst builds.

Closes #420.
Closes #677.
Closes #1274.
Closes #1417.

### Description

After the work in #1310 caused the build output folder to be named based on the platform, I realized this actually solved one of the remaining questions around where to put Catalyst output files... now they end up in `build/Debug-maccatalyst` or `build/Release-maccatalyst`

We detect if we are building for Catalyst (as opposed to iOS Device vs iOS Simulator) by the `--target=mac` option.

### Testing

`cordova build ios --target=mac` works
`cordova run ios --target=mac` works

### Checklist


erisu (approved):
LGTM

Tested mobilespec on iOS and Catalyst builds.

Geolocation plugin might need additional investigation for Catalyst.

dpogue:
> Geolocation plugin might need additional investigation for Catalyst.

Looks like location access needs a capability to be turned on in the entitlements file for macOS/Catalyst. Once you do that, it works as expected. We can update the plugin to inject that.

PR #1315, opened 2023-04-20
fix(splashscreen): positioning after rotation

### LINKED ISSUE #1231: Splashscreen content not fitting or centered after screen rotation
# Bug Report

## Problem
When the screen is rotated, the splashscreen does not fit the screen anymore and the content is not centered. The view retains the size it had when the app started. This can be seen:
- at app startup, while the splashscreen is still visible (and rotating the screen)
- at runtime, when showing the splashscreen with `navigator.splashscreen.show()` (and the screen has been rotated)

The splashscreen should always fit the screen and has its content centered.

The splashscreen does not fit the screen. The webview content is then visible in the area of the screen not covered by the splashscreen view.

## Information
My `config.xml` uses the following (single image):
```
...
<preference name="AutoHideSplashScreen" value="false" />
<preference name="FadeSplashScreen" value="false" />
<preference name="ShowSplashScreenSpinner" value="false" />
<preference name="RemoveSplashScreenOnPause" value="false" />
...
<splash src="res/ios/Default@2x~universal~anyany.png" />
...
```

### Version information
cordova-ios: 6.2.0
Xcode: 13.3.1
iOS: 15.4.1

## Checklist

dpogue:
### Platforms affected
ios

### Motivation and Context

Closes #1231.

### Description

Add constraints to the splashscreen view so that it keeps the aspect ratio and positioning when the device is rotated.

@goffioul This is mostly the code that you proposed, do you want to be credited on the commit for this?

### Testing

Tested on an iOS Simulator.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -501,18 +501,36 @@
     webViewBounds.origin = self.view.bounds.origin;
 
     UIView* view = [[UIView alloc] initWithFrame:webViewBounds];
+    view.translatesAutoresizingMaskIntoConstraints = NO;
     [view setAlpha:0];
 
     NSString* launchStoryboardName = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"UILaunchStoryboardName"];
     if (launchStoryboardName != nil) {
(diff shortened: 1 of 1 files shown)


erisu (approved):
LGTM

goffioul:
@dpogue You created this PR, but I would indeed appreciate partial credit. Thanks for asking.

PR #1323, opened 2023-05-03
feat!: Bump minimum deployment target to 12.0

breautek:
### Platforms affected

iOS

### Motivation and Context

Some features are being introduced (such as #1309) which requires iOS 12+

### Description

Bump all occurrences `IPHONEOS_DEPLOYMENT_TARGET` to `12.0`, including the app project template, CordovaLib, and test files.

### Testing

Letting CI do the testing.

### Checklist


breautek:
Screenshot 2023-05-03 at 9 48 57 AM

dpogue:
Question: When we override the iOS deployment target via a preference, does that apply to just the app project or also to CordovaLib?

Essentially, if somebody wanted to provide all the old icon sizes using a hook or something, can they lower the deployment target back to 11 using config.xml, or will Xcode error because CordovaLib requires a higher version (even though there's not actually anything in CordovaLib that requires iOS 12)

breautek:
> It's still not clear if we are going to merge #1309 since some people might want to keep using different icons for different devices or for push notifications.
> 
> We should have a more formal mail (or vote?) about bumping the deployment target.

I had the same thought and a discuss thread should be arriving shortly. Will post link once it appears in ponymail.

erisu:
Can it be confirmed if CordovaLib can stay at 11.0, but only bump the project template to 12.0?

Maybe the project template could be increased to support new features (one-icon), but allowing anyone else who use only the CordovaLib in their non-Cordova project to continue to support iOS 11.

breautek:
> Can it be confirmed if CordovaLib can stay at 11.0, but only bump the project template to 12.0?
> 
> Maybe the project template could be increased to support new features (one-icon), but allowing anyone else who use only the CordovaLib in their non-Cordova project to continue to support iOS 11.

I didn't test via code change, but setting the deployment targets manually in xcode will still allows us to build as long as the App target is >= the deployment target of `CordovaLib` target.

The bigger concern is how XCode behaves if you attempt to configure a target to a version that it no longer supports (and is not selectable via the UI). To test this, I installed cordova-ios@5 platform which tries to set the deployment target to 10.0.

The observed behaviour is in the XCode UI, under the App target, you see Deployment Target 10 (when it's normally not selectable in the dropdown menu). Under the `CordovaLib` target, the target is set to `default`

Screenshot 2023-05-13 at 9 02 41 AM

When attempting to build this project, the following errors are produced:

```
note: Building targets in dependency order
/Users/norman/development/cordova/test/iostest/platforms/ios/CordovaLib/CordovaLib.xcodeproj: warning: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 10.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'CordovaLib' from project 'CordovaLib')
/Users/norman/development/cordova/test/iostest/platforms/ios/HelloCordova.xcodeproj: warning: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 10.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'HelloCordova' from project 'HelloCordova')
/Users/norman/development/cordova/test/iostest/platforms/ios/HelloCordova.xcodeproj: error: Signing for "HelloCordova" requires a development team. Select a development team in the Signing & Capabilities editor. (in target 'HelloCordova' from project 'HelloCordova')
warning: Run script build phase 'Copy www directory' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'HelloCordova' from project 'HelloCordova')
** ARCHIVE FAILED **
```

So there are two parts:

```
/Users/norman/development/cordova/test/iostest/platforms/ios/CordovaLib/CordovaLib.xcodeproj: warning: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 10.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'CordovaLib' from project 'CordovaLib')
```

Shows that despite the UI, CordovaLib **is** being attempted to built with a deployment target of 10.0, as expected for Cordova-ios@5, but XCode 14.3 is refusing to build it since it can only build deployment targets 11+

Likewise, the same erroris also reproduced for the app target:

```
/Users/norman/development/cordova/test/iostest/platforms/ios/HelloCordova.xcodeproj: warning: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 10.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'HelloCordova' from project 'HelloCordova')
```

This is further reinforced by their documentation for minimum SDK and requirements:

image

So based on this current behaviour there is a chance that XCode 15 will drop support for iOS 11. If that happens and if we aren't proactive on it, then we may need to make another major release to drop support for iOS 11 at that time. I don't think we have any particular deadlines to meet so we can wait to see what XCode 15 will support when it's beta comes out.

dpogue:
> I don't think we have any particular deadlines to meet so we can wait to see what XCode 15 will support when it's beta comes out.

WWDC is June 5th, so we should be able to find out fairly soon.

dpogue:
Testing with Xcode 15 reveals that it has iOS 12 as the minimum supported SDK, and produces a warning (for each project) if the target version is lower. But it builds and runs fine with no issues.

I don't think there's a way to get rid of that warning without bumping the version number, but there's also no specific need to bump it.

jcesarmobile:
One thing that was mentioned on the Android minSDK bump was the Chrome version. We bumped to minSDK 24 and was mentioned that the Chrome version included in the Android image was 52, so we would stick to Chrome's 52 features/support on the javascript side.
Any idea if iOS 11 WebView supports same features as Chrome 52?

In example the eslint changes from latest `@cordova/eslint-config` made the code of [screen orientation plugin](https://github.com/apache/cordova-plugin-screen-orientation/pull/112) not compatible with old android versions just because it uses let/const instead of var.
That doesn't seem an issue on iOS 11, but not sure if other rule changes could be not compatible with iOS 11.

breautek:
> One thing that was mentioned on the Android minSDK bump was the Chrome version. We bumped to minSDK 24 and was mentioned that the Chrome version included in the Android image was 52, so we would stick to Chrome's 52 features/support on the javascript side. Any idea if iOS 11 WebView supports same features as Chrome 52?
> 
> In example the eslint changes from latest `@cordova/eslint-config` made the code of [screen orientation plugin](https://github.com/apache/cordova-plugin-screen-orientation/pull/112) not compatible with old android versions just because it uses let/const instead of var. That doesn't seem an issue on iOS 11, but not sure if other rule changes could be not compatible with iOS 11.

It is good point,

ES6 (which includes stuff liek arrow functions, native promises, let/const, etc...) has good support since iOS 10, and since Chrome 51.

Specific ES6 features like `let` keyword may show slightly different compatibility, for example [caniuse](https://caniuse.com/let) says it's only supported in iOS 11 (iOS 10 had a buggy version of it), and chrome supported it in Chrome 49, or 41 if strict mode was enabled.

But overall, ES6 features should be pretty safe to use without any external tools, even if we keep iOS 11 as our minimum requirement.

https://caniuse.com/?search=ES6

breautek:
Sounds like we prefer to keep iOS 11 as the minimum target, so I'll close this.

Issue #1071, opened 2021-02-03
memory leak for hybrid apps with more than one CDVViewController


dbiaz:
# Bug Report

## Problem
retain cycle causes a memory leak when releasing CDVViewController in a hybrid app

### What is expected to happen?
CDVViewController and related objects should be released from memory

### What does actually happen?
memory not released 

## Information
CDVViewController holds a strong reference to CDVWebViewEngine
CDVWebViewEngine holds a strong reference to CDVURLSchemeHandler
CDVURLSchemeHandler holds a strong reference to CDVViewController

### Version information
cordova: cordova-ios 6.1.1 via cocoapods

## Checklist

- [x ] I searched for existing GitHub issues
- [ x] I updated all Cordova tooling to most recent version
- [x ] I included all the necessary information above

Issue #920, opened 2020-06-24
CDVViewController leaks  CDVWebViewWeakScriptMessageHandler when destroyed 


j-crosson:
# Bug Report

### What is expected to happen?

No leak

### What does actually happen?

leak

## Information
I’m working on a plugin that uses a subclassed CDVViewController in a modal view.  Works fine except for leaking CDVWebViewWeakScriptMessageHandler every time the view goes away.  I was able to stop the leak with the following change to CDVViewController.m.         

- (void)dealloc
{
    .
    .
    .

  WKWebView* wkWebView = (WKWebView*)self.webViewEngine.engineWebView;
  [wkWebView.configuration.userContentController removeScriptMessageHandlerForName:(@"cordova")]; //CDV_BRIDGE_NAME 

  self.webViewEngine = nil;
}

Not that this is the final fix, but it does point out the issue.

In my plugin subclass I use this:

    deinit {
        let wkWebView = webViewEngine.engineWebView as! WKWebView
        wkWebView.configuration.userContentController.removeScriptMessageHandler(forName: "cordova")
      }

The ordering is a bit different, but it seems to work.

This would also be an issue when embedding Cordova WebViews in a native app.

A related issue stems from the integration of the SplashScreen plugin into CDVViewController.  I’m showing a non-startup view and don’t want to see a startup image (a Webview embedded in a native app would have the same issue)  In my subclassed ViewController “viewDidLoad”, I replace “launchView” with a UIVIew with a selectable background color.  No flashes or startup images show. Works fine. My issue is that I have to use techniques I’d rather not use to expose “private” class members

### Environment, Platform, Device
iOS 13.5.1

### Version information
Cordova v9
Cordova-ios 6.1.0
Xcode 11.5
macOS 10.15.5

## Checklist

- [x ] I searched for existing GitHub issues
- [x ] I updated all Cordova tooling to most recent version
- [x ] I included all the necessary information above

Durzan666:
I use a subclass of CDVViewController with a NavigationController and observed that removing CDVViewController from NavigationController results in not calling dealloc because of strong references. I had to remove the strong references such as the MessageHandler in viewWillDisappear so dealloc got called.
In short: I copied the whole code of dealloc to viewWillDisappear including the above fix. That fixed the leak of CDVViewController in my case.

PR #1326, opened 2023-05-19
(ios) fix: memory leak when removing the CDVViewController

cbernier2:
### Platforms affected
iOS

### Motivation and Context
I am using Cordova inside a React Native application so I often need to destroy and re-create the CDVViewController.

### Description
Since I am using a custom scheme, it is leaking (the CDVViewController is never deallocated despite being remove from its parentViewController)

### Testing
I confirm that by changing strong to weak, the CDVViewController is deallocated as soon as it is removed from its parent view controller.

### Checklist

- [yes] I've run the tests to see all new and existing tests pass
- [no] I added automated test coverage as appropriate for this change
- [yes ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [no ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [no ] I've updated the documentation if necessary

### DIFF
--- a/CordovaLib/include/Cordova/CDVURLSchemeHandler.h
+++ b/CordovaLib/include/Cordova/CDVURLSchemeHandler.h
@@ -25,7 +25,7 @@
 
 @interface CDVURLSchemeHandler : NSObject <WKURLSchemeHandler>
 
-@property (nonatomic, strong) CDVViewController* viewController;
+@property (nonatomic, weak) CDVViewController* viewController;
 
 @property (nonatomic) CDVPlugin* schemePlugin;
 


dpogue:
This sounds like it solves https://github.com/apache/cordova-ios/issues/1071, right?

And then https://github.com/apache/cordova-ios/issues/920 is a different, unrelated memory leak?

cbernier2:
Oh right, it should fix those 2 issues, at least for me there's no more leak: both the webView and view controller get deallocated.

dpogue (approved):

PR #1340, opened 2023-06-07
chore!: Remove podspec support from framework tag

### LINKED ISSUE #812: remove type="podspec" from framework tag
It's deprecated and there is a `podspec` tag now, we should remove the `type="podspec"` from the `framework` tag and only read the `podspec` tag

dpogue:
Closes #812.

### Platforms affected
iOS

### Motivation and Context

`<framework type="podspec">` has been deprecated in favour of the `<podspec>` tag that allows much more control over pod dependencies.

### Description

Remove support for `<framework type="podspec">` and print an error if such framework tags are found.

### Testing

Removed unit tests around special case handling for podspec framework tags.

### Checklist


erisu (approved):
Since this is a major release, I think it is safe to remove the deprecated `framework` tag's attribute `podspec`.

The new method `<podspec>` was introduced in Cordova-iOS@5.0.0 (Feb 07, 2019).

Plugin developers had 4 years to migrate to the new tag.

PR #1374, opened 2023-09-28
fix: WASM MIME type error by specifying it in Info.plist template

### LINKED ISSUE #1370: wasm MIME type is not supported
# Bug Report

`application/wasm` should be returned as MIME type for wasm files from static file server in iOS.

wasm libraries not working due to this error:
`"Unexpected response MIME type. Expected 'application/wasm'"`

## Information

Here is an MR to Ionic that might be helpful:
https://github.com/ionic-team/capacitor/pull/5585/files

and another one in capacitor:
https://github.com/ionic-team/capacitor/pull/6675

### Command or Code

Just try to install and use a wasm library and run the app on simulator.

### Version information

Cordova: Cordova CLI 12.0.0, Cordova ios 7.0.0

## Checklist

michaelkamphausen:
### Platforms affected

iOS

### Motivation and Context

Fixes #1370 where WebAssembly cannot be loaded in iOS production build due to the error `"Unexpected response MIME type. Expected 'application/wasm'"`. The error causes a blank app screen and prevents any user interaction.

### Description

I added `UTExportedTypeDeclarations` to the Info.plist template file, specifying the expected MIME type for loading WASM files.

### Testing

I modified my project's Info.plist file by adding the `UTExportedTypeDeclarations` key and value as shown in the commit, built and ran my project on the device, and verified that the app loads and is displayed as expected.

The automated tests don't run on my macOS version as it does not match CordovaLibTests’s deployment target. I don't know how to get the tests running.

### Checklist

### DIFF
--- a/templates/project/__PROJECT_NAME__/__PROJECT_NAME__-Info.plist
+++ b/templates/project/__PROJECT_NAME__/__PROJECT_NAME__-Info.plist
@@ -45,5 +45,25 @@
 		<string>UIInterfaceOrientationPortraitUpsideDown</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>UTExportedTypeDeclarations</key>
+	<array>
+		<dict>
+			<key>UTTypeConformsTo</key>
+			<array>
+				<string>public.data</string>
+			</array>
+			<key>UTTypeDescription</key>
+			<string>WebAssembly</string>
+			<key>UTTypeIdentifier</key>
+			<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+			<key>UTTypeTagSpecification</key>
+			<dict>
+				<key>public.filename-extension</key>
+				<string>wasm</string>
+				<key>public.mime-type</key>
+				<string>application/wasm</string>
+			</dict>
+		</dict>
+	</array>
 </dict>
 </plist>
--- a/templates/project/__PROJECT_NAME__/__PROJECT_NAME__-Info.plist
+++ b/templates/project/__PROJECT_NAME__/__PROJECT_NAME__-Info.plist
@@ -45,7 +45,7 @@
 		<string>UIInterfaceOrientationPortraitUpsideDown</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
-	<key>UTExportedTypeDeclarations</key>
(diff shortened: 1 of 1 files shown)


templates/project/__PROJECT_NAME__/__PROJECT_NAME__-Info.plist
```diff
@@ -45,5 +45,25 @@
 		<string>UIInterfaceOrientationPortraitUpsideDown</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>UTExportedTypeDeclarations</key>
```
dpogue, line 48: I'm curious if this also works if you use `UTImportedTypeDeclarations` as the key

I don't think we want to declare all Cordova apps as owners of the WebAssembly file type by default, so `UTExportedTypeDeclarations` seems like it might be the wrong key to use. Even with that though, I think injecting this when needed using `config-file` in config.xml for your app is probably a better choice that unconditionally including this declaration in the template for all Cordova apps.

michaelkamphausen: I like your idea of using `UTImportedTypeDeclarations` key instead! I tested it and I can confirm that it works as well.

I understand there are differing views on whether loading WebAssembly code should be supported by default or only when needed.

I would argue for making this a default because I assume quite a few developers will be unaware of the problem until it suddenly occurs in their Testflight or release build. The cause is not easy to find, because the problem does not seem to occur in debug builds, which makes debugging difficult. This could be confusing and will cost time. At the same time, I don't see any downside how the `UTImportedTypeDeclarations` entry could cause problems.

Nevertheless, at least the documentation should point out the WASM loading error in release builds and explain how to fix it when needed using `config-file` in `config.xml`.

mohammadrafigh: Maybe it's optimistic, but it would be great if we could have a script to detect if the code is using WASM or not, then warn users to enable usage of WASM in config.xml. But for now this could be enough to somehow easily enable WASM support in config.xml. Putting all those config lines in config.xml is not a minimal way to enable WASM. Maybe a boolean preference like `EnableWASM` makes sense.

`<preference name="EnableWASM" value="true" />`

erisu (approved):
LGTM

PR #1400, opened 2024-02-21
fix!: Deprecate(/Remove) scrollView category extension on UIView

### LINKED ISSUE #1399: Please remove org_apache_cordova_UIView_Extension because it isn't necessary since iOS 8.
# Feature Request

Cordova contains `org_apache_cordova_UIView_Extension` in [`CDVPlugin.h`](https://github.com/apache/cordova-ios/blob/master/CordovaLib/include/Cordova/CDVPlugin.h#L27-L31):
```
@interface UIView (org_apache_cordova_UIView_Extension)

@property (nonatomic, weak) UIScrollView* scrollView;

@end
```

with an implementation in [`CDVPlugin.m`](https://github.com/apache/cordova-ios/blob/master/CordovaLib/Classes/Public/CDVPlugin.m#L26-L41):
```
@implementation UIView (org_apache_cordova_UIView_Extension)

@dynamic scrollView;

- (UIScrollView*)scrollView
{
    SEL scrollViewSelector = NSSelectorFromString(@"scrollView");

    if ([self respondsToSelector:scrollViewSelector]) {
        return ((id (*)(id, SEL))objc_msgSend)(self, scrollViewSelector);
    }

    return nil;
}

@end
```

It appears Cordova only uses the `scrollView` property on `WKWebView` instances, and [`WKWebView` has exposed a `scrollView` property since iOS8](https://developer.apple.com/documentation/webkit/wkwebview/1614784-scrollview), and [Cordova's iOS Platform Guide says iOS 11 is the minimum version required for devices to run Cordova](https://cordova.apache.org/docs/en/10.x/guide/platforms/ios/#requirements-and-support). Thus, this extension and artificial property is unnecessary.

## Motivation Behind Feature

Apps integrating Cordova with Swift codebases will not compile if Swift classes feature `scrollView` members in `UIView` subclasses with visibility less than `public`. Integrating Cordova would be easier without this extension.

Integrators can manually delete the `@interface` from the public header file to avoid compilation issues.

dpogue:
### Platforms affected
iOS

### Motivation and Context

As mentioned in #1399, this API was added as a compatibility shim at the time we supported both UIWebView and WKWebView, when WKWebView did not expose a public `scrollView` property. That property was made available in iOS 8, and this extension now risks causing problems due to being applied globally to all UIView subclasses.

Closes #1399.

### Description

Deprecate the category extension to expose a `scrollView` property on all UIView object instances, and remove it entirely from visibility to Swift.

### Testing

* Existing tests pass.
* Tested against all core Apache plugins.
* Manually tested that this resolves the issue in #1399 where a Swift UIView subclass defines a `scrollView` property

### Checklist

### DIFF
--- a/CordovaLib/include/Cordova/CDVPlugin.h
+++ b/CordovaLib/include/Cordova/CDVPlugin.h
@@ -27,11 +27,15 @@
 #import <Cordova/CDVViewController.h>
 #import <Cordova/CDVWebViewEngineProtocol.h>
 
+#ifndef __swift__
+// This global extension to the UIView class causes issues for Swift subclasses
+// of UIView with their own scrollView properties, so we're removing it from
+// the exposed Swift API and marking it as deprecated
+// TODO: Remove in Cordova 9
 @interface UIView (org_apache_cordova_UIView_Extension)
-
-@property (nonatomic, weak) UIScrollView* scrollView;
-
+@property (nonatomic, weak) UIScrollView* scrollView CDV_DEPRECATED(8, "Check for a scrollView property on the view object at runtime and invoke it dynamically.");
 @end
+#endif
 
 extern NSString* const CDVPageDidLoadNotification;
 extern NSString* const CDVPluginHandleOpenURLNotification;


jcesarmobile (approved):

dpogue:
This causes issues for the statusbar plugin which relies on this to access the scroll view without knowing the specific type of the web view implementation.

So putting this on hold for the moment.

dpogue:
Status Bar plugin is fixed, but this API is used in a number of 3rd party plugins as well... I feel like our only option is going to be deprecation instead of removal (which doesn't actually address the issue that was raised about this) 😞

jcesarmobile:
I don’t see what’s the problem with introducing a breaking change.
Changing the template to swift is also breaking to some plugins and it’s merged.

dpogue:
The best short term solution seems to be the following:

* Remove the extension from being visible to Swift code. This solves the problem raised in #1399
  * This breaks any Swift plugins that use this API, but there seem to only be a few and they are not maintained and don't have high download counts
* Mark the Objective C method as deprecated so that existing 3rd party plugins that call it will see warnings
* Prepare to remove this entirely in Cordova iOS 9 as per the deprecation

erisu (approved):
LGTM

I tested and confirmed that Obj-C based plugins with `self.webView.scrollView` will continue to work as expected and displays the deprecation warning.

Swift based plugins will need to update to continue working, which is expected.

```swift
var webViewScrollView: UIScrollView? {
    let scrollViewSelector = NSSelectorFromString("scrollView")
    
    if webView.responds(to: scrollViewSelector) {
        return webView.perform(scrollViewSelector)?.takeUnretainedValue() as? UIScrollView
    }
    
    return nil
}
```

dpogue:
The equivalent Objective C code would be (copied from the statusbar plugin):

```objc
- (UIScrollView *)webViewScrollView
{
    SEL scrollViewSelector = NSSelectorFromString(@"scrollView");

    if ([self.webView respondsToSelector:scrollViewSelector]) {
        return ((id (*)(id, SEL))objc_msgSend)(self.webView, scrollViewSelector);
    }

    return nil;
}
```

PR #1406, opened 2024-03-06
feat: add privacy-manifest config support

knaito-asial:
### Platforms affected
ios

### Motivation and Context
To support setting ios privacy manifest according to config.xml setting.

### Description
Introducing <privacy-manifest> tag in config.xml ios platform directive.

If user add following directive in config.xml (platform ios)
```
<privacy-manifest>
	<key>NSPrivacyTracking</key>
	<true/>
	<key>NSPrivacyCollectedDataTypes</key>
	<array/>
	<key>NSPrivacyAccessedAPITypes</key>
	<array/>
	<key>NSPrivacyTrackingDomains</key>
	<array/>
</privacy-manifest>
```
the ios project PrivacyManifest.xcprivacy is updated as config.xml.
If user does not set `<privacy-manifest>` tag in config.xml, the cordova template 
default PrivacyManifest.xcprivacy is used.

Note: This `<privacy-manifest>` tag overrides whole PrivacyManifest.xcprivacy file in project each time 
user execute cordova prepare.

### Testing
I locally checked this feature with sample project.
I locally exec `npm run coverage`.

### Checklist

### DIFF
--- a/lib/ConfigParseriOS.js
+++ b/lib/ConfigParseriOS.js
@@ -1,0 +1,45 @@
+
+/**
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+'use strict';
+
+const et = require('elementtree');
+const fs = require('fs-extra');
+const path = require('path');
+const ConfigParser = require('cordova-common').ConfigParser;
+
+class ConfigParseriOS extends ConfigParser {
+    constructor (path) {
+        super(path);
+    }
+
+    /** getPrivacyManifest */
+    getPrivacyManifest () {
+        const platform_manifest = this.doc.find(`./platform[@name="ios"]/privacy-manifest-ios`);
+        if (platform_manifest != null) {
+            return platform_manifest;
+        }
+        const manifest = this.doc.find('privacy-manifest-ios');
+        return manifest;
+    }
+}
+
+module.exports = ConfigParseriOS;
+
--- a/lib/ConfigParseriOS.js
+++ b/lib/ConfigParseriOS.js
@@ -20,19 +20,12 @@
 
 'use strict';
 
-const et = require('elementtree');
-const fs = require('fs-extra');
-const path = require('path');
 const ConfigParser = require('cordova-common').ConfigParser;
 
 class ConfigParseriOS extends ConfigParser {
-    constructor (path) {
-        super(path);
-    }
-
     /** getPrivacyManifest */
     getPrivacyManifest () {
-        const platform_manifest = this.doc.find(`./platform[@name="ios"]/privacy-manifest-ios`);
+        const platform_manifest = this.doc.find('./platform[@name="ios"]/privacy-manifest-ios');
         if (platform_manifest != null) {
             return platform_manifest;
         }
@@ -42,4 +35,3 @@ class ConfigParseriOS extends ConfigParser {
 }
 
 module.exports = ConfigParseriOS;
-
--- a/lib/PlatformConfigParser.js
+++ b/lib/PlatformConfigParser.js
@@ -17,8 +17,6 @@
     under the License.
 */
 
-'use strict';
-
 const ConfigParser = require('cordova-common').ConfigParser;
 
 class PlatformConfigParser extends ConfigParser {
--- a/lib/PlatformConfigParser.js
+++ b/lib/PlatformConfigParser.js
@@ -22,7 +22,7 @@
 
 const ConfigParser = require('cordova-common').ConfigParser;
 
-class ConfigParseriOS extends ConfigParser {
+class PlatformConfigParser extends ConfigParser {
     /** getPrivacyManifest */
     getPrivacyManifest () {
         const platform_manifest = this.doc.find('./platform[@name="ios"]/privacy-manifest-ios');
@@ -34,4 +34,4 @@ class ConfigParseriOS extends ConfigParser {
     }
 }
 
-module.exports = ConfigParseriOS;
+module.exports = PlatformConfigParser;
--- a/lib/PlatformConfigParser.js
+++ b/lib/PlatformConfigParser.js
@@ -1,4 +1,3 @@
-
 /**
     Licensed to the Apache Software Foundation (ASF) under one
     or more contributor license agreements.  See the NOTICE file
@@ -25,12 +24,11 @@ const ConfigParser = require('cordova-common').ConfigParser;
 class PlatformConfigParser extends ConfigParser {
     /** getPrivacyManifest */
     getPrivacyManifest () {
-        const platform_manifest = this.doc.find('./platform[@name="ios"]/privacy-manifest-ios');
+        const platform_manifest = this.doc.find('./platform[@name="ios"]/privacy-manifest');
         if (platform_manifest != null) {
             return platform_manifest;
         }
-        const manifest = this.doc.find('privacy-manifest-ios');
-        return manifest;
+        return null;
     }
 }
 
--- a/lib/PlatformConfigParser.js
+++ b/lib/PlatformConfigParser.js
@@ -22,13 +22,12 @@
 const ConfigParser = require('cordova-common').ConfigParser;
 
 class PlatformConfigParser extends ConfigParser {
-    /** getPrivacyManifest */
+    /**
+     * Returns the privacy manifest node, if available.
+     * Otherwise `null` is returned.
+     */
     getPrivacyManifest () {
-        const platform_manifest = this.doc.find('./platform[@name="ios"]/privacy-manifest');
-        if (platform_manifest != null) {
-            return platform_manifest;
-        }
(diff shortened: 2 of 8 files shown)


lib/ConfigParseriOS.js
erisu, line 1: Rename to `PlatformConfigParser.js`

lib/ConfigParseriOS.js
```diff
@@ -0,0 +1,37 @@
...
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+'use strict';
+
+const ConfigParser = require('cordova-common').ConfigParser;
+
+class ConfigParseriOS extends ConfigParser {
```
erisu, line 25:
```suggestion
class PlatformConfigParser extends ConfigParser {
```

IMO I think it would be better to make the class name generic. 
For example, if Android implement its own extended config parser class then it would follow identical naming conventions.

Almost as if we created and followed a platform template and would be easy to follow if we create a new platform in the future.

lib/prepare.js
```diff
@@ -35,6 +36,7 @@ const FileUpdater = require('cordova-common').FileUpdater;
 const projectFile = require('./projectFile');
 const Podfile = require('./Podfile').Podfile;
 const check_reqs = require('./check_reqs');
+const ConfigParseriOS = require('./ConfigParseriOS');
```
erisu, line 39: If class name changes, update here as well.

lib/prepare.js
```diff
@@ -43,9 +45,16 @@ const CDV_ANY_SIZE_CLASS = 'any';
 module.exports.prepare = function (cordovaProject, options) {
     const platformJson = PlatformJson.load(this.locations.root, 'ios');
     const munger = new PlatformMunger('ios', this.locations.root, platformJson, new PluginInfoProvider());
-
     this._config = updateConfigFile(cordovaProject.projectConfig, munger, this.locations);
 
+    const parser = new ConfigParseriOS(cordovaProject.projectConfig.path);
```
erisu, line 50: If class name changes, update here as well.

erisu (commented):
The suggested changes aim to establish a standard filename convention for use across platforms. This would facilitate the process for anyone intending to create a new platform by copying an existing one and making modifications without the need for renaming files.

lib/ConfigParseriOS.js
```diff
@@ -0,0 +1,37 @@
...
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+'use strict';
```
erisu, line 21: Might be able to remove this line. Most of our files does not declare `'use strict';`

lib/prepare.js
```diff
@@ -23,6 +23,7 @@ const fs = require('fs-extra');
 const path = require('path');
 const unorm = require('unorm');
 const plist = require('plist');
+const et = require('elementtree');
```
erisu, line 26: `cordova-ios` does not have `elementtree` package installed.

I know this works because the package exists in `cordova-common` but we should also install it to this package as a dependency.

```
npm install elementtree
```

lib/ConfigParseriOS.js
knaito-asial: Ok. I updated it

lib/ConfigParseriOS.js
knaito-asial: Yes, I updated it

lib/prepare.js
knaito-asial: Yes I updated it.

lib/prepare.js
knaito-asial: Thanks, I updated it.

lib/prepare.js
knaito-asial: Ok. I add elementree in package.json.

breautek:
> If user does not set <privacy-manifest-ios> tag in config.xml, the cordova template
default PrivacyManifest.xcprivacy is used.
>
>Note: This <privacy-manifest-ios> tag overrides whole PrivacyManifest.xcprivacy file in project each time
user execute cordova prepare.

How does this work with plugins? Can a plugin use `<privacy-manifest-ios>` within the `plugin.xml` or will the application's `config.xml` overwrite plugin entries?

I don't want to introduce scope creep but I just want to make sure this was thought about. I think it would be ideal for plugins to be able to declare their privacy entries and the results will get merged into one file for the application. Otherwise it will be a painful process for plugin authors to document and hope end users follows instructions to add the necessary privacy entries.

lib/PlatformConfigParser.js
```diff
@@ -0,0 +1,37 @@
...
+class PlatformConfigParser extends ConfigParser {
+    /** getPrivacyManifest */
+    getPrivacyManifest () {
+        const platform_manifest = this.doc.find('./platform[@name="ios"]/privacy-manifest-ios');
+        if (platform_manifest != null) {
+            return platform_manifest;
+        }
+        const manifest = this.doc.find('privacy-manifest-ios');
+        return manifest;
+    }
```
breautek, line 34: Personally I prefer a more structured approach, I don't think we need to support finding the privacy-manifest node in both root or the ios platform block. I think enforcing it to be inside the ios platform block is sufficient. Supporting both is twice the ongoing maintenance work.

Additionally, I think we can rename `privacy-manifest-ios` to `privacy-manifest`, we don't need the extra `-ios`. This creates for a more generic tag name to have consistency with other platforms (should they copy Apple and require a similar feature). Chopping of the `-ios` suffix is a way of hopefully future-proofing ourselves.

With both of these suggestions, the method will look more like:

```suggestion
    getPrivacyManifest () {
        return this.doc.find('./platform[@name="ios"]/privacy-manifest');
    }
```

Supporting just a platform / privacy-manifest way allows for more straight forward documentation and helps prevent easy mistakes like someone having a privacy-manifest in both locations, etc...

knaito-asial: Thanks. In this PR, manifest settings in plugins are not considered yet.
Next step, I want to develop the feature which is collecting each plugin's 
manifest if exists and generating application privacy manifest file according to 
the collected plugin's plugin.xml and application's config.xml.

knaito-asial: I think your suggestion about tag structure and name is great.
Please wait a moment.

erisu:
> How does this work with plugins? Can a plugin use `<privacy-manifest-ios>` within the `plugin.xml` or will the application's `config.xml` overwrite plugin entries?

The PR will solely focus on adding the minimum support for the `privacy-manifest` tag to the `config.xml` for app developers.

We aim to ensure that both Config.xml (for App Developers) and Plugin.xml (for Plugin Developers) support this tag and are completed and released before Apple's Privacy Manifest requirement takes effect. However, in the event that we are unable to complete the support for plugin.xml, at least the support for config.xml will be completed and merged with this PR.

> I don't want to introduce scope creep but I just want to make sure this was thought about. I think it would be ideal for plugins to be able to declare their privacy entries and the results will get merged into one file for the application. Otherwise it will be a painful process for plugin authors to document and hope end users follows instructions to add the necessary privacy entries.

As you mentioned, we're aiming to keep the scope specific for this PR. However, the ultimate goal aligns with your idea of gathering all `privacy-manifest` tag values and merging them into one file for the application.

lib/PlatformConfigParser.js
```diff
@@ -0,0 +1,37 @@
+
```
erisu, line 1: Can you remove the blank like at the top.
```suggestion
```

lib/PlatformConfigParser.js
```diff
@@ -0,0 +1,37 @@
...
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+'use strict';
```
erisu, line 21:
```suggestion
```

lib/PlatformConfigParser.js
```diff
@@ -0,0 +1,35 @@
...
+
+class PlatformConfigParser extends ConfigParser {
+    /** getPrivacyManifest */
+    getPrivacyManifest () {
+        const platform_manifest = this.doc.find('./platform[@name="ios"]/privacy-manifest');
+        if (platform_manifest != null) {
+            return platform_manifest;
+        }
+        return null;
+    }
```
breautek, line 32: I think the condition here is redundent.

We are:

1. creating a variable without making any use of it,
2. checking if it's not null and returning it
3. Otherwise returning null, which is the value of said variable if it reached this point.

I think we can simplify it just simply returning the result of `this.doc.find(...)`.

If we want to make it clear that null can be returned, we can perhaps add a JSDoc

```suggestion
    /**
     * Returns the privacy manifest node, if available.
     * Otherwise `null` is returned.
     */
    getPrivacyManifest () {
        return this.doc.find('./platform[@name="ios"]/privacy-manifest');
    }
```

knaito-asial: Thank you. I updated the code as you suggested.

dpogue (commented):
Formatting and code style comments aside, the overall implementation looks good to me :+1:

lib/PlatformConfigParser.js
breautek: thank you!

breautek (approved):
lgtm

BenLaKnet:
When this version could be used?

BenLaKnet:
I do not know how to test version 7.1.0-dev.
Please help.

BenLaKnet:
How to build this version to test ios manifest ?

BenLaKnet:
I do not see files generated, but when I uploaded a new version with 7.1.1.nightly-, I did not receive an email.

Is this possible?

shajz:
Hi, there is an issue with running `cordova prepare ios` that messes up the order of elements in the `privacy-manifest` element.

see: https://github.com/apache/cordova-ios/issues/1418

PR #1412, opened 2024-03-20
chore(ci): Fix ObjC testing with latest Xcode

dpogue:
### Platforms affected
iOS

### Motivation and Context

Tests would fail to run in Xcode 15 due to not having a iPhone 8 simulator with the latest runtime, and `CONFIGURATION_BUILD_DIR` was causing errors about Cordova.framework not being found and sometimes warnings about stale files.

### Description

* Move macOS CI to run on macOS 14 on Apple Silicon, using Xcode 15
* Set the tests to run with an iPhone 15 simulator
* Fix a warning about a missing cast to `AppDelegate*`
* Set `-derivedDataPath` rather than `CONFIGURATION_BUILD_DIR` so that Apple doesn't cache stale data about previous runs

### Testing

Ran in GitHub CI and locally with latest Xcode 15

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         node-version: [16.x, 18.x, 20.x]
-        os: [macos-latest]
+        os: [macos-14]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


erisu (approved):
👍 LGTM

PR #1413, opened 2024-03-24
chore(deps): Modernize some dependencies

dpogue:
### Platforms affected
iOS

### Motivation and Context

This allows us to drop the `fs-extra` and `unorm` library by replacing their use with built-in NodeJS standard library calls.

### Description

* Replace `unorm` module with `String.prototype.normalize`
* Replace `fs-extra` usage with equivalent NodeJS `fs` calls
* Upgrade `which` to latest version
* Use `path.join` in a bunch of spots where we're constructing paths
* Remove the (unused) `nopt` dependency

### Testing

- [x] Existing unit tests pass
- [x] Created and built iOS test project

### Checklist

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -24,7 +24,3 @@ overrides:
 
   - files: [cordova-js-src/**/*.js]
     extends: '@cordova/eslint-config/browser'
-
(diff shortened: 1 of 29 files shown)


erisu (approved):
LGTM

Issue #1367, opened 2023-08-29
drop xcode 11-12-13 support


jcesarmobile:
This is more a docs thing, but we claim to support Xcode 11 and newer.
Apple requires Xcode 14.1 or newer since April 25, 2023.

In the 8.0.0 release announcement we should announce that we drop support for older, and also update https://cordova.apache.org/docs/en/dev/guide/platforms/ios/index.html page to specify the same version.

Depending on when the 8.0.0 release happens we might also want to drop Xcode 14 since Xcode 15 will probably be required at some point of 2024.

FelipeAzambuja:
![image](https://github.com/apache/cordova-ios/assets/1438188/56526ab9-da51-410a-9d30-a8b259bd2e65)

Today my development environment that my machine allows is this
macos 10.15
and xcode: 12.0
When I sign up I can use the most current machines, but for development and testing this is my reality.
How could I continue working?
I am using docker osx project

dpogue:
@FelipeAzambuja Cordova-iOS will work fine on your current setup for development and testing, but you will not be able to build and submit apps to the App Store from that environment because Apple requires using Xcode 14

jcesarmobile:
There is no guarantee that it will work after cordova-ios 8 is released, because once we drop support of those obsolete Xcode versions, some plugins that have code to support older Xcode versions might remove that code. And also cordova-ios itself might remove some code that I added recently to allow compiling on Xcode 11.

You'll have to stick with cordova-ios 7.x or buy a new dev computer.

PR #1425, opened 2024-04-23
chore!: Bump minimum Xcode version to 15

jcesarmobile:
Closes https://github.com/apache/cordova-ios/issues/1367

~we could even drop Xcode 14 and go to 15 directly since Apple is going to require Xcode 15 at some point this month~


dpogue:
I would say go with 15, since that has support for all the Privacy Manifest stuff

breautek (approved):

PR #1426, opened 2024-04-23
chore: Update CordovaLib.xcodeproj fixture

jcesarmobile:
While searching for some files I noticed this fixtures file that is very out of date with the original file, so I’ve replaced the old file with the real file.


dpogue (approved):
LGTM

PR #1430, opened 2024-04-27
WIP - feat: Add support for SPM plugins

jcesarmobile:
Remove CordovaLib from the template and instead add a local "empty" swift package, in which `cordova-plugin-add` will add the plugins as dependencies if they are swift package manager compatible.

Plugin example to test with: https://github.com/apache/cordova-plugin-device/pull/196

Using this approach plugins that want to add Swift Package Manager dependencies will have to be converted to be Swift Package Manager compatible and add dependences as they would to a regular Swift Package instead of relying in a new tag in `plugin.xml`


dpogue:
There's a bit of Xcode mangling the project file (stripping licence comments, removing quotes around `__PROJECT_NAME__` and `__PROJECT_ID__`, etc.) that will probably need to be fixed by hand.

The idea of a SwiftPM subproject is a good one and will make the process of adding/removing plugins significantly easier than trying to deal with the Xcode project directly. That said, would it make sense for CordovaLib to be referenced directly from the Xcode project rather than through the SwiftPM library?

One worry is that we won't be able to ship with CordovaLib pointing at `master`, we'll need to update that to point at a tag (which is annoying because that needs to happen as part of creating the tag to which it will point)

templates/project/__PROJECT_NAME__.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,15 @@
+{
```
dpogue, line 1: We probably want to gitignore this file

jcesarmobile:
Keeping CordovaLib was a good call as it allows plugins to not have a dependency to cordova-ios, I tried several things like using an environment variable for setting the cordova-ios versions to use based on the installed version, and worked when running the app from the CLI, but was not working if running from Xcode since the variable is unset when the node process stops.
A similar option was to set the version on the `Package.swift` on platfor/plugin add, but as you mentioned, the tag has to exists, which is a problem for dev versions.

Also tried using cordova-ios from node_modules directly and works fine for the `CordovaPluginsSPM` project, but for plugins it would require to add cordova-ios as a dev dependency on plugins and then adjusting paths once installed in an app since the path to cordova-ios would be different when used from an app and from the plugin directory itself.

While I would have loved using the local version so we have the privacy manifest linked and on plugins we could run `xcodebuild` command to see if the plugin code builds without adding it to an app, keeping `CordovaLib` is the easier and less conflicting option.

dpogue:
hmmm, I'd been hoping that we would be able to reference CordovaLib using SwiftPM, but directly in the app's Xcode project rather than through the CordovaPluginsSPM project's Package.swift... but if I'm reading your comment correctly that still causes issues for plugins? 😞

jcesarmobile:
I didn't try that, but still, it would be a complex thing to do since if you add it to the template, when the template is turned into an app the path to cordova-ios's Package.swift will be different, so should be adjusted in the xcodeproj file.

lib/Api.js
```diff
@@ -279,6 +279,11 @@ class Api {
                     return this.addPodSpecs(plugin, podSpecs, installOptions);
                 }
             })
+            .then(() => {
+                if (plugin != null && plugin.dir && fs.existsSync(path.join(plugin.dir, 'Package.swift'))) {
```
dpogue, line 283: Looking at this PR again, and the one thing I don't really love is the "magic" behaviour based on the existence of a Package.swift file in the plugin folder. It feels safer to rely on an explicit declaration in the plugin.xml file.

We've already overloaded the `<framework>` tag to do a billion things, so adding this wouldn't be totally unreasonable. i.e.,
```xml
<framework src="Package.swift" type="spm" />
```

Alternatively, the `lib-file` tag exists on Android but you could maybe stretch the definition to encompass Swift packages.

Using `source-file` is an option, but feels like that goes back to "magic" behaviour based on file names.

jcesarmobile:
Anyway, I’m closing the PR, it was painful enough to address the requested changes on the project.pbxproj file, now it have conflicts that will be impossible to fix

Issue #1429, opened 2024-04-26
CDVWebViewUIDelegate should take a CDVViewController


dpogue:
# Feature Request

## Motivation Behind Feature

`CDVWebViewUIDelegate` handles a bunch of interactions with the UI and view controllers, and currently doesn't have a pointer to the `CDVViewController` that owns its associated `WKWebView`.

## Feature Description

Provide an initializer for `CDVWebViewUIDelegate` that takes the `CDVViewController` and holds a weak reference to it.

## Alternatives or Workarounds

We make some unsafe assumptions like the root view controller being safe to access, which has led to issues such as https://github.com/apache/cordova-ios/issues/1120. By having a reference to the `CDVViewController` we could ensure that any popups are correctly triggered by that view controller.

PR #1437, opened 2024-05-28
fix: alerts could freeze the application

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes GH-1120.
Closes GH-1121.
Closes GH-1429.

### Description

In cases where the system was presenting something on the screen (such as a permission prompt, or an undo dialog) or in cases where the `CDVViewController` was not the root view controller (such as in a tabbed navigator), trying to show any dialogs triggered from the WebView would attempt to create an alert on the root view controller and get the view hierarchy into a bad state.

Now, we only try to display alerts on the `CDVViewController` that owns the WebView. This means that alerts can potentially end up behind other system dialogs, but the view controller no longer freezes.

(As an aside, having access to the `CDVViewController` here will be helpful for some future things too...)

### Testing

Tested manually by triggering an alert while a permission prompt was being displayed and confirmed that the app continued working as expected.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -192,7 +192,8 @@
         self.CDV_ASSETS_URL = [NSString stringWithFormat:@"%@://%@", scheme, hostname];
     }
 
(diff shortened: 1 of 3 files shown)


erisu (approved):

dpogue:
I did some testing of alert dialogs with this PR while the camera and file pickers were open, and it seemed to work as expected with no freezes.

erisu (approved):
LGTM

I retested the new changes, and they appear to be working as expected.

For example:

- If the camera is open and an alert dialog appears, the user must close the dialog before interacting with or closing the camera capture window.
- When the camera capture window is closed, the app does not freeze and continues to operate as expected.
- Calling multiple system dialogs will not render stack; instead, each dialog will display sequentially, after the previous one is closed.

PR #1451, opened 2024-08-01
(ios) Adds a routine to flush WkWebView diskcache when the bundle ver…

BBlashko:
…ion changes

### Platforms affected
 - ios

### Motivation and Context

This Pull Request adds a routine to the Boot of cordova-ios so that we can flush diskcache in the WkWebView. There is a problem with iOS where the diskcache aka .js, html, and css etc files are cached between .ipa installs. 

There is no issue for this at this time, I was discussing it with Norman Breau in the Slack Space

### Description

In the AppDelegate.m file I have added a routine to clear the `WKWebsiteDataTypeDiskCache` for the domain from the `Bundle ID`, and `localhost`. This cache needs to be flushed because it latches on to previous `js, html, css, etc` files from a previous `.ipa` installation.  These cached files include the `cordova.js` and `cordova_plugins.js` files. I know this because I added hacks into these files, but they were not registered on the target production application.

The main problem this addresses is removals of Cordova Plugins. Where the cached cordova_plugins.js thinks that it needs to load a removed plugin, but since the plugins www directory was removed it doesn't exist. 

An example error message is as follows:
```
Module branch-cordova-sdk.Branch does not exist.
```

Now I cannot confirm at this time, but it does indicate that with this cached files, that any additions to cordova_plugins.js may not be loaded at runtime. 

I want to mention that this problem is extremely hard to reproduce in local environments. It _needs_ to be on device, and it needs to be an update to an already installed application in production. Eg. I have 3.72.0 on my device, a new 3.73.0 version is released with removed/updated plugins, then this case _can_ happen it is not guaranteed to happen for all users.

### Testing

Our QA engineer had an instance of the application stuck in this state. The Error: 
```
Module branch-cordova-sdk.Branch does not exist.
```

This error was thrown everytime they booted the application, where the UI was stuck on the Splashscreen. 

Updating to the latest .ipa (with these changes in the PR) and booting, the user was able to get "unstuck".  However, it's not a _perfect_ solution because while the cache did get flushed, it didn't immediately repopulate it, and the user had to reboot the app. On second boot after this fix was introduced the user was able to boot the application successfully.

### Checklist

- [*] I've run the tests to see all new and existing tests pass
- [*] I added automated test coverage as appropriate for this change
- [*] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [*] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [*] I've updated the documentation if necessary

### DIFF
--- a/templates/project/__PROJECT_NAME__.xcodeproj/project.pbxproj
+++ b/templates/project/__PROJECT_NAME__.xcodeproj/project.pbxproj
@@ -35,6 +35,7 @@
 		4E7CA2B7272ABB0D00177EF9 /* www in Copy Staging Resources */ = {isa = PBXBuildFile; fileRef = 301BF56E109A69640062928A /* www */; };
 		6AFF5BF91D6E424B00AB3073 /* CDVLaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 6AFF5BF81D6E424B00AB3073 /* CDVLaunchScreen.storyboard */; };
 		90B630EF2AECBBD0009EF368 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 90B630EE2AECBBD0009EF368 /* PrivacyInfo.xcprivacy */; };
+		D0A4184E2C599F4800A109D2 /* WkWebViewCacheClear.m in Sources */ = {isa = PBXBuildFile; fileRef = D0A4184D2C599F4800A109D2 /* WkWebViewCacheClear.m */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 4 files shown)


dpogue:
I wonder if an easier/better fix would be to prevent the local web content from getting cached at all (since it's already getting loaded from the device itself)?

BBlashko:
> I wonder if an easier/better fix would be to prevent the local web content from getting cached at all (since it's already getting loaded from the device itself)?

Ya that would make sense to me 🤔 

Could you point me in a bit of a direction as to where `cordova-ios` initializing the `WKWebView`? so i can see if I can put some no cache configuration on it? I'm a bit inexperienced with Native iOS.  I understand the `AppDelegate`, and that initializes the `MainViewController`, but where does the framework initialize the `WebView`?

dpogue:
I _think_ what we'd need to do is explicitly set a cache policy on the `NSURLRequest` that is created when we load a URL into the `WKWebView`: https://github.com/apache/cordova-ios/blob/3a426fd37d478fb8ee93d6de613e55888c4fa02b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m#L334

By default, that's going to use "default protocol cache behaviour", which is documented here: https://developer.apple.com/documentation/foundation/nsurlrequestcachepolicy/nsurlrequestuseprotocolcachepolicy?language=objc

We might want to set the policy to `NSURLRequestReloadIgnoringLocalCacheData` and test if that solves the problem.

BBlashko:
I created a new pull request with the updated. Since it was a completely different approach I figured a new branch and new pull request was the best approach. 
 
https://github.com/apache/cordova-ios/pull/1471

dpogue:
Closing to close this PR in favour of getting #1471 merged in.

PR #1455, opened 2024-08-14
refactor!: Easy deprecation cleanups

### LINKED ISSUE #1428: CDVPlugin's viewController should just be a CDVViewController
# Feature Request

## Motivation Behind Feature

`CDVPlugin` has a `viewController` property that is currently typed as `UIViewController*`. However, it is only ever set from within  `CDVViewController` and is always set to a `CDVViewController*` instance.

## Feature Description

We should just update the `CDVPlugin` API so that `viewController` is typed as `CDVViewController*` rather than `UIViewController*`.

Since `CDVViewController` is a subclass of `UIViewController` this shouldn't break any plugins, but it is an API change and should be handled in the next major.

Currently the standard practice is to just cast `(CDVViewController*)self.viewController` everywhere, which works fine but is kinda annoying.

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes #1428.

### Description

* Added deprecation annotations to things we want to discourage people from using (some of these can be removed in Cordova iOS 9)
* Remove some conditionals for iOS versions pre-11 and dead code
* Change CDVPlugin's `viewController` from a `UIViewController*` to a `CDVViewController*`
* Move a webview engine-related delegate to a header that makes more sense

### Testing

Unit tests pass.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVLaunchScreen/CDVLaunchScreen.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVLaunchScreen/CDVLaunchScreen.m
@@ -24,16 +24,12 @@
 
 - (void)show:(CDVInvokedUrlCommand*)command
 {
-    if ([self.viewController isKindOfClass:[CDVViewController class]]) {
-        [(CDVViewController*)self.viewController showLaunchScreen:YES];
-    }
+    [self.viewController showLaunchScreen:YES];
 }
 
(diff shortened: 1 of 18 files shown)


erisu (approved):
LGTM

Tested with:
- Xcode 16 (beta 5) & iOS 18 (beta 5)
- Cordova default application and confirmed that the `Device is Ready` is displayed.
- Mobile spec to run against all of Apache Cordova's plugins.

PR #1456, opened 2024-08-14
fix: Fix memory leak from user script handler

### LINKED ISSUE #920: CDVViewController leaks  CDVWebViewWeakScriptMessageHandler when destroyed

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes #920.

### Description

Ensure that we remove the WKWebView script message handler when the web view plugin is disposed. Otherwise it stays in memory holding references to stuff.

### Testing

Tests pass

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -273,6 +273,14 @@
     NSLog(@"Using WKWebView");
 }
 
+- (void)dispose
+{
+    WKWebView* wkWebView = (WKWebView*)_engineWebView;
+    [wkWebView.configuration.userContentController removeScriptMessageHandlerForName:CDV_BRIDGE_NAME];
+
+    [super dispose];
+}
+
 - (void) onAppWillEnterForeground:(NSNotification*)notification {
     if ([self shouldReloadWebView]) {
         NSLog(@"%@", @"CDVWebViewEngine reloading!");


erisu (approved):
LGTM

Issue #904, opened 2020-06-17
MainViewController.m omit empty `viewDidLoad` to avoid warnings


msmtamburro:
# Bug Report

## Problem

Thanks for removing the deprecated viewDidUnload!  If you also remove some of the other empty methods (e.g., `viewDidLoad`) this gives developers a way to easily create an extension on MainViewController.m to implement those methods without warning.

### What is expected to happen?

A developer should be able to implement `viewDidLoad` in MainViewController.m

### What does actually happen?

Using the existing stub creates either architectural issues (e.g., changing a file that will get overwritten) or warnings (as mentioned the extension approach above).

## Information

You might just remove all empty methods, and replace them with a comment like:

The best way to customize `viewDidLoad` is to make a Swift Extension of MainViewController and implement the method there.  You can add your extension to your project by including it in a plugin.

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

dpogue:
We actually use `viewDidLoad` now in the MainViewController to control the visibility of the splashscreen image (since doing it in `CDVViewController` causes it to appear unexpectedly when embedding and using `CDVViewController` as one view in a larger app) 😞

Issue #674, opened 2019-09-25
Any ideas to supports multiple windows?


OctHex2016:
Any ideas to supports multiple windows?
When I click to supports multiple windows, it become black screen. As I know, we need to change to CDVAppDelegate to support UIResponder with UIWindowSceneDelegate, however, I can't make it success. Thanks!

dpogue:
I've thought about this a bit over the past few weeks, and one challenge is that UISceneDelegate is only available in iOS 13, and cordova-ios still needs to support versions as far back as (currently) iOS 10.  It can be done in a conditional way, but that runs the risk of popular plugins being incompatible due to changes in behaviour of methods on the UIAppDelegate vs the UISceneDelegate.

Definitely something we need to examine in more detail as part of the next major version.

OctHex2016:
I just make my app can supports multiple windows. 
But still need to test. Mainly need to add the SceneDelegate and CDVAppDelegate.

PR #1457, opened 2024-08-14
refactor(template)!: Update app template to Swift & storyboards

dpogue:
### To Do

- [x] Confirm the template Xcode project has all the right files (I think it might be missing the Staging folder, and maybe one of the build.xcconfig files)

### Platforms affected
iOS

### Motivation and Context

Closes https://github.com/apache/cordova-ios/issues/674.
Closes https://github.com/apache/cordova-ios/issues/904.

### Description

Replace the existing template project with one that uses modern Swift classes, storyboards, and customizes the `CDVViewController` using storyboard properties rather than relying on specifically-named colour variables being globally available.

This is one step that should make it easier and much more pleasant for people wanting to embed `CDVViewController` instances within their existing iOS apps.

### Testing

Built to a simulator.
All existing tests pass.

### Checklist


dpogue:
Current feedback is that some plugins are (unsafely!) trying to add category extensions to the template's `AppDelegate` class. This was never really a safe thing to do, since people can consume CordovaLib as a framework in their own projects with whatever class names they want.

The correct way for plugins to add behaviour to the app delegate is by extensions of the `CDVAppDelegate` base class.

The same issue probably exists with the view controller class as well.

---

For the sake of compatibility though, I'm looking at wrapping the `AppDelegate` and `ViewController` Swift classes in `@objc` and create empty header files for them with deprecation warnings.

```c
// AppDelegate.h
#import <Cordova/Cordova.h>

#warning It is unsafe to rely on the AppDelegate class as an extension point. \
         Update your code to extend CDVAppDelegate instead -- \
         This code will stop working in Cordova iOS 9!

@interface AppDelegate : CDVAppDelegate
@end
```

```swift
// AppDelegate.swift
import Cordova;

@objc class AppDelegate : CDVAppDelegate
{
}
```

The downside of this method is that we're stuck with these compatibility hacks for at least one major version.

We also have to alias the `ViewController` class as `MainViewController` in ObjC for compatibility.

erisu (approved):
~~Though the PR is currently in `Draft`,~~ I think it LGTM.

I tested:

- Building the Cordova sample app with Xcode 16 Beta 5.
- Building and running against all Apache Cordova default plugins.
- Running on iOS 18 beta 5 Simulator
- Tested against a plugin that uses CocoaPods

When testing a plugin that uses CocoaPods, `warning: Stale file` may print out in the build output. These are only warnings and the build still succeeded. These type of warnings are not uncommon and believe to be an issue with Xcode.

PR #1458, opened 2024-08-15
refactor!: Add CDVSettingsDictionary class

dpogue:
### Platforms affected
iOS

### Motivation and Context

This will replace the extension category on NSDictionary, which has always felt a little bit hacky and has caused issues in weird circumstances due to requiring special build flags to work properly.

### Description

Add a new `CDVSettingsDictionary` class that is used for accessing preferences.

Yes, subclassing `NSDictionary` is generally considered a bad idea because the internal implementation details are weird and cause problems. However, in this case we're essentially wrapping a real `NSMutableDictionary` and proxying everything to it, so this _should_ work the way we'd expect it to.

### Testing

Added new unit test providing coverage for all the CDVSettingsDictionary methods.

Added a test file to check Swift behaviour as well, but adding it to the test suite project doesn't work (weird Xcode swift errors). I'd like to refactor the test project setup as well, but I'll deal with that after getting through the main CordovaLib project...

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/CDVCommandDelegateImpl.m
+++ b/CordovaLib/Classes/Private/CDVCommandDelegateImpl.m
@@ -173,7 +173,7 @@
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), block);
 }
 
-- (NSDictionary*)settings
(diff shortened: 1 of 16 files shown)


erisu (approved):
Code LGTM

PR #1459, opened 2024-08-15
chore: Clean up ancient plist-to-config.xml cruft

### LINKED ISSUE #1414: [Next Major] Delete bin/cordova_plist_to_config_xml script
This old Python script was to help the transition from Cordova 1.x to newer Cordova versions using config.xml. I don't know what version of Python it even targets, but let's assume it's not recent. Cordova 1.x was over a decade ago and there's not really a feasible upgrade path from there to here simply by running a script.

I doubt anyone will notice or care if this goes missing, but to be safe let's do it as part of a major version bump.

There's also a [block of code in CDVViewController](https://github.com/apache/cordova-ios/blob/f7e29536474378249ba6278ce88aa853fef1f923/CordovaLib/Classes/Public/CDVViewController.m#L128-L132) that, if it fails to find config.xml, encourages people to run this script. We should remove that too.

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes #1414.

This was for migrating from PhoneGap 1.x to Cordova, and is definitely not compatible with any modern version of Python.

### Description

Remove old Python script, and a reference in an error message to that script.

### Testing

None?


erisu (approved):
LGTM

breautek (approved):

PR #1463, opened 2024-08-15
refactor!: Move CDVAllowList out of public API

dpogue:
### Platforms affected
iOS

### Motivation and Context

This isn't really accessible outside of implementation details of the internal CDVIntentAndNavigationFilter plugin, so it can be a project-level internal header that isn't exposed publicly.

### Description

Moved CDVAllowList from a public header to a project-level header.

### Testing

Existing tests pass.

### Checklist


erisu (approved):
The changes LGTM

PR #1466, opened 2024-08-18
chore!: Bump NodeJS versions in CI & update deps

dpogue:
### Platforms affected
iOS

### Motivation and Context

Node 16 is EOL and Node 22 will be LTS in October.

### Description

* Dropped Node 16 from CI testing
* Added Node 22 to CI testing
* Updated `nyc`, `jasmine`, and `rewire` deps to latest versions

Technically all the breaking dependency updates were only in devDependencies, so the platform should still work for end users on Node 16, but we're not testing that in CI anymore so we will have no way of making any such guarantees.

### Testing

All existing tests pass.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


erisu (approved):
LGTM

Issue #1365, opened 2023-05-02
Plugin creates an empty Podfile for some reason (problem for build services)


AshleyScirra:
# Bug Report

## Problem

For some reason, including `cordova-plugin-inappbrowser` ends up creating an empty Podfile in the Xcode project. If you then run `pod install` on the project, it ends up broken - you get a weird error:

`The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.`

I've searched the web, tried deleting files and clean rebuilding, etc. etc. I could not find any workaround. It seems to just leave a permanently broken Xcode project.

This might sound weird - don't run `pod install`, right? But this caused us serious problems. We run a build service that builds people's Cordova projects for them. We'd like our build service to run `pod install`, but we can't, as it might break the project if it uses `cordova-plugin-inappbrowser`. So ideally  `cordova-plugin-inappbrowser` would not create an empty Podfile.

### What is expected to happen?

 `cordova-plugin-inappbrowser` should not create an empty Podfile, as it does not appear to actually use any pods.

### What does actually happen?

 `cordova-plugin-inappbrowser` creates an empty Podfile that looks like this:

```
# DO NOT MODIFY -- auto-generated by Apache Cordova

platform :ios, '11.0'

target 'HelloWorld' do
	project 'HelloWorld.xcodeproj'

end
```

If you run `pod install` with that, you get weird errors and the Xcode project is basically broken. However in some cases we want to automatically run `pod install`, and this complicates that.

### Command or Code

The smallest repro is to run these commands:

`cordova create hello com.example.hello HelloWord`
`cd hello`

Now edit config.xml and add the line:

`<plugin name="cordova-plugin-inappbrowser"/>`

Then run:
`cordova platform add ios`
`cordova prepare`

Now look in the subfolder `platforms\ios`. There's an empty Podfile. If `cordova-plugin-inappbrowser` is not included, then no Podfile is created.

I'm not sure what's going on here but would appreciate any insight in to why this file is created, or if any change could be made to avoid creating it. Thanks for any help!

### Environment, Platform, Device
iOS, Cordova v11.1.0, cordova-plugin-inappbrowser@5.0.0

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

jcesarmobile:
Moved the issue to cordova-ios since this behavior happens with any plugin, not just with InAppBrowser.

[...]

PR #1467, opened 2024-08-18
fix(Podfile): Avoid creating an empty Podfile on plugin install

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes GH-1365.

### Description

Calling `new Podfile(...)` will create an empty Podfile in the project if it does not exist. If we don't have any pods to add for a given plugin, we can skip doing anything at all with the Podfile and avoid needlessly creating it.

Note: I moved an `if` block to do an early return, which de-indented the majority of the code in that function. This PR is easier to review with whitespace changes disabled.

### Testing

Tested manually by creating a new project and installing the `cordova-plugin-inappbrowser` and confirmed that no Podfile was present in the `platforms/ios` directory.

### Checklist

### DIFF
--- a/lib/Api.js
+++ b/lib/Api.js
@@ -338,6 +338,10 @@ class Api {
      * @return  {Promise}  Return a promise
      */
     addPodSpecs (plugin, podSpecs, installOptions) {
+        if (!podSpecs.length) {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


erisu (approved):
Code LGTM

PR #1472, opened 2024-08-21
refactor: Modernize some ObjC code

dpogue:
### Platforms affected
ios

### Motivation and Context

Some of the code in CordovaLib dates back to the very early dates of cordova-ios and an era when iOS 4 introduced breaking changes and Objective C was having new language features introduced (yes, Cordova actually predates the transition to ObjC ARC 💀).

Nowadays there are some conventions that are pretty standard (like `instancetype` for initializer methods) that we should aim to adopt for consistency.

Also, refactored `CDVConfigParser` to avoid needing a round-trip through weird `CDVViewController` methods to do anything useful. I can split this commit out into its own PR if that makes it easier for review.

### Description

* Various small ObjC cleanups
* Add some static methods to `CDVConfigParser` to facilitate parsing config.xml

### Testing

All existing unit tests pass.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/CDVCommandDelegateImpl.h
+++ b/CordovaLib/Classes/Private/CDVCommandDelegateImpl.h
@@ -17,7 +17,7 @@
  under the License.
  */
 
-@import UIKit;
+#import <UIKit/UIKit.h>
 #import <Cordova/CDVCommandDelegate.h>
 
 @class CDVViewController;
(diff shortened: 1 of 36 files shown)


erisu (approved):
LGTM

Tested with Xcode 16 beta 5, iOS 18 beta, Apache Cordova plugins and one plugin that uses CocoaPods.

PR #1474, opened 2024-08-23
refactor: Clean up properties/ivars in CDVViewController

dpogue:
### Platforms affected
iOS

### Motivation and Context

A bunch of properties on CDVViewController were publicly exposing internal state that should be private to the instance. We can't completely solve that problem right now, but we can make the API boundary more clearly delineated.

### Description

Clean up properties/ivars in CDVViewController

Rename wwwFolderName to webContentFolderName so that it gets a useful label in Xcode's Interface Builder for people embedding CDVViewController inside their own apps (IB showed it as "Www Folder Name" and there's no way to alias that, so now it shows up as "Web Content Folder Name")

### Testing

Existing tests pass.

### Checklist


erisu (approved):
LGTM

Tested with Xcode 16 beta 6, iOS 18 beta 7, Apache Cordova plugins and a plugin with CocoaPods.

PR #1477, opened 2024-08-24
refactor: Move CDVURLSchemeHandler to private API

dpogue:
### Platforms affected
iOS

### Motivation and Context

This is only used by the WKWebView plugin, and there's no way to do anything meaningful with it as public API, so it probably shouldn't be public API.

### Description

Moved `CDVURLSchemeHandler` files out of the public API.

### Testing

Existing tests pass.

### Checklist


erisu (approved):
Tested with the Cordova sample app and an app with all Apache Cordova Plugins loaded..

The test also included setting the custom URL scheme.

PR #1482, opened 2024-08-28
fix: CDV.h wasn't bringing in Cordova.h consistently

dpogue:
### Platforms affected
iOS

### Motivation and Context

Plugins that included only `<Cordova/CDV.h>` were seeing issues around extending `CDVPlugin`.

### Description

Make `CDV.h` consistently import `Cordova.h` to define the full module.

The fact that there's a circular dependency between the headers isn't actually a problem in Objective-C because it uses `#import` rather than `#include` and the compiler ensures each header is only included once.

Somehow that was also causing it to only evaluate the header once and not differentiate imports where the macro was defined vs not defined.

### Testing

Compiled a test app whose plugins used `#import <Cordova/CDV.h>` and were broken before this change.

### Checklist


erisu (approved):
LGTM - I have tested and confirmed that it resolved the issue.

Issue #959, opened 2020-07-28
WKWebView cancelled request(NSURLErrorCancelled -999) leads to errorUrl page instead of the second request


arunn:
# Bug Report

## Problem

A request cancelled due to another request ends up showing the error page instead of response of the second request.

### What is expected to happen?

The response of the second request should be shown.

### What does actually happen?

The errorUrl page is shown.

## Information

The error is similar to https://github.com/apache/cordova-plugin-wkwebview-engine/issues/83 . We have two requests being run one after the other. The first request is cancelled once the second request is made. Instead of waiting for the response of the second request, the didFailLoadWithError shows the errorUrl page.

### Environment, Platform, Device

cordova-ios, iPhone 8(iOS 13.6)

### Version information

cordova(9.0.0)
cordova-ios(5.1.1)
cordova-plugin-wkwebviewengine(1.2.1)

Filing it here since the code has been integrated as of cordova-ios 6.0.0 and the issue doesn't seem to be fixed in the code.

## Checklist

- [x] I searched for existing GitHub issues
- [] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

tdfeslo:
Hi. Is there any news about this problem?
The solution proposed [here](https://github.com/apache/cordova-plugin-wkwebview-engine/pull/84/commits/bcb4e90f94516737a9348092e166e0d0ffa800b4) solves the problem but may be introduce other problems (regression) ?
Thanks

PR #1483, opened 2024-08-28
fix: Avoid showing error page on cancelled navigation

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes GH-959.

### Description

If the navigation was cancelled, don't show an error page. This can cause issues where a subsequent request never loads into the webview because the error page is being shown.

This was previously implemented back in the UIWebView days as CB-13222 in https://github.com/apache/cordova-ios/pull/334 and was then proposed in https://github.com/apache/cordova-plugin-wkwebview-engine/pull/84 for the WKWebView plugin, but that didn't get reviewed/merged before the repo was archived. The fix is still valid, so we'll port it to the current version of the WKWebView plugin.

### Testing

All existing tests passed.

### Checklist


erisu (approved):
LGTM

Issue #1464, opened 2024-08-16
Rename generated Xcode project to "CordovaApp" or something consistent


dpogue:
# Feature Request

## Motivation Behind Feature

Currently, the generated Xcode project uses the application name for both the project and the target name. This means that file paths in generated projects are not consistent and we have to do regex-based token replacements during project creation. It also makes modifying the template project a pain because opening the template project directly in Xcode causes a bunch of those tokens to break.

The deeper problem is that some tooling (notably Cocoapods) struggles with project and target names with non-ASCII characters. By making the project name consistent, we can hopefully avoid those problems.

This could potentially also solve long-standing issues around `<config-file>` and `<edit-config>` for things like the Info.plist (currently named `{{projectname}}-Info.plist`) and node-xcode corrupting other targets due to not being able to restrict changes to just the app target.

## Feature Description

The generated Xcode project would be named `App.xcodeproj` or `CordovaApp.xcodeproj`.
The application target within that project would be named `App` or `CordovaApp`.

**Prior Art:** Capacitor projects are named `App.xcodeproj` with a default target named `App`.

## Alternatives or Workarounds

The current method mostly works (as long as you don't have non-ASCII characters in the project name), so there's a valid argument for keeping the existing behaviour.

The workaround for now is to generate a project with an ASCII name and then override the app's display name, which is pretty much what this change would accomplish by default.

Ref https://github.com/apache/cordova-ios/issues/717.
Ref https://github.com/apache/cordova-ios/issues/1150.
Ref https://github.com/apache/cordova-ios/issues/1289.

erisu:
I think having a consistent project name is a great idea.

Beyond the issues caused by non-ASCII project names, I've also heard of cases where hook scripts were needed to modify something within the project workspace. However, because the project name was inconsistent, they had to parse directory paths to extract the project name. A consistent project name would simplify this process.

It would also help clean up some of the node side in Cordova-iOS.

erisu:
@Mrkvozrout you might want to take a look at PR #1485

The above PR is refactoring to make the name of the app template/Xcode project consistent and will be hard set to `App`. E.g. `App` dir, `App.xcworkspace`, & `App.xcodeproj` etc...

The `name` element will remain the same `<name short="Short App Name">Long Name of My App</name>` but wont be used for defining any project file or directory names.

Mrkvozrout:
I saw it and understood but thought that custom project name might be useful and offered solution.

[...]

Issue #1289, opened 2023-02-21
cordova-ios 6.2.0: onesignal-cordova-plugin@3.3.0 #import <OneSignal/OneSignal.h>file not found when name contains non-latin1 symbols


anatoly-spb:
Thank you for the great tool such as Cordova!

I have Ionic6 project powered by Cordova (https://github.com/anatoly-spb/ionic6-hello) and get **#import <OneSignal/OneSignal.h>file not found** during build. 

My ionic info:
```
Ionic:

   Ionic CLI                     : 6.20.8 (/Users/jenkins/jenkins_slave/tools/jenkins.plugins.nodejs.tools.NodeJSInstallation/node16.19.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.5.4
   @angular-devkit/build-angular : 15.1.6
   @angular-devkit/schematics    : 15.1.6
   @angular/cli                  : 15.1.6
   @ionic/angular-toolkit        : 8.0.0

Cordova:

   Cordova CLI       : 11.1.0
   Cordova Platforms : ios 6.2.0
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 5.0.0, (and 6 other plugins)

Utility:

   cordova-res : 0.15.4
   native-run  : 1.7.1

System:

   ios-deploy : 1.11.4
   ios-sim    : 8.0.2
   NodeJS     : v16.19.0 (/Users/jenkins/jenkins_slave/tools/jenkins.plugins.nodejs.tools.NodeJSInstallation/node16.19.0/bin/node)
   npm        : 8.19.3
   OS         : macOS Monterey
   Xcode      : Xcode 14.2 Build version 14C18
```

I have found the cause of this problem. This is non latin1 symbols in name in the config.xml:
```
    <name>Тестовый проект Ionic6</name>
```
https://github.com/anatoly-spb/ionic6-hello/blob/main/config.xml#L3

Steps to reproduce the problem:

1. install node 16.19.0 cordova@11.1.0 cordova-res@0.15.4 native-run@1.7.1 @ionic/cli@6.20.8
2. git clone https://github.com/anatoly-spb/ionic6-hello
3. cd ionic6-hello
4. npm install
5. ionic cordova platforms add ios
6. cd platforms/ios
7. pod repo update
8. pod install
9. cd ../..
[...]

Issue #1150, opened 2021-09-21
Pod error on app with scandinavian chacters in the name


cbjerg:
# Bug Report

## Problem
Using @havesource/cordova-plugin-push 2.0.0 in an app with a special character in the name (like æ ø å)results in a linker error. Pods install fine, but Xcode shows a build error: ld: framework not found Pods_Special_L_ringsportal (Project name is Special Læringsportal). All pods are in the correct places, and named correctly in the filesystem

### What is expected to happen?
App would build, as it does without adding the @havesource plugin

### What does actually happen?
Xcode shows a build error: ld: framework not found Pods_Special_L_ringsportal

### Environment, Platform, Device
MacOS: Catalina 10.15.4
Xcode: 12.4
Cordova-ios: 6.2.0 (tried 6.1.0 and 6.0.0 withg same result)

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

meidlinga:
Probably same as: https://github.com/havesource/cordova-plugin-push/issues/95

TripShade:
We are also experiencing this.

meidlinga:
@TripShade , we worked around this issue the following way:
* Set the name itself to an alias without special characters (e.g. part of the package name). This wont be visible in the end.
* Set the short name to the prefered app name containing special characters
Example: `<name short="{{= name}}">{{= alias}}</name>`

* Set the Android App label explicitly in the Manifest
`
<edit-config file="app/src/main/AndroidManifest.xml" mode="merge" target="/manifest/application">
[...]
    <application android:label="{{= name}}" />
</edit-config>
`

TripShade:
> @TripShade , we worked around this issue the following way:
> 
> * Set the name itself to an alias without special characters (e.g. part of the package name). This wont be visible in the end.
> 
> * Set the short name to the prefered app name containing special characters
> 
> Example: `<name short="{{= name}}">{{= alias}}</name>`
> 
> 
> 
> * Set the Android App label explicitly in the Manifest
> 
> ```
> 
> <edit-config file="app/src/main/AndroidManifest.xml" mode="merge" target="/manifest/application">
> 
> [...]
> 
>     <application android:label="{{= name}}" />
> 
> </edit-config>
> 
> ```

Thx I'll give this a go 😃

meidlinga:
@TripShade : I added a step i forgot to mention initially

[...]

Issue #717, opened 2019-11-11
Podfile multiple targets generated if single quote is used in app name


mikealo:
# Bug Report
Podfile multiple targets generated if single quote is used in app name, this issue is not created if app does not contain single quote

## Problem
If a single quote is used in the app name for example "Test's single quote" and there is any additional plugin that use any pod library then the podfile generated contains multiple targets and cordova fail the plugin installation

### What is expected to happen?
The podfile generated should look like this:

```
platform :ios, '9.0'
target 'Test\'s single quote' do
	project 'Test\'s single quote.xcodeproj'
	pod 'GoogleMaps', '~> 2.0.0'
end

```
### What does actually happen?
The podfile generated look like this:

```
platform :ios, '9.0'
target 'Test\'s single quote' do
	project 'Test\'s single quote.xcodeproj'
end
target 'Test\'s single quote' do
	project 'Test\'s single quote.xcodeproj'
	pod 'GoogleMaps', '~> 2.0.0'
end
```

### Command or Code
```
cordova create "Test's single quote" com.test.singlequote "Test's single quote"
cordova platform add ios
cordova plugin add cordova-plugin-googlemaps

```

### Version information
MacOS 10.15
Xcode 11.2
Node v12.5.0
Ruby Gem 3.0.6
CocoaPods 1.8.4
Cordova 9.0.0 (cordova-lib@9.0.1)
Cordova iOS platform 5.0.1

## Checklist
- [X] I searched for existing GitHub issues
- [X] I updated all Cordova tooling to most recent version
- [X] I included all the necessary information above

mibzman:
I'm having this issue as well.  @mikealo, did you discover a workaround?

mikealo:
@mibzman unfortunatly i ended app changing app name and removing single quote.

Even if you change the podfile generated in the project>platform>ios folder manually you can then create errors during build process so i avoided doing so.

The file on ios platform that generate the podfile should be this one:

https://github.com/apache/cordova-ios/blob/master/bin/templates/scripts/cordova/lib/Podfile.js 

But even if change it manually this can lead to other errors so i think this is something that should be easily corrected by who works on it.

mibzman:
Thanks for the response.  I've also ended up changing my app name.  I removed the single quote from the  `name` property and went with a combination of [this plugin](https://github.com/mibzman/cordova-plugin-app-name) and [this config](https://stackoverflow.com/a/47583130/2653503) to change the display name back to a name with the single quote.

[...]

PR #1485, opened 2024-08-30
[WIP] refactor!: Consistent app template/Xcode project name

dpogue:
### Platforms affected
iOS

### Motivation and Context

Currently, the generated Xcode project uses the application name for both the project and the target name. This means that file paths in generated projects are not consistent and we have to do regex-based token replacements during project creation. It also makes modifying the template project a pain because opening the template project directly in Xcode causes a bunch of those tokens to break.

The deeper problem is that some tooling (notably Cocoapods) struggles with project and target names with non-ASCII characters.

Possibly closes GH-717.
Possibly closes GH-1150.
Possibly closes GH-1289.
Closes GH-1464.

### Description

Remaining work: 
- [ ] Set the `PRODUCT_NAME` to the config.xml name value so that the resulting app/ipa file has the expected name
- [ ] See if we can set `PRODUCT_BUNDLE_IDENTIFIER` in a better way at create-time to avoid token replacement
- [ ] Test with https://github.com/apache/cordova-common/pull/212 to see if we can rename `App-Info.plist` to just `Info.plist`
- [ ] Test with Cocoapods to make sure they still work as expected
- [ ] Test with Cocoapods and non-ASCII project names to ensure everything works

### Testing

All unit tests pass.
Tested manually with mobilespec.

### Checklist


erisu (approved):
LGTM

Built & launch mobile spec test app. Ran test against default plugins.

PR #1493, opened 2024-10-04
fix(compat): Set the app delegate viewController for plugins

dpogue:
### Platforms affected
iOS

### Motivation and Context

Some plugins expect to be able to access the applications's CDVViewController via the AppDelegate. While previously we guaranteed that the `viewController` property was set to a non-nil CDVViewController, there was never any guarantee that it was set to the CDVViewController that is actually displaying the app (particularly in cases such as apps using CordovaLib only for a few pages as part of a larger app).

We've attempted to deprecate the property here, but some plugins still rely on the assumption that it will be non-nil, so we are trying to maintain compatibility in the common case here by assigning to it when a CDVViewController is loaded. This means there are still times in the app lifecycle where the `viewController` property is nil, but the deprecation warning will hopefully spur plugin developers to move away from that pattern.

> [!NOTE]
> CDVPlugin always has a non-nil viewController property of its own, which is set to the CDVViewController instance the plugin was loaded into. This is almost always the better choice for plugins to use.

### Testing

Existing unit tests pass.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -22,6 +22,7 @@
 #import <WebKit/WebKit.h>
 #import <objc/message.h>
 
+#import <Cordova/CDVAppDelegate.h>
 #import <Cordova/CDVPlugin.h>
 #import "CDVPlugin+Private.h"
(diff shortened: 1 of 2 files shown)


erisu (approved):
LGTM

PR #1495, opened 2024-10-08
fix(emulators): Fix listing emulator targets

dpogue:
### Platforms affected
iOS

### Motivation and Context

I think I accidentally broke this during earlier refactoring for 8.0.0, but it was now trying to (and incorrectly) listing the simulator devices rather than the simulator device/runtime combinations.

### Description

Restore the correct listing for simulator devices/runtimes when using `cordova run ios --list`

### Testing

Tested running `npx cordova run ios --list` and ensured the output matched between these changes and 7.1.1

### Checklist

### DIFF
--- a/lib/listEmulatorBuildTargets.js
+++ b/lib/listEmulatorBuildTargets.js
@@ -18,7 +18,6 @@
 */
 
 const execa = require('execa');
-const { events } = require('cordova-common');
 
 /**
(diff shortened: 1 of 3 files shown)


erisu (approved):
LGTM

Built a sample app, added the fixed platform, ran `cordova run ios --list` command, and confirmed that the simulators were listed.

Issue #1372, opened 2023-09-20
cordova creating hundreds of Xcode 14 "stale files outside root paths" warnings


rolinger:
Every build from CLI:  `cordova build ios` or `ionic cordova build ios` - is taking almost 6 minutes to complete the build.  After the build, then from Xcode 14 to deploy the app to a device takes another 3 minutes.

In Xcode, during the deploy to device, the `issue navigator` is tracking nearly 2500 `stale files outside of allowed root path` warnings.  If I clean and run the project `<CMD> + SHIFT + K` & `<CMD> + SHIFT + R` - which takes a few minutes to complete, then all those stale files are removed and the deploy then takes like 15 seconds.  

But the next time I do a `cordova build ios` all those stale files return.  I believe the 6 minute build time is because cordova is creating those 2500 stale files...and Xcode is trying to process/analyze them during the deploy causing 3+ more minutes of delay.

What is causing these stale files to be created...and what can I do to fix it?

My Mac environment:
```
Cordova 11.1.0
Cordova-ios: 6.0.2
Ionic CLI: 6.20.8
Ionic toolkit: 3.2.15
native-run: 1.7.1
```

Example Warnings:
```
Stale file '/Users/rpo/dev/myApp/platforms/ios/build/emulator/nanopb.framework' is located outside of the allowed root paths.
Stale file '/Users/rpo/dev/myApp/platforms/ios/build/emulator/nanopb.framework.dSYM/Contents/Resources/DWARF/nanopb' is located outside of the allowed root paths.
Stale file '/Users/rpo/dev/myApp/platforms/ios/build/emulator/nanopb.framework/Headers' is located outside of the allowed root paths.
Stale file '/Users/rpo/dev/myApp/platforms/ios/build/emulator/nanopb.framework/Headers/nanopb-umbrella.h' is located outside of the allowed root paths.
Stale file '/Users/rpo/dev/myApp/platforms/ios/build/emulator/nanopb.framework/Headers/pb.h' is located outside of the allowed root paths.
```

dpogue:
Can you try the latest `cordova-ios 7.0.1` and see if you're still getting these warnings? There were some changes to Xcode derived data paths in the 7.x release that will hopefully resolve these sorts of issues

breautek:
> To do so I must also upgrade to Cordova 12 and I am not quite ready to go through all the pain yet.

You could try upgrading to cordova-ios@7 while staying on CLI v11 but some features might work.

[...]

PR #1496, opened 2024-10-08
fix: Resolve Xcode stale files warnings

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes GH-1372.

### Description

To get output app/ipa files in the expected places, we were overriding the `SYMROOT` value for command-line builds. However, this doesn't get applied when building in Xcode, and then it complains about "Stale files" in the Derived Data folder due to some things having been compiled with one SYMROOT value and others with a different one.

Setting SYMROOT is a hack anyways, what we really want to do is set an install path and force the app bundles to be installed into the destination folders after they are built. This allows Xcode to use the same Derived Data path for all builds, and avoids cluttering the Cordova destination folders with intermediate build artifacts.

Except, that we can't do this for iOS app bundles because changing the install path corrupts the xcarchives that are produced as part of exporting an ipa bundle. So we need to make all this `INSTALL_PATH` and `DEPLOYMENT_ROOT` stuff conditional based on whether we're building for a Simulator, Catalyst/macOS, or for a real iOS device.

### Testing

Alternated building a project in both Xcode and command-line and ensured that there were no warnings about stale files.

(Note if you are trying this yourself, it needs to be a new project that has not been built already, otherwise you need to clean it to remove any existing stale files pre-dating this change)

### Checklist


erisu (approved):
LGTM

- Cleared out DerivedData.
- Created new project
- Built app in CLI and confirmed no stale output
- Built app in Xcode. Didn't notice stale output in build report tab.

PR #1498, opened 2024-10-14
docs: Add upgrade guide for plugin authors for 8.x

dpogue:
### Platforms affected
Docs

### Motivation and Context

We want to make the migration process easier for plugin authors who want to support Cordova iOS 8, in light of the deprecations and other breaking changes.

### Description

Adds an "Upgrading Plugins to Cordova iOS 8.x" article to the DocC documentation bundle. This will get published on GitHub Pages, as well as being available from within Xcode.

### Testing

Built the documentation locally.

### DIFF
--- a/CordovaLib/CordovaLib.docc/upgrading-8.md
+++ b/CordovaLib/CordovaLib.docc/upgrading-8.md
@@ -1,0 +1,300 @@
+# Upgrading Plugins to Cordova iOS 8.x
+<!--
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+# http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+-->
+
+A guide for plugin authors to understand the API changes in Cordova iOS 8.
+
+Cordova iOS 8 introduces some significant changes to the exposed API for plugin authors and framework consumers. Many of these changes have been made to better align the framework with modern iOS development practices, such as adopting the `UIScene` APIs and fixing conflicts with SwiftUI projects, as well as work to improve the API experience for apps that consume Cordova as a framework (sometimes called the "platform centric workflow"). In all cases, great care has been taken to try to minimize the risk of breakage to existing 3rd party plugins.
+
+Many plugins will notice new deprecation warnings when built against Cordova iOS 8, rather than outright breaking changes. This document aims to explain the changes, the rationale behind the changes, and offer sample code for plugin authors to ensure their plugin code is compatible with future versions of Cordova iOS.
+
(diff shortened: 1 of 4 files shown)


CordovaLib/CordovaLib.docc/upgrading-8.md
```diff
@@ -0,0 +1,189 @@
...
+```objc
+// Old code
+// Include example here of how to get the root window
+```
+
+
+```objc
+// New code
+// Include example here of how to get the root window
+```
```
dpogue, line 72: TODO: Need sample code here

CordovaLib/CordovaLib.docc/upgrading-8.md
```diff
@@ -0,0 +1,189 @@
...
+// New code
+#import <objc/message.h>
+
+UIScrollView *scroller;
+SEL scrollViewSelector = NSSelectorFromString(@"scrollView");
+
+if ([self.webView respondsToSelector:scrollViewSelector]) {
+    scroller = ((id (*)(id, SEL))objc_msgSend)(self.webView, scrollViewSelector);
+}
+```
```
dpogue, line 101: TODO: Maybe add a Swift example here too?

CordovaLib/CordovaLib.docc/upgrading-8.md
```diff
@@ -0,0 +1,189 @@
...
+
+```swift
+// New code
+self.commandDelegate.send(CDVPluginResult(status: .ok), callbackId: command.callbackId);
+```
+
+These aliases were introduced in and are backwards compatible with all existing versions since Cordova iOS 5.0.0. See ``CDVCommandStatus`` for a list of the enum values.
+
+## Public API Removals
```
dpogue, line 174: TODO: Reformat and clean up everything listed below here

CordovaLib/CordovaLib.docc/upgrading-8.md
```diff
@@ -0,0 +1,189 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+-->
+
+A guide for plugin authors to understand the API changes in Cordova iOS 8.x.
+
```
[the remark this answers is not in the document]
dpogue: TODO: Include sample code for how to detect Cordova iOS 8 using the `CORDOVA_VERSION_MIN_REQUIRED` and `__CORDOVA_8_0_0` macros

erisu (approved):
LGTM

CordovaLib/CordovaLib.docc/upgrading-8.md
GitToTheHub: Could a helper method created for this? I don't understand, why someone has to this steps by its own, when he want to get the scrollView.

dpogue: The helper method that we have now is monkeypatched onto the base `UIView` class as a public property in a category extension, and this causes errors for Swift classes that are subclasses of `UIView` with their own `scrollView` properties that declare `package`, `fileprivate`, or `internal` visibility (because our helper requires those override to be `public`). See https://github.com/apache/cordova-ios/issues/1399 for details.

Removing the `UIView` category is the best way to prevent conflicts with other code, including system libraries.

Unfortunately, when a plugin accesses the WebView in a Cordova app, it's typed as a `UIView` because we don't know the actual implementation type of the web view (`WKWebView`, `UIWebView`, or some 3rd party webview that's only allowed to exist in the EU). Hence the need to dynamically dispatch the selector.

However, it is hopefully pretty rare for plugins to need to directly interface with the webview's internal scrolling implementation...

GitToTheHub: Ok I understand, I think also, this should be are rare circumstance to access the scrollView I think.

UIWebView should not a think anymore, because Apple forces to use WKWebView, but I didn't know it could be a 3rd party WebView in a Cordova app.

Issue #1422, opened 2024-04-18
Copy deployment-target from root config.xml to platform-specific config.xml before usage in Podfile


expcapitaldev:
**Issue**
Currently, the deployment-target variable is not being copied from the root config.xml file to the platforms/ios/[App Name]/config.xml before it is accessed in the code. As a result, the script uses the default deployment target from the template (https://github.com/apache/cordova-ios/blob/rel/7.0.1/templates/project/__PROJECT_NAME__/config.xml) instead of the user-specified value in the root config.xml.
It is mean in Podfile we will have deployment-target from root config.xml only first successful prepare hook.
Reproduce is very easy: try to install any pod library with ios target version more than 11. -> you will have error

**Expected Behavior**
The deployment-target set in the root config.xml should be reflected in the platforms/ios/[App Name]/config.xml before any script or process accesses it.

**Proposed Solution**
Modify the lib/prepare.js file to ensure that deployment-target and possibly other preferences are copied from the root config.xml to the platforms/ios/[App Name]/config.xml early in the preparation process. This could potentially be implemented at line 250, before the preferences are required by subsequent processes in the script.

**Additional Information**
This change will ensure that the deployment target and other settings are correctly set according to the project's configuration before they are needed, preventing any default settings from causing conflicts or errors during build or runtime.

**Workaround:**
- do form from cordova-ios library or do any postinstall here: https://github.com/apache/cordova-ios/blob/master/lib/Podfile.js#L40. to `this.minDeploymentTarget = minDeploymentTarget || '13.0';`

breautek:
This was an issue in cordova-ios 6.3 but I believe was patched in 7.0 (definitely in 7.1).

I have a google maps plugin that requires iOS 15 and I used to manually workaround this issue but it was corrected after I updated my projects to 7.1.

I used:

```xml
<platform name="ios">
  <preference name="deployment-target" value="15.0" />
</platform>
```

adding the plugin and then the platform works as expected, with the generated podfile properly containing `15.0` instead of the default `11.0`. This also works if you add the platform first, or use the preference outside of the `<platform name="ios">` block. Updating the preference and running `cordova prepare ios` also shows the changes reflected in both the platform's `Podfile` and `config.xml`.

> The deployment-target set in the root config.xml should be reflected in the platforms/ios/[App Name]/config.xml before any script or process accesses it.

[...]

PR #1502, opened 2024-11-11
fix(create): Copy config.xml data during creation

dpogue:
### Platforms affected
iOS

### Motivation and Context

This ensures that by the time plugin installation happens, the preferences from the root application config.xml can be read as expected.

Closes GH-1422.

### Description

Merge the app's config.xml into the defaults.xml at creation time for the platform config.xml.

Also removed comments about the `link` option since that's no longer relevant (with CordovaLib consumed as a Swift package, it is always linked).

### Testing

Existing tests pass, and added a new unit test to verify new behaviour.

### Checklist

### DIFF
--- a/lib/Api.js
+++ b/lib/Api.js
@@ -107,8 +107,6 @@ class Api {
      * @param  {Object}  [options]  An options object. The most common options are:
      * @param  {String}  [options.customTemplate]  A path to custom template, that
      *   should override the default one from platform.
-     * @param  {Boolean}  [options.link]  Flag that indicates that platform's
-     *   sources will be linked to installed platform instead of copying.
      * @param {EventEmitter} [events] An EventEmitter instance that will be used for
      *   logging purposes. If no EventEmitter provided, all events will be logged to
      *   console
--- a/lib/create.js
+++ b/lib/create.js
@@ -19,7 +19,7 @@
 
 const path = require('node:path');
 const fs = require('node:fs');
-const { CordovaError, events } = require('cordova-common');
+const { ConfigParser, CordovaError, events, xmlHelpers } = require('cordova-common');
 const xcode = require('xcode');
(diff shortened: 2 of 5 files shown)


erisu (approved):
Code LGTM

jkronborg:
Hi, thanks for the fix!

The issue fixed here prevents us from upgrading to version 17 of [cordova-plugin-firebasex](https://github.com/dpa99c/cordova-plugin-firebasex) because our minimum deployment version setting of 14 is not being set in time for the plugin installation, meaning the Cordova default of 11 is used, causing the build to fail (Firebase v. 11 requires a minimum target of at least 13):

> [!] CocoaPods could not find compatible versions for pod "FirebaseAuth":   In Podfile:     FirebaseAuth (= 11.2.0)  Specs satisfying the `FirebaseAuth (= 11.2.0)` dependency were found, but they required a higher minimum deployment target.

This plugin upgrade is a requirement for building with Xcode 16 due to https://github.com/dpa99c/cordova-plugin-firebasex/issues/892, which Apple requires by April 24.

Do you plan to release this before April 24?

jkronborg:
> > Do you plan to release this before April 24?
> 
> Due to the volunteer nature of Cordova development and maintenance, we can't make any commitments around release timelines.
> 
> This will be part of the cordova-ios 8.0.0 major version, and there are a few other big features to wrap up for that, namely support for Swift Package Manager.

OK, makes sense. Just wanted to make you aware of the issue, as I imagine quite a few Cordova projects use the Firebase plugin. I'll try to find an acceptable workaround.

Thanks for all your work on Cordova 🙏

jkronborg:
Our actual issue turned out to be https://github.com/ionic-team/cordova-plugin-ionic-webview/issues/695. Sorry for the confusion.

PR #1509, opened 2024-11-22
refactor(tests): Use XCTestPlans for unit testing

dpogue:
### Platforms affected
iOS

### Motivation and Context

This allows us to set up default behaviour of re-running tests up to 3 times if they fail, which should significantly cut down the number of failing builds in CI.

### Description

Move from Xcode schemes to XCTestPlans.

I also opted to only keep the unit tests for the case where CordovaLib is consumed as a framework, since that's now the default behaviour for the template app (using Swift Packages) rather than a static library.

Also cleaned up the test app project naming and moved some boilerplate stuff over to Swift.

### Testing

Tests pass locally and in CI.

### Checklist


erisu (approved):
LGTM

PR #1515, opened 2024-12-17
feat(spm): Support plugins as Swift packages

### LINKED ISSUE #1089: Add Swift Package Manager Support for iOS Plugins
# Feature Request

With recent versions of Xcode, including frameworks became problematic due to the introduction of the M1 Mac and the inclusion of certain simulator architectures within frameworks.  To address this, framework consumers use Lipo to strip architectures, or (better) framework providers moved to XCFramework.  Other framework providers jumped directly to offering their projects as Swift Packages to overcome this challenge.  It would be nice to add a way for plugins to specify Swift Package Manager dependencies, just like they currently can specify frameworks.

## Motivation Behind Feature

There are complicated or manually intensive ways to add Swift Packages to a Cordova project, so it would make sense to add real support, as this does appear the direction things are going.

https://swift.org/package-manager/
https://developer.apple.com/documentation/xcode/adding_package_dependencies_to_your_app

## Feature Description

Plugins would be able to add a line something like this:

`<package src="https://github.com/ReactiveX/RxSwift.git" branch="optional" version="6.0.0" revision="3920840923809482309482039"/>`

You could write a hook that calls something on the command line to add this, or (worse) manually add the package each time through Xcode.

dpogue:
### Platforms affected
iOS

### Motivation and Context
Ref GH-1089.

Swift package manager is the built-in way to handle 3rd party dependencies for Xcode and Swift projects, and Cocoapods is no longer in active development. We want to provide a way for plugins to express and manage their dependencies as Swift packages.

The best way to do that seems to be for the plugins themselves to (optionally) become Swift packages.

### Description
When a plugin declares in its plugin.xml file that it supports `<platform name="ios" package="swift">` and contains a Package.swift file in the plugin root, we will treat it as a Swift package and add it as a dependency of the app project (technically a dependency of a stub dependency, because that's easier than monkeying further with Xcodeproj files).

### Why This is Complicated
Each Swift package is treated as its own self-contained library, with access only to the dependencies that it declares. This means that plugins cannot access CordovaLib headers unless the declare a dependency on CordovaLib in their own Package.swift file. That would end up tying them to a very specific version of Cordova and potentially resulting in conflicts due to multiple plugins requesting different versions that don't match the app's version.

It feels like the only workaround is for the plugin install process to edit the Package.swift of each plugin and adjust the CordovaLib dependency to point to the local app one, but then we need to be able to parse and modify arbitrary Package.swift files.

Even with the plan to rewrite, it's not clear (for our own Apache plugins) if we can point them at the master branch of cordova-ios and still pass a release vote, because they would be pointing to a dependency that has not had a release vote. This could result in needing to churn the version dependency in every plugin every time we release a new cordova-ios version (even if we don't need to publish those changes).

### Approach Taken
Rather than editing Package.swift in the top-level plugins folder (which we hope one day to get rid of), we've added a packages directory to the platforms/ios folder and will copy plugins there if they are using Swift Package Manager. The added advantage here is that we don't have files in the platform project pointing to things in node_modules, so the project is more self-contained.

Since we now have a copy of the plugin inside the platform folder, we can edit its Package.swift to update any references to CordovaLib to point to the same module as the rest of the app so there are no issues with version mismatches.

### Remaining Work
- [x] Add unit tests around the CordovaLib path fixup, for both linked and non-linked CordovaLib
- [x] Figure out what we should do about CocoaPods
    (By this, I mean, what we should do when a plugin wants to consume a library through CocoaPods for compatibility with older releases but wants to consume the same library through SwiftPM going forward)
- [ ] Write some documentation for plugin authors about how to set up their plugins to use SwiftPM
- [x] Figure out how we want to (or are allowed to) reference CordovaLib for our own Apache plugins, if we want to move them to using SwiftPM at all

### Testing

Tested locally against this branch of cordova-plugin-device: https://github.com/dpogue/cordova-plugin-device/tree/spm

### Checklist


breautek:
> Even with the plan to rewrite, it's not clear (for our own Apache plugins) if we can point them at the master branch of cordova-ios and still pass a release vote, because they would be pointing to a dependency that has not had a release vote. This could result in needing to churn the version dependency in every plugin every time we release a new cordova-ios version (even if we don't need to publish those changes).

I would think it's not acceptable to point it to master branch as that would effectively be using the dev build. At minimum it would need to be using the `rel/<version>` tag. But we could potentially have a "Local" SPM package that is just simply shipped with the cordova-ios tarball. The app template could reference it locally via file path. Not 100% sure how that will work with plugins however.

> Each Swift package is treated as its own self-contained library, with access only to the dependencies that it declares. This means that plugins cannot access CordovaLib headers unless the declare a dependency on CordovaLib in their own Package.swift file.

I think this is more of a side issue because our plugins are not proper modules. They are just loose source code that eventually makes their way part of the App project. And I assume SPM packages are importing these packages as their own build targets, which means they have their own build configurations (and thus needs a reference to the CordovaLib framework).

This in itself shouldn't be a blocker though, as long as `CordovaLib` framework is consumed as a dynamic library, instead of a static library. Plugins should be able to declare the dependency on CordovaLib and compile against it without providing the CordovaLib framework itself. The application project will be the one responsible of importing the `CordovaLib` framework, and making sure that the dynamic library is available during runtime (by packaging it with it's product). Because most plugins are not pre-compiled, if they are any incompatibilities, it should appear during build time. Plugin authors pre-building their plugins and distributing a prebuilt binary however would need to take extra care to not call on an API that might not exist without properly guarding so, as it would cause runtime crashes.

> I think this is more of a side issue because our plugins are not proper modules.

Extending on this, a year or two ago I was experimenting with an alternate approach in authoring plugins, so that plugins can have their own native projects and can compile (and potentially run) independently of a Cordova project. Would be very useful to have for using native unit test features. This involved having plugins be in their own projects and having their own build targets, so each plugin would produce their own `.framework` which gets imported by the Cordova application. This mostly worked but I did run into runtime errors where the cordova framework failed to look up any symbol that came from a dynamic library, wihch the cordova framework does when processing native bridge calls [here](https://github.com/apache/cordova-ios/blob/3737f515ba0b30ac8e32595a9ab9626e7ae9401d/CordovaLib/Classes/Public/CDVCommandQueue.m#L184-L193) if my memory serves me right.

Just some thoughts. I've been experimenting with SPM with my own framework, and will be sharing what worked for me once I have something but it is a very different architecture than Cordova.

dpogue:
> This in itself shouldn't be a blocker though, as long as `CordovaLib` framework is consumed as a dynamic library, instead of a static library. Plugins should be able to declare the dependency on CordovaLib and compile against it without providing the CordovaLib framework itself.

This actually works if CordovaLib ***is*** a static library for the app target, but not when it's consumed as a Swift package itself (which is how we're using it in 8.0.0)

> I assume SPM packages are importing these packages as their own build targets, which means they have their own build configurations (and thus needs a reference to the CordovaLib framework)

This is correct, when a plugin is a Swift package, it becomes its own framework target that is linked to the app project, rather than copying plugin source files into the project directly.

dpogue:
In the case where it worked with the static CordovaLib, the plugin did not specify a dependency on CordovaLib as part of its Package.swift file

When using Package.swift, even if the app asks to consume CordovaLib at the top level as a static library, plugins cannot find Cordova headers unless they also include a dependency on CordovaLib

dpogue:
Okay, I've updated this into something that I hope might be workable. There are essentially 2 main changes:

1. All plugins are copied into a `packages` directory within the `platforms/ios` folder. The primary reason for this is so that the platform folder remains self-contained and can be zipped up and moved to another machine without references to things outside the platform folder itself.

2. Any package dependency on cordova-ios within a plugin will (hopefully) be rewritten when the plugin is copied, to that it points to the local path to CordovaLib (which is also now copied into the packages folder). This ensures that every plugin and the app itself are using the same version of CordovaLib.

This seems to work in my (fairly basic) testing with the cordova-plugin-device branch linked in the PR description.

TODO:
- [ ] Manage the path to cordova-ios when it is using `--link` (because we don't copy it to `packages` in that case)

erisu:
> The primary reason for this is so that the platform folder remains self-contained and can be zipped up and moved to another machine without references to things outside the platform folder itself.

I want to add that this statement might be partially correct.

If the host machine preparing the platforms is either Windows or Linux and the plugin contains pods, the process might not install and prepare the plugin correctly. I believe during `plugin add`, when `prepare` is triggered, the pod information is added to the `Podfile` and then `pod install` is executed. Since neither Linux nor Windows has the pod tooling, this step would fail.

Additionally, on a successful `pod install`, the next step would be to update the swift version information on a file in `Pods/Pods.xcodeproj`. I believe the entire `Pods` directory is created by `pod install`. If we suggested to the app developers to run it manually, they would probably need to figure out and set the Swift version as well.

If the host machine is macOS then this should not be a problem.
If the app does not use any plugins with pods, then this should not be an issue as well.

This would need testing for confirmation. If my assumptions are correct, we should document this behavior so it would be made aware.

msmtamburro:
This is great work.  On the topic of plugin consumption, considering the "tale of two audiences:"  I would imagine that native devs with "platform centered approach" projects would prefer to include the plugins as swift packages directly.  This begs the question of how the other bits in that native project (e.g., the post-processed config.xml) would get updated, other than manually.

dpogue:
> I would imagine that native devs with "platform centered approach" projects would prefer to include the plugins as swift packages directly.

The complication here (if managing plugins manually in a project's Package.swift) becomes making sure that all those plugins refer to the same version of cordova-ios for their CordovaLib dependency. That's why we're needing to rewrite the Package.swift when we install them in this PR.

SwiftPM doesn't have particularly good support for overriding library versions from sub-dependencies or a concept like npm's peerDependencies 😞

dpogue:
Regarding the `nospm` attribute on `<pod>` tags in plugin.xml:

* if `<platform name="ios">` and `<pod ...>`
  * **install**
* if `<platform name="ios" package="swift">` and `<pod ...>`:
  * **install**
* if `<platform name="ios" package="swift">` and `<pod nospm="false" ...>`:
  * **install**
* if `<platform name="ios" package="swift">` and `<pod nospm="true" ...>`:
  * _skip_
* if `<platform name="ios">` and `<pod nospm="true" ...>`:
  * **install** (the attribute is ignored if `package="swift"` is no specified on the platform)

This behaves like the `nomodule` option on `<script>` tags in HTML (except it's XML, so we can't use a true boolean attribute).

A boolean string `"true"` means that the pod should **not** be installed when the project supports Swift packages (but should be installed in older projects), `"false"` (or no attribute at all) means it should always be installed.

This provides a mechanism for plugins to transition their pod dependencies to SwiftPM dependencies while retaining backwards compatibility with older cordova-ios versions.

lib/SwiftPackage.js
```diff
@@ -0,0 +1,110 @@
...
+            const pkgSwiftPath = path.join(localPluginPath, 'Package.swift');
+
+            let cordovaPath = path.relative(localPluginPath, path.join(this.root, 'packages', 'cordova-ios'));
+            if (!fs.existsSync(path.join(localPluginPath, cordovaPath))) {
+                cordovaPath = path.relative(localPluginPath, ROOT);
+            }
+
+            const pkg_fd = fs.openSync(pkgSwiftPath, 'r+');
+            let packageContent = fs.readFileSync(pkg_fd, 'utf8');
+            packageContent = packageContent.replace(/package\(.+cordova-ios.+\)/gm, `package(name: "cordova-ios", path: "${cordovaPath.replaceAll(path.sep, path.posix.sep)}")`);
```
erisu, line 61:
```suggestion
            // Note: May match unintended packages, e.g. 'cordova-ios-plugin', due to flexible patterns. Strict filtering would risk valid exclusions.
            packageContent = packageContent.replace(/package\(.+cordova-ios.+\)/gm, `package(name: "cordova-ios", path: "${cordovaPath.replaceAll(path.sep, path.posix.sep)}")`);
```

Should we add a comment about the pattern detection and why it can not be strict?

erisu (approved):
LGTM

Ran a simple test by updating an existing plugin to test the following use case:

> if `<platform name="ios" package="swift">` and `<pod nospm="true" ...>`:
>    skip

I confirmed: 

- When using platform version 7.x, the pod is still installed.
- When using this branch, the pod was not installed by Pod and was controlled by SPM.

I also included a test when two `<pod>` elements existed, one with and without `nospm="true"`. In this test case, I just added a new pod element to an existing plugin and confirmed that the `Podfile` was updated and the podspec was checked out.

- The pod element with `nospm="true"` was skipped by pods and handled by SPM
- The pod element without `nospm="true"` continued to install by Pods.

I left some other comments but are not blockers.

dpogue:
I've updated unit tests to cover the CordovaLib path rewriting and address the comments, so I think this is probably pretty well ready to go.

erisu (approved):
Tested and confirmed the PR to work.

There was a run issue with the test plugin but appeared I needed to clear out the DerivedData directory before running it again. After clearing the DerivedData it worked as expected.

breautek:
I know it's a bit late for feedback, but thought to at least write down my experience somewhere and maybe it can be a foundation for future improvements.

> SwiftPM doesn't have particularly good support for overriding library versions from sub-dependencies or a concept like npm's peerDependencies 

From my recent work in doing my upgrades of several internal plugins and such I don't really believe this is the case.

Because SPM will not accept duplicate packages of varying versions. Every package that declares a dependency on a package will cause that package to resolve to the latest version of that package that is still satisfies all the version constraints across all packages. If there is a version incompatibility (say one package wants v1 of a dependency, and another package wants v2 of that same dependency) then package resolution will fail and the project simply won't be buildable.

For example I have the following packages:

sqlite v0.4.5 (transient dependency, app does not declare this dependency directly)
tilegen depends on sqlite using ^0.4.5 (upToNextMinor)
cordova-plugin-sqlite depends on sqlite using exact 0.4.5

The app resolves the sqlite package to 0.4.5.

So in theory plugins can define a loose version range on cordova-ios, either a particular major `(upToNextMajor(from: "8.0.0"))` which accepts all version 8, like NPM's `^8.0.0`. 

Couple of issues I can foresee:

1) I'm not sure what SPM does if you have a local dependency, I believe it acts as an any version -- so with the current configuration where the Cordova Plugins SPM package depends on the local cordova-ios which is technically versionless I think it may defeat the purposes of explicitly declaring supported versions to begin with.

2) We also can't simply do away with the local SPM module because the framework code is coupled with the CLI implementation for installing iOS platform & plugins, and if we pin to a specific version, it will obviously fetch a copy of cordova-ios rather than using what is bundled within NPM's tarball.

Issue #1488, opened 2024-09-06
Double dash prevents forwarding platform-specific options


luxaritas:
# Bug Report

### What is expected to happen?

`npx cordova build ios --release --device -- --automaticProvisioning` would pass `-allowProvisioningUpdates` to `xcodebuild`

### What does actually happen?

`xcodebuild` is not passed `-allowProvisioningUpdates`

## Information

Both the [cordova-cli docs](https://cordova.apache.org/docs/en/12.x/reference/cordova-cli/index.html#cordova-build-command) and [cordova-ios docs](https://cordova.apache.org/docs/en/12.x/guide/platforms/ios/index.html#using-flags) specify you need to add an extra `--` in order to provide platform-specific options. This is indeed the case for `cordova-android`. However for `cordova-ios`, adding the extra double dash appears to prevent the extra arguments from being processed.

It's possible that because I use `npx` something weird is happening here, but as far as I can tell, the behavior is inconsistent with `cordova-android`, so that shouldn't be the issue.

### Command or Code

`cordova build`

### Environment, Platform, Device
N/A

### Version information

Node 20, cordova 12.0.0, cordova-ios 7.1.0

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

dpogue:
I've confirmed this, and confirmed that cordova-ios never actually handles the `options.argv` that are forwarded from the CLI.

PR #1516, opened 2024-12-21
fix(build): Allow platform args with double dash

dpogue:
### Platforms affected
iOS

### Motivation and Context

The Cordova docs say to provide platform-specific arguments after a `--` on the CLI. Currently this works on Android, but does not work for iOS.

Closes GH-1488.

### Description

Ensure that we parse any extra (platform-specific) arguments passed from cordova-cli in to the iOS build command through `options.argv`. These are arguments that are provided after a `--` to the CLI command.

There is no change for existing arguments passed without the double dash.

The priority of arguments is now:
1. Platform-specific arguments passed after a double-dash.
2. Arguments passed without a double-dash (current behaviour).
3. Values provided in a build.json config file.

### Testing

Added unit tests to cover argument parsing and fallback behaviour.

### Checklist


erisu (approved):
LGTM

Tested command while passing in `buildFlag`.
- Before `--`
- After `--`
- Before and After `--`
  - Arguments after `--` are used. Before will be dropped as expected.
- Before `--` with `build.json`
  - Arguments in `build.json` are dropped in favor for CLI args. No merging.
- After `--` with `build.json`
  - Arguments in `build.json` are dropped in favor for CLI args. No merging.
- `build.json` only

Issue #1491, opened 2024-09-12
The macOS statusbar overlays the web view on Mac Catalyst


qpwr:
# Bug Report

### What is expected to happen?

If building for Mac Catalyst, the statusbar shouldn't overlay the web view.

### What does actually happen?

The statusbar (which can't be disabled/hidden on macOS) overlays the web view and covers content which can't be reached through scrolling.

## Information

The issue can be easily reproduced by using the HelloWorld example app and adding this right after the body tag in www/index.html:

`<div style="font-size:5rem;">EXAMPLE TEXT</div>`

After building the app **for Mac Catalyst** (not for Designed for iPad!), the view should look like this, the added text is cut off at the top:

![Bildschirmfoto 2024-09-12 um 10 50 39](https://github.com/user-attachments/assets/3baa40b3-07e5-41c0-a3eb-d8e4b7e64da6)

More info:

- The plugin cordova-plugin-statusbar doesn't help with the issue, I've tried all possible combinations
- None of the following settings in xcode seem to make any difference:
   - Mac Catalyst Interface: Optimize for Mac vs. Scaled to match iPad
   - Status Bar Style
   - Hide during application launch (on vs. off)
   - Requires Full screen (on vs. off)

### Environment, Platform, Device

- Macbook Pro 16'' 2023
- macOS Sonoma 14.5

### Version information

- cordova-ios 7.1.1

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

breautek:
I really don't know but couple more things to try that wasn't mentioned:

- `StatusBarOverlaysWebView` preference (feature provided by the [cordova-plugin-statusbar](https://github.com/apache/cordova-plugin-statusbar?tab=readme-ov-file#configxml)). If the plugin is installed it also defaults to true so it could explain this behaviour. You can try setting it to false.
- Does the webview respond to CSS safe area inset top?

```css
body {
    margin-top: env(safe-area-inset-top, 0px);
}
```

Normal iOS behaviour is if the status bar is in overlay mode then CSS safe-area-inset-top is set accordingly so that you can use CSS to clear the statusbar for positioning your DOM nodes. If the statusbar is not in overlay mode, then the native webview starts just under the status bar, and the safe-area-inset-top will have a value of `0px` accordingly (unless if the device has some other screen gimmick)

qpwr:
Thanks for replying!

[...]

PR #1517, opened 2024-12-22
fix(catalyst): Fix content behind titlebar

dpogue:
### Platforms affected
macOS Catalyst

### Motivation and Context

Closes GH-1491.

### Description

When showing a webview on macOS Catalyst, always position it below the titlebar when a titlebar is displayed. This has the side effect of making all the `safe-area-inset` values 0, regardless of the `viewport-fit` value.

Also adds a `HideDesktopTitlebar` preference to make the titlebar invisible on macOS Catalyst with content extending behind it. I named this in a generic way because it sounds like this should also be feasible to apply to the cordova-electron platform.

When using a preference to disable drawing the titlebar, `safe-area-inset` works as expected (but generally requires `viewport-fit=cover` to be specified).

### Testing

Tested with `viewport-fit=cover` and `viewport-fit=contain` and measured the calculated `safe-area-inset-top` value.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVPlugin.m
+++ b/CordovaLib/Classes/Public/CDVPlugin.m
@@ -21,7 +21,6 @@
 #import <Cordova/CDVPlugin+Resources.h>
 #import "CDVPlugin+Private.h"
 #import <Cordova/CDVViewController.h>
-#include <objc/message.h>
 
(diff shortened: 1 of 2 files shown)


erisu (approved):
LGTM

Tested building app with the flag set to true and false, view-port set to contain and cover, and with and without the safe-area's defined.

PR #1520, opened 2025-01-16
refactor!(SPM): rename Swift package from CordovaLib to Cordova

dpogue:
### Platforms affected
iOS

### Description

This ends up being less confusing for everyone because the resulting framework is named Cordova.framework and the header paths are all `<Cordova/*.h>` but currently you need to reference the Swift dependency as "CordovaLib" rather than "Cordova".

As far as I can tell from a quick search of public Package.swift files on GitHub, nobody has actually consumed this as a Swift package, so any breakage should be minimal and we should take the opportunity to do this right as part of the major version bump.

### Testing

Existing tests pass.

### Checklist

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,7 +28,7 @@ let package = Package(
         .macCatalyst(.v13)
     ],
     products: [
-        .library(name: "CordovaLib", targets: ["Cordova"])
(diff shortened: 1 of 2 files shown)


erisu (approved):
LGTM

PR #1522, opened 2025-01-17
refactor(template): Copy CordovaLib into a packages folder

dpogue:
### Platforms affected
iOS

### Motivation and Context

Pointing a relative path to CordovaLib in the node_modules folder means that the Xcode project is no longer self-contained, which is significant change in terms of assumptions people can make about the project files in CI environments (i.e., preparing all platforms once, then copying just the platform folders to different runners to build).

### Description

We now copy the CordovaLib folder into a `platforms/ios/packages` directory and point to it there. (We will need to do the same for SPM plugins, which sorta turns out to our advantage...)

As a side effect, it seemed worth bringing back the `--link` option to continue pointing to node_modules for easier development.

### Testing

Tested with mobilespec and unit tests.

### Checklist

### DIFF
--- a/lib/Api.js
+++ b/lib/Api.js
@@ -107,6 +107,8 @@ class Api {
      * @param  {Object}  [options]  An options object. The most common options are:
      * @param  {String}  [options.customTemplate]  A path to custom template, that
      *   should override the default one from platform.
+     * @param  {Boolean}  [options.link]  Flag that indicates that platform's
+     *   sources will be linked to installed platform instead of copying.
      * @param {EventEmitter} [events] An EventEmitter instance that will be used for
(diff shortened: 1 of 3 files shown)


erisu (approved):
- Tested building the sample app with no plugins.
- Confirmed that `platforms/ios/packages/cordova-ios/CordovaLib/` was created.
  - Also confirmed `relativePath = "packages/cordova-ios";` to be set correctly.
- Confirmed that the `--link` flag does not create the `packages/cordova-ios/CordovaLib` directory.
  - Also confirmed `relativePath = "../../node_modules/cordova-ios";` to be set correctly.

PR #1523, opened 2025-01-29
feat(statusbar): Simple built-in status bar support

dpogue:
### Platforms affected
iOS

### Motivation and Context

Web content authors should have control over how the status bar is handled using the built-in web platform functionality for UI colouring and opting in to showing behind the status bar or not.

### Description

This adds a background behind the status bar automatically when a page does not request to extend itself behind the status bar using a meta tag with `viewport-fit=cover`. It automatically shows and hides as the value of the viewport-fit option changes.

On iOS 16.4 and newer, it will be coloured according to the meta `theme-color` tag, and a default colour can be set in the storyboard for a CDVViewController.

We expose a very basic JS API hanging off the existing `window.statusbar` property that allows controlling the visibility (of the status bar contents entirely, not just of the background view) and overriding the background colour.

The colour is determined as follows:
  1. Any colour set explicitly with the JS API (if any)
  2. Any colour pulled from the meta tag (iOS 16.4+, if any)
  3. The status bar background colour specified in the storyboard (if any)
  4. The general background colour specified in the storyboard (if any)
  5. The default system background colour

Efforts were made to ensure that this does not interfere with the existing CDVStatusBar plugin implementation (although maybe this can replace it for most common use cases), by making the display of this built-in conditional on not having the CDVStatusBar plugin installed.

### Remaining Work
- [x] Wire the `StatusBarBackgroundColor` preference up to an XCAsset colorset in the template and set it in the Main.storyboard file
- [x] Do a bunch of testing with `CDVStatusBar` installed to make sure there are no conflicts

### Testing

Tested locally in an iPhone 16 Pro Max simulator.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/CDVViewController+Private.h
+++ b/CordovaLib/Classes/Private/CDVViewController+Private.h
@@ -17,19 +17,12 @@
  under the License.
  */
 
-#import "CDVLaunchScreen.h"
-#import "CDVViewController+Private.h"
+#import <Cordova/CDVViewController.h>
 
-@implementation CDVLaunchScreen
+@interface CDVViewController (Private)
 
-- (void)show:(CDVInvokedUrlCommand *)command
-{
-    [self.viewController showSplashScreen:YES];
-}
+- (void)setStatusBarWebViewColor:(UIColor *)color;
 
-- (void)hide:(CDVInvokedUrlCommand *)command
-{
-    [self.viewController showSplashScreen:NO];
-}
+- (void)showStatusBar:(BOOL)visible;
 
 @end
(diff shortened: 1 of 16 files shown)


dpogue:
TODO: think about whether the StatusBarOverlaysWebView preference from CDVStatusBar should change default behaviour here (override viewport-fit?)

dpogue:
> TODO: think about whether the StatusBarOverlaysWebView preference from CDVStatusBar should change default behaviour here (override viewport-fit?)

On second thought, the whole point of this is to give control to the web content so leaving the status-bar specific stuff to the plugin and keeping this simple seems best

erisu (approved):
LGTM

PR #1526, opened 2025-02-17
VCI-835 - cordova support spm

maheshm2023:
### Checklist


Package.swift
```diff
@@ -0,0 +1,38 @@
+// swift-tools-version: 5.9
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "Cordova",
+    platforms: [
+        .iOS(.v13)
```
thanarajjeganathan, line 9: why 13, we updated to 14 right?

PR #1536, opened 2025-05-17
test: Improve test coverage of Bridging Headers & fix edge case

dpogue:
### Platforms affected
iOS

### Motivation and Context

Our fixture file for Bridging Header tests didn't contain a bunch of stuff that the Bridging Header code tries to handle, so that code was never properly tested.

### Description

Update the Bridging Header fixture file to include more comment types and inline comments and not have a final newline.
That missing newline ended up uncovering a bug in the Bridging Header code where headers from multiple plugins could end up on a single line without separation, so that edge case is also fixed now.

### Testing

Ensured existing unit tests pass

### Checklist

### DIFF
--- a/lib/BridgingHeader.js
+++ b/lib/BridgingHeader.js
@@ -111,7 +111,7 @@ BridgingHeader.prototype.__parseForBridgingHeader = function (text) {
         }
     }
     if (start < i) {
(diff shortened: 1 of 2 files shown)


erisu (approved):
LGTM

PR #1543, opened 2025-06-08
chore(template): Move xcconfig files into the template folder

dpogue:
### Platforms affected
iOS

### Motivation and Context

The Xcode template project references these .xcconfig files, but they didn't actually live under the project folder so they always showed up as missing.

### Testing

I've confirmed that moving these is fine because the templates/cordova folder will get merged with the templates/project/cordova folder when it is copied.

### Checklist


erisu (approved):
LGTM

PR #1544, opened 2025-06-08
fix(swift): Fix some new Swift deprecation warnings

dpogue:
Can't even go one Xcode minor version without new stuff getting deprecated 🙄

### Platforms affected
iOS

### Motivation and Context

Fix some new Swift deprecation warnings about `@_objcImplementation`

### Description

Changed `@_objcImplementation` to `@objc @implementation` if we're using a Swift 6.1 compiler or newer.

To be fair, this is now an officially supported Swift feature rather than an experimental option, so this is better in the long run.

### Testing

Confirmed that this builds without deprecation warnings in Xcode 16.4.

### Checklist

### DIFF
--- a/templates/project/App/AppDelegate.swift
+++ b/templates/project/App/AppDelegate.swift
@@ -20,7 +20,11 @@
 import UIKit
 
 @main
+#if compiler(>=6.1)
+@objc @implementation
+#else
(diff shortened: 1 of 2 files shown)


erisu (approved):
LGTM

PR #1552, opened 2025-07-28
White Screen of Death issue | Enhance WebView reload detection to handle empty title strings

abroa01:
Improves white screen of death prevention by checking for both nil and empty string titles when determining if WebView needs reloading.
This catches more edge cases where the WebView enters a broken state but has an empty title instead of nil.

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -297,7 +297,7 @@
 
 - (BOOL)shouldReloadWebView:(NSURL*)location title:(NSString*)title
 {
-    BOOL title_is_nil = (title == nil);
+    BOOL title_is_nil = (title == nil) || [title isEqualToString:@""];
     BOOL location_is_blank = [[location absoluteString] isEqualToString:@"about:blank"];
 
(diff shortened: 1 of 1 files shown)


horner:
This is the cordova master: https://github.com/apache/cordova-ios/blob/47f23d54ba261c8c554d5d1be9b5dbbc3af0e057/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m

We need the links to what Meteor is using now (which is likely not the apache:master)

abroa01:
### Meteor uses v7.1.1 of cordova-ios
Link  - https://github.com/apache/cordova-ios/blob/7.1.1/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m

This is the file that gets downloaded when iOS platform is added in meteor. And I am running Meteor 3.3 
image

abroa01:
image

abroa01:
Meteor uses `cordova-ios@7.1.1`, which includes the `CDVWebViewEngine.m` file we're modifying.

- Link to the Cordova integration in Meteor 3.3:  
  https://github.com/meteor/meteor/tree/release-3.3/tools/cordova

- The version is pinned here in the devel branch:  
  https://github.com/meteor/meteor/blob/devel/tools/cordova/index.js#L28

This confirms that the file being patched comes from `cordova-ios@7.1.1`, the version bundled by Meteor.

PR #1577, opened 2025-10-28
docs: Add note about CDVPluginResult Swift nullability

### LINKED ISSUE #1574: CDVPluginResult non-nullability in Cordova-iOS 8 is a breaking change for Swift
Need to add this to the plugin upgrade guide that `CDVPluginResult` now returns a nonnull object so it's no longer an optional in Swift. If plugins were force unwrapping, this will now result in errors.

I think we want it to be nonnull, but we should double check that and make sure we have a section in the upgrade guide that explains how to write Swift code that is compatible with both older and current versions.

dpogue:
### Motivation and Context

Closes #1574.

### Description

Added a note to the Cordova iOS 8 upgrade guide about `CDVPluginResult` nullability changing and how that potentially impacts Swift plugins that try to force unwrap results.

### Checklist

### DIFF
--- a/CordovaLib/CordovaLib.docc/upgrading-8.md
+++ b/CordovaLib/CordovaLib.docc/upgrading-8.md
@@ -167,6 +167,26 @@ import Foundation
 import UIKit
 ```
 
+### `CDVPluginResult` Swift optionality
+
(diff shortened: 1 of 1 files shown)


erisu (approved):
LGTM

Issue #1393, opened 2024-01-10
(Black screen) when App Reload is triggered on iPhones with Notch:


Tarek200:
# Bug Report
(Black screen) when App Reload is triggered on iPhones with Notch:

We are using the "Construct 3" game engine. 
Every time the App triggers "Reload" it goes to a black screen. 
The "Reload"  triggers in various ways:

-Switching between Cordova Apps
or
-Resuming the App while its been for a long period on paused in the background.

**Note:**
This happens only on iPhones that have notch like for example "XS Max"

## Problem
When the App Reloads it goes to a black screen

Note:
Also, I don't think it should force the App to reload while switching Apps as it should just resume from pause to continue from where the App was before was closed.

### What is expected to happen?

That it doesn't go to a black screen

### What does actually happen?

It goes to a black screen when it forces a Reload

### Command or Code

**Steps to reproduce the bug:**
1-Install my cordoba project with Xcode and Run the App
2-When is fully loaded close the App by swapping the notch
3-Download and Open any of the C3 Cordova Games in the store like (Littel Dog Bob or Punch Golf)
4-Then switching back to my Test App it triggers the Reload and it goes to Black Screen.

**Video:**
https://github.com/apache/cordova/assets/41597700/f450fda6-8d42-4794-ae59-8cde1d4b42eb

**Cordova Project:**
[iOS canvas Bug.zip](https://github.com/apache/cordova/files/13889417/iOS.canvas.Bug.zip)

### Environment, Platform, Device

iOS (16 & 17)
iPhone XS Max

Other users could reproduce the bug on different iPhone models.

Note:
This issue only happens on iPhones that have Notch

### Version information

Cordova Platforms iOS
Xcode 15.0.1
iOS 7.0.1

## Checklist

- [ ] I searched for existing GitHub issues
- [ ] I updated all Cordova tooling to most recent version
- [ ] I included all the necessary information above

Issue #1355, opened 2023-07-04
Method shouldReloadWebView in CDVWebViewEngine misses condition, results in Blank Screen


rcorin:
# Bug Report

## Problem
I'm investigating a situation when the app comes to foreground from background and the webview content process has been killed by the OS.
In shouldReloadWebView , CDVWebViewEngine.m:283,  there is this code:

    BOOL title_is_nil = (title == nil);
    BOOL location_is_blank = [[location absoluteString] isEqualToString:@"about:blank"];

    BOOL reload = (title_is_nil || location_is_blank);

however location is nil, hence is false and also title_is_nil is false, resulting in reload being false.

### What is expected to happen?

If the location is nil, reload should be true, so the condition should be:

 BOOL location_is_blank = location == nil  || [[location absoluteString] isEqualToString:@"about:blank"];

### What does actually happen?

the app gets stuck in a white screen because the reload isn't issued, due to shouldReloadWebView being false.

## Information

To reproduce the issue, i had to actually run the app in simulator, kill the com.apple.WebKit.WebContent process with the app in background, and then comment out the other webview reloads until reaching the point of shouldReloadWebView.

### Command or Code

See above info.

### Environment, Platform, Device

iOS simulator

### Version information

Xcode 14.3, iOS 12, Cordoba 10.0.0.

## Checklist

- [X ] I searched for existing GitHub issues
- [ X] I updated all Cordova tooling to most recent version
- [ X] I included all the necessary information above

BenLaKnet:
Hi,

I have the same issue. What could I do, cordova-ios 7.1.1.

horner:
### Problem
It appears that when I background the Cordova app and use another application on iOS that uses lots of memory, when I return to the Cordova app it shows a white screen and appears to be frozen. There is no way to reload the page or navigate so I’m forced to kill the app and relaunch. 

Captured the White Screen of Death in the debugger:

Here is a short: https://youtube.com/shorts/4rV0T_peIew?si=U60VhrfND7t9yftR

<img width="1392" height="897" alt="Image" src="https://github.com/user-attachments/assets/0813b19f-1646-4117-bf74-cd9f7303d61b" />

```
CDVWebViewEngine shouldReloadWebView::
CDVWebViewEngine shouldReloadWebView title: Stress-Energy Tensor Inquiry - Ozwell
CDVWebViewEngine shouldReloadWebView location: http://localhost:12992/session/sjrSNPmbYmHHkyAvb
[...]

PR #1235, opened 2022-05-17
Fix #1232: Reload issue resulting in white screen

JH7:
Fixes #1232
Fixes meteor/meteor#11811

### Platforms affected

iOS

### Motivation and Context

If the iOS system kills the WKWebView processes due to e.g. memory issues, the Cordova app tries to reload the current URL on becoming foreground here:

https://github.com/apache/cordova-ios/blob/67b0bb2cfceb3f04fcd25a09222e86404805c594/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m#L530-L533

Due to client side routing, the current page may not be the correct initial URL (which includes an access token e.g.). Hence, some apps show a blank screen due to failure of reloading.

See https://github.com/apache/cordova-ios/issues/1232 and https://github.com/meteor/meteor/issues/11811 for an in-depth analysis of the problem and a feasible replication.

https://github.com/apache/cordova-ios/issues/1232 also indicates that the logic behind `shouldReloadWebView` could be an additional problem. @gouteru describes a further fix in https://github.com/apache/cordova-ios/issues/1232#issuecomment-1112812092 where they change this logic. As I found that the tests in

https://github.com/apache/cordova-ios/blob/67b0bb2cfceb3f04fcd25a09222e86404805c594/tests/CordovaLibTests/CDVWebViewEngineTest.m#L191-L193

explicitly state that an empty title should not result in a reload, I decided to not include they fix. But it could also be possible that the fix is required, as a result, the tests would have to be changed accordingly. If checking the title for 0 lengthiness is strictly required for your use case, please correct me @gouteru.

### Description

Since a reload is not sufficient, `CDVWebViewEngine.m` gets the initial app URL via `CDVViewController` (by making `appUrl` a public method) and initiates a new request with the initial app URL. The splashscreen is shown before the new request to hide a temporary white screen from the user.

### Testing

I used the changes with my Meteor/Cordova app on different iOS devices and I could not detect any white screen.

### Checklist
- [ ] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [x] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [ ] I've updated the documentation if necessary

Issue #1232, opened 2022-04-20
My app stops white blank screen  when app back to foreground from background


gouteru:
# Bug Report

Many of my app users has been reporting the issue that my app stops as white screen when it back to foreground from background.

# Problem

## What is expected to happen?

When back to foreground, app should display its webview contents properly.

## What does actually happen?

When back to foreground, app displays white blank screen and no way to control by user(it looks freezing).

# Information

I have been researching the issue a couple of months because I haven't reproduced.
Finally I could reproduce the issue by killing 2 webview processes with Acitivity Monitor of MacOS.

### Here’s the procedure.

1. Start app with XCode Simulator(ex: iPhone12 mini).
2. By using Simulator, get app switch to background. 
3. Kill 2 processes(com.apple.Webkit.networking, com.apple.Webkit.WebContent) of app by using Activity Monitor. Clicking “Forced termination“.  
4. Back the app to foreground. 
5. The app shows white screen and no way to control.

When the app is back to foreground, app shows white screen but the killed 2 processes are reloaded in Activity Monitor view. But those 2 processes seem never communicate with app main process properly.

I guess, cordova-ios detects that webview processes are terminated and do webview.reload() but after that the reloaded processes are not communicating with app main process. Then my Javascript code is not loaded, not working, not rendering anything,  so I can do nothing.

### Here’s the log of Xcode.

```
2022-04-19 21:56:05.574242+0900 Miley PTA[1372:4054372] app entered background
2022-04-19 21:56:05.574344+0900 Miley PTA[1372:4054372] Set state background
2022-04-19 21:56:05.574434+0900 Miley PTA[1372:4054372] Disconnected from FCM
2022-04-19 21:56:05.590935+0900 Miley PTA[1372:4054372] [Snapshotting] Snapshotting a view (0x7fd97b078a00, UIKeyboardImpl) that has not been rendered at least once requires afterScreenUpdates:YES.
2022-04-19 21:56:05.893567+0900 Miley PTA[1372:4054372] pause!
2022-04-19 21:58:28.746089+0900 Miley PTA[1372:4054372] [Process] 0x111aca330 - NetworkProcessProxy::didClose (Network Process 0 crash)
2022-04-19 21:58:28.746328+0900 Miley PTA[1372:4054372] [Process] 0x111af1600 - [PID=1511] WebProcessProxy::didClose: (web process 0 crash)
[...]

PR #1582, opened 2025-11-05
fix: Add preference for WKWebView crash recovery behaviour

dpogue:
### Platforms affected
iOS

### Motivation and Context

Provide a preference to control whether the WKWebView will refresh or reload completely after a crash.

Closes GH-1232.
Closes GH-1235.
Closes GH-1355.
Closes GH-1393.
Closes GH-1533.
Closes GH-1552.

### Description

This adds a `CrashRecoveryBehavior` preference with options `"refresh"` (default - try to refresh the WKWebView URL) and `"reload"` (load the starting page again in the WKWebView).

This doesn't solve the root issue where sometimes refreshing the WKWebView after a crash results in a blank screen, but gives people the option to force a full reload (which seems to be what people are using as a workaround for the blank screen bug).

### Testing

Added API tests to confirm the intended behaviour.

### Checklist

- [x] I've updated the documentation if necessary: https://github.com/apache/cordova-docs/pull/1440

### DIFF
--- a/tests/CordovaLibTests/CDVWebViewEngineTest.m
+++ b/tests/CordovaLibTests/CDVWebViewEngineTest.m
@@ -20,22 +20,17 @@
 #import <UIKit/UIKit.h>
 #import <XCTest/XCTest.h>
 #import "CDVWebViewEngine.h"
-#import "CDVWebViewProcessPoolFactory.h"
+#import <Cordova/CDVWebViewProcessPoolFactory.h>
 #import <Cordova/CDVSettingsDictionary.h>
 #import <Cordova/CDVAvailability.h>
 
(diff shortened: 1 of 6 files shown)


erisu (approved):
LGTM

Built a Cordova test app and tested the following cases:

- No settings: After crashing the WebContent, the app's WebView reloaded and the previously set hash value remained in the `href`.
- `CrashRecoveryBehavior` set to `refresh`: After crashing the WebContent, the app's WebView reloaded and the previously set hash value remained in the `href`.
- `CrashRecoveryBehavior` set to `reload`: After crashing the WebContent, the app's WebView reloaded and the `href` resets to the original starting URL.

erisu (approved):

PR #1583, opened 2025-11-08
feat!: drop ios-sim & use simctl directly

### LINKED ISSUE #567: advisory: don't upgrade to ios-sim 9.x
This is yet unreleased but ios-sim 9 will require changes to how cordova-ios uses it, since the exports are different, plus other bits.

erisu:
This PR is currently set as draft while waiting for `simctl@3.0.0` release.

### Motivation and Context

Drop `ios-sim` and use `simctl` directly.

Closes #567

### Description

Removed usage of `ios-sim`.  Fetching of iOS simulators and running on simulator logic to call `simctl` directly.

### Testing

- `npm t`
- built Cordova sample app.
- Ran `cordova run ios --list`
- Ran `cordova run ios`

### Checklist

### DIFF
--- a/tests/spec/unit/lib/list-emulator-images.spec.js
+++ b/tests/spec/unit/lib/list-emulator-images.spec.js
@@ -21,16 +21,27 @@
 // it ends up requiring the specific macOS environment bits that
 // allow for interacting with iOS Simulators. On Windows+Linux we are
 // bound to not-have-that.
-const list_emus = require('../../../../lib/listEmulatorImages');
+const fs = require('node:fs');
+const path = require('node:path');
 const simctl = require('simctl');
+const listEmulatorImages = require('../../../../lib/listEmulatorImages');
 
-describe('cordova/lib/listEmulatorImages', () => {
+let json;
+
+function fixtureJson (output) {
+    const file = path.resolve(__dirname, `../fixtures/${output}`);
+    return JSON.parse(fs.readFileSync(file, { encoding: 'utf-8' }).toString());
+}
+
+describe('listEmulatorImages', () => {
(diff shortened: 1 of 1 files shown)


tests/spec/unit/lib/list-emulator-images.spec.js
```diff
@@ -1,38 +1,38 @@
...
 // it ends up requiring the specific macOS environment bits that
 // allow for interacting with iOS Simulators. On Windows+Linux we are
 // bound to not-have-that.
 if (process.platform === 'darwin') {
     const list_emus = require('../../../../lib/listEmulatorImages');
-    const iossim = require('ios-sim');
+    const simctl = require('simctl');
 
-    describe('cordova/lib/listEmulatorImages', () => {
+    xdescribe('cordova/lib/listEmulatorImages', () => {
```
dpogue: For this test, I think we can remove the darwin check, require simctl unconditionally, and just mock the JSON response? Is there something else that causes this to fail?

tests/spec/unit/lib/simctlHelper.spec.js
```diff
@@ -0,0 +1,502 @@
...
+let json;
+let simctlHelper;
+let emitSpy;
+
+function fixtureJson (output) {
+    const file = path.resolve(__dirname, `../fixtures/${output}`);
+    return JSON.parse(fs.readFileSync(file, { encoding: 'utf-8' }).toString());
+}
+
+if (process.platform === 'darwin') {
```
dpogue: No need for these to be wrapped in a darwin check, since we're mocking the actual calls

dpogue (approved):

PR #1602, opened 2026-02-05
fix: add details for deprecation of `CDVWebViewProcessPoolFactory``

GitToTheHub:
- WKProcessPool is deprecated since iOS 15.0
- Add deprecation reason from Apple: `Creating and using multiple instances of WKProcessPool no longer has any effect.`

### Checklist

### DIFF
--- a/CordovaLib/include/Cordova/CDVWebViewProcessPoolFactory.h
+++ b/CordovaLib/include/Cordova/CDVWebViewProcessPoolFactory.h
@@ -22,8 +22,15 @@
 @class WKProcessPool;
 
 /**
- Apple deprecated the use of WKProcessPool since iOS 15.0 with the reason:
- Creating and using multiple instances of WKProcessPool no longer has any effect.
+   Apple has deprecated the WKProcessPool API, saying that it has no effect
+   in iOS 15 and newer. As such, the CDVWebViewProcessPoolFactory API is
+   marked as deprecated, but still exists to support iOS 13 and 14.
+
+   The CDVWebViewProcessPoolFactory API was also problematic because it
+   exposed WebKit-specific API types to the public API interface of Cordova,
+   potentially causing issues if those APIs need to change in the future. 
+   With this deprecation and eventual removal, Cordova is better insulated
+   from upstream WebView changes.
  @Metadata {
     @Available(Cordova, introduced: "6.2.0", deprecated: "8.0.0")
  }


GitToTheHub:
For me it was not obvious why `CDVWebViewProcessPoolFactory` is deprecated and also thought this is a deprecation by Apple since iOS 26. For plugin developers it's better to give them a hint why `CDVWebViewProcessPoolFactory` is deprecated and what's the reason behind it. As I understand, `CDVWebViewProcessPoolFactory` is deprecated, because Apple deprecated `WKProcessPool` since iOS 15 and has no effect anymore when using it. I made this more clear in the deprecation message.

dpogue (commented):
Note: Cordova users won't see Apple's deprecation warning because it only shows up if your minimum iOS version is higher than iOS 15.0 *and* you're building with the iOS 26 or newer SDK. Apple didn't actually mark WKProcessPool as deprecated until iOS 26, but retroactive marked it as having been deprecated in iOS 15.

So the whole thing is a confusing mess, and I want to avoid adding more confusion for plugin authors by just stating that Cordova has deprecated our CVDWebViewProcessPoolFactory API (with intent to remove it completely in the next major)

GitToTheHub:
> Note: Cordova users won't see Apple's deprecation warning because it only shows up if your minimum iOS version is higher than iOS 15.0 and you're building with the iOS 26 or newer SDK. Apple didn't actually mark WKProcessPool as deprecated until iOS 26, but retroactive marked it as having been deprecated in iOS 15.

Thanks for the explanation. I was already wondering, why I get no deprecation warning when `processPool` is set. I just tested it by setting `deployment-target` to `15.0` and get started to get the warning.

> So the whole thing is a confusing mess, and I want to avoid adding more confusion for plugin authors by just stating that Cordova has deprecated our CVDWebViewProcessPoolFactory API (with intent to remove it completely in the next major)

This is really confusing. As a plugin developer, if you want to resolve the deprecation warning, you have to anyway deal with it and must understand what's happening, or? Should we add your note, you mentioned here, to the plugin migration guide for cordova-ios 8? I would add this also to the `CVDWebViewProcessPoolFactory` class documentation. Is the comment for the class currently fine, or do you want no comment there? When you say `I want to avoid adding more confusion for plugin authors` do you  meant to remove the things you mentioned in the plugin migration guide for cordova-ios 8?

GitToTheHub:
> This is a very niche API, used only by a handful of plugins that provide WKWebView-based webview implementations or WKWebView-based in-app browser implementations.

This is true. I was thinking the same, that this should not affect many plugins and will be more Cordova intern. I will close this PR.

Issue #1615, opened 2026-02-25
Setting `deployment-target` in `plugin.xml` does not update `Package.swift` of `CordovaPlugins`


GitToTheHub:
# Bug Report

## Problem

When I only set the `deployment-target` via `plugin.xml`:

```xml
<config-file target="config.xml" parent="/*">
    <preference name="deployment-target" value="16.0" />
</config-file>
```
the version in `Package.swift` of `CordovaPlugins` stays at 13.0. I also noticed that the Target `App` stays on 13.0.

I first added cordova-ios (master branch) and then my plugin with the appropriate `plugin.xml`. When I do a `cordova prepare`, the `deployment-target` gets changed in Target `App` but not in `Package.swift`.

I use also CocoaPods for cordova-ios 7 via `nospm="true"`. ChatGPT mentioned that the Swift Package change is not reached in `prepare.js` when a PodFile is present. When I copy the code before the PodFile code, the change works on `cordova prepare` but not on `cordova plugin add`.

### What is expected to happen?

The version should be updated to 16.0 in `Package.swift` and also on the target `App`, if `deployment-target` is only set in `plugin.xml`.

## Information

Setting the `deployment-target` only in `plugin.xml` and do a `cordova plugin add` is also broken in cordova-ios 7, where the `deployment-target` is not set in the Target `Hello Cordova`. When doing a `cordova prepare ios`, the `deployment-target` is correctly updated in the Target.

## Checklist

- [ ] I searched for existing GitHub issues
- [ ] I updated all Cordova tooling to most recent version
- [ ] I included all the necessary information above

PR #1616, opened 2026-02-25
fix(spm): Ensure the deployment target always gets set

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes GH-1615.

### Description

The block of code to update the version in the Podfile contains a return statement, meaning that in some cases the Package.swift file was never getting updated. Reorder the code so it always updates the Package.swift first.

### Checklist

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -336,6 +336,12 @@ function handleBuildSettings (platformConfig, locations, infoPlist) {
 
     project.write();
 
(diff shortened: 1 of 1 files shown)


GitToTheHub:
> cordova platform add and cordova plugin add are responsible for adding platforms and plugins, not for getting the project into a buildable state.

This is not true. As I noted I always worked with just only `cordova platform add` and `cordova plugin add` and get a buildable state. Test it by yourself, just do `cordova create hello-world && cd hello-world && cordova platform add ios` and open it on XCode. It will run and contains everything from the www folder.

Issue #1617, opened 2026-02-26
LIBRARY_SEARCH_PATHS uses PRODUCT_NAME instead of "App" folder


rajinikanth0601:
# Bug Report

### What is expected to happen?
When a plugin declares `<source-file src="libfoo.a" framework="true"/>`, the generated `LIBRARY_SEARCH_PATHS` in the Xcode project should reference the actual project folder `$(SRCROOT)/App/Plugins/...`.

### What does actually happen?
`LIBRARY_SEARCH_PATHS` is set to `$(SRCROOT)/<PRODUCT_NAME>/Plugins/...` (e.g., `$(SRCROOT)/MyApp/Plugins/...`). Since cordova-ios 8.0.0 always uses `App/` as the project folder but `PRODUCT_NAME` is derived from `config.xml`'s `<name>`, the path doesn't exist on disk and the linker fails with `library not found`.

## Information
cordova-ios 8.0.0 changed the project structure to always use `App/` as the project folder name. However, when processing `<source-file framework="true"/>` plugin entries, the `xcode` npm module's `searchPathForFile()` function (`pbxProject.js:1704`) constructs the library search path using `proj.productName`:

`return '"\\"$(SRCROOT)/' + proj.productName + fileDir + '\\""';
`

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

PR #1618, opened 2026-02-26
fix(xcode): Fix library search paths for target

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes GH-1617.

### Description

Ensure we add a library search path that uses the target name instead of the product name. For now, we continue to add the one with the product name (which is what node-xcode does by default).

### Testing

Added unit test to confirm correct path value.

### Checklist


dpogue:
I've tested this with a plugin that uses `<source-file framework="true">` and confirmed that the correct `LIBRARY_SEARCH_PATHS` are set to include the App/Plugins/${plugin.id} path

It does produce a warning for the (non-existent) search path that uses the product name... maybe it's best to just remove that?

PR #1628, opened 2026-03-08
chore: Fix missing licence headers

dpogue:
### Platforms affected
iOS

### Description

Ensure all project files have a proper licence header.

### DIFF
--- a/.rat-excludes
+++ b/.rat-excludes
@@ -17,22 +17,10 @@
 
 # This file used by ATR
 
-**/*.txt
-
 .git/
-
-# yes, not .gitignore
-templates/project/App/gitignore
-
-# Other .gitignore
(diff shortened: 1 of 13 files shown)


erisu (approved):
LGTM

If possible, I would like to cleanup the rat ignore files a bit more:

Both files can be identical with the following:

```
.git/
coverage/
node_modules/

# Files that do not support comments
**/*.txt
**/*.xctestplan
```

The `.ratignore` has defined `**/*.txt` twice.

The `.rat-exclude` has `tests/spec/fixtures/` which wont matter once this PR is merged, which updated the last missing files.
It also has `tests/CordovaLibTests/CordovaTests.xctestplan` but will be replaced with line `**/*.xctestplan`

Issue #419, opened 2018-09-27
Why do we depend on a global ios-deploy?


oliversalzburg:
Why can't ios-deploy be installed locally?

oliversalzburg:
@janpio Yes, I think there were issues in the past, but those should really *be in the past*. Executable npm modules should all register entrypoints in `node_modules/.bin/` and should generally be easily invokable from there. If everything else fails, I think [npx](https://www.npmjs.com/package/npx) is widely accepted as a solution in this regard.

janpio:
Afaik Cordova support much older versions of node/npm than those that first added a stable version of `npx`.

But in general it would of course be better, to have a local dependency on `ios-deploy` than the current state. Could also get rid of that code that gives you the error message then. I think we would be very happy in merging a PR that changes this.

brodybits:
There were 2 reasons we did not explicitly depend on `ios-deploy` in the past ([[1]](https://lists.apache.org/thread.html/8f01ff201d46dce3fadec5479fd715b96f8b366aac38b4dd2cdfe411@%3Cdev.cordova.apache.org%3E)):
* viral license
* possible Xcode version mismatch

[1] <https://lists.apache.org/thread.html/8f01ff201d46dce3fadec5479fd715b96f8b366aac38b4dd2cdfe411@%3Cdev.cordova.apache.org%3E>

brodybits:
Viral license was especially bad in the past when "we" had to include dependencies in the bundle. (I was not part of the "we" before 2018.)

This should no longer be necessary in next major release where we are dropping Node.js 4 support. But it may still be an issue to add such a dependency in an Apache project. I would also personally downvote such an explicit dependency due to the viral license.

Sorry I took so long to compose the response.

oliversalzburg:
@brodybits You would be opposed to adding an ios-deploy dependency to cordova because ios-deploy uses a problematic license? Why? And why is it then fine to still rely on it?

PR #1636, opened 2026-03-13
feat(devicectl): Use devicectl for deploying to devices

### LINKED ISSUE #429: Use node-ios-device to deploy
<https://github.com/appcelerator/node-ios-device> - with APL

Thanks to @oliversalzburg for pointing it out on Slack.

### LINKED ISSUE #1594: Device deploy fails on Xcode 26 / iOS 26: ios-deploy requires DeveloperDiskImage.dmg (DDI-only)
# Bug Report
With Xcode 26 and iOS 26 devices, `cordova run ios --device` fails because `ios-deploy`
expects `DeveloperDiskImage.dmg` in DeviceSupport, but iOS 26 no longer ships that
image (Xcode uses DDI-only).

Deploy to a connected iOS device should install and launch the app (`cordova run ios --device`) using the available Xcode device support.

`cordova-ios` calls `ios-deploy`, which fails with:
Unable to locate DeviceSupport directory with suffix 'DeveloperDiskImage.dmg'
Command failed with exit code 253: ios-deploy --justlaunch -d -b <app> --no-wifi

Even with Xcode 26 and device support installed, `DeveloperDiskImage.dmg` no longer exists for iOS 26 (DDI-only). Deploy to real device fails.

### Command or Code
cordova run ios --device

## Additional info / workaround
Using `xcrun devicectl` works:
xcrun devicectl device install app --device "<device>" <path-to-app>
xcrun devicectl device process launch --device "<device>" --terminate-existing <bundle-id>

Suggestion: replace ios-deploy usage with `devicectl` on Xcode 26+, or fallback to devicectl when `DeveloperDiskImage.dmg` is missing.

[run.js](https://github.com/user-attachments/files/24291005/run.js)

## Checklist

dpogue:
### Platforms affected
iOS

### Motivation and Context

`ios-deploy` does not properly handle deploying to newer iOS versions, and Xcode now includes a built-in `devicectl` tool for this purpose (with the caveat that it does not handle deploying to older iOS versions <17).

This also allows us to slowly transition away from requiring a global installation of the `ios-deploy` tool, whose licence prevents its direct inclusion or distribution with cordova-ios.

Ref https://github.com/apache/cordova-ios/issues/419
Closes https://github.com/apache/cordova-ios/issues/429
Closes https://github.com/apache/cordova-ios/issues/1594

### Description

Bring in a new [`devicectl`](https://github.com/ios-control/devicectl) library to support using `devicectl` to install and launch apps on connected devices. We keep `ios-deploy` as a fallback.

The format of device target identifiers is different between `devicectl` and `ios-deploy` namely in that the UDID from `devicectl` includes a dash. Currently we use that dash to determine (given a target ID) whether to invoke `devicectl` or `ios-deploy` but I'd like to find a way to avoid that. It would be nice to be able to provide something like `--target="My iPhone"` (which `devicectl` will happily support) instead of `--target="lotsOfNumbers"`.

### Testing

Manual testing deploying an app to a connected iOS 26 device.

Needs more unit tests.

### Checklist


erisu (approved):
LGTM

Tested:

- `npm t`
- `platform add`
- `build ios`
- `run ios`
  - Confirmed that it called `devicectl.install`

GitToTheHub:
Is this PR good to go?

dpogue:
This still needs unit tests written for all the run.js stuff, and more actual testing with devices

GitToTheHub:
Ok

PR #1606, opened 2026-02-23
fix(webview): Ensure scheme task is always finished

dpogue:
### Platforms affected
iOS

### Motivation and Context

Closes GH-1605.

### Description

Ensure that we always mark a scheme task that we are handling as being completed when it is cancelled.

### Checklist

- [ ] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))

PR #1640, opened 2026-03-19
fix: `NSInternalInconsistencyException: "No response has been sent for this task"` - second attempt

### LINKED ISSUE #1605: NSInternalInconsistencyException: "No response has been sent for this task" — race condition in CDVURLSchemeHandler in 8.0.0
# Bug Report

## Problem
App crashes happening due to race condition in CDVURLSchemeHandler
URL scheme tasks started by CDVURLSchemeHandler should always receive a response - without crashing the app.
This worked correctly in cordova-ios 7.1.1.

After upgrading from cordova-ios 7.1.1 to 8.0.0, the app crashes on startup with:
```
Fatal Exception: NSInternalInconsistencyException
No response has been sent for this task
CDVURLSchemeHandler.m:169
__50-[CDVURLSchemeHandler webView:startURLSchemeTask:]_block_invoke
```

## Information
I am running a "hosted web app" pattern where the local page hosted by WKWebView quickly navigates to an external url using `window.location.replace`. I believe this app crash happens due to a race condition in `CDVURLSchemeHandler webView:startURLSchemeTask` and switching to the external URL.

In [7.1.1](https://github.com/apache/cordova-ios/blob/rel/7.1.1/CordovaLib/Classes/Public/CDVURLSchemeHandler.m) The file read and response is sent back to WebKit in the same call, all on the main thread.

In [8.0.0](https://github.com/apache/cordova-ios/blob/rel/8.0.0/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVURLSchemeHandler.m), this is done in a background thread `    [self.viewController.commandDelegate runInBackground:^{`.
I suspect when WebKit is cancelling the task due to url change, there is a chance of no response being returned:

1. startURLSchemeTask: is called — the task is added to handlerMap and the background block is enqueued (but not yet running)
2. The web view navigates away (e.g. window.location.replace() to an external URL), or the page simply cancels the request
3. WebKit calls stopURLSchemeTask:, which removes the task from handlerMap
4. The background block finally executes — every taskActive: check returns NO because the task was removed from the map
5. No response is ever sent (didReceiveResponse:, didFailWithError:, and didFinish are all skipped)
6. WebKit throws NSInternalInconsistencyException: "No response has been sent for this task"

- cordova-ios: 8.0.0 (worked on 7.1.1)
- cordova: 13.0.0
- iOS: 17+
- Device: iPhone (reproducible on both physical devices and simulators)

### Version information

- cordova 13.0.0
- cordova-ios 8.0.0

## Checklist

### LINKED ISSUE #1638: NSInternalInconsistencyException "This task has already been stopped" in CDVURLSchemeHandler (regression from PR #1606)
### Problem

The fix for #1605 (PR #1606, commit a706334) introduced a new crash by calling `[urlSchemeTask didFinish]` inside `webView:stopURLSchemeTask:`. Per Apple's WKURLSchemeTask documentation, calling any methods on a scheme task after `webView:stopURLSchemeTask:` is invoked throws an `NSInternalInconsistencyException`  "This task has already been stopped". [[LINK](https://developer.apple.com/documentation/webkit/wkurlschemetask/didfinish())]

When WebKit calls `webView:stopURLSchemeTask:`, CDVURLSchemeHandler should clean up its internal state without calling any methods on the `urlSchemeTask` object.

The app crashes with:
-[CDVURLSchemeHandler webView:stopURLSchemeTask:]
NSInternalInconsistencyException - This task has already been stopped

### Root Cause

In the current `CDVURLSchemeHandler.m` (after PR #1606), `stopURLSchemeTask:` calls `[urlSchemeTask didFinish]` for tasks owned by CDVURLSchemeHandler:

```
- (void)webView:(WKWebView *)webView stopURLSchemeTask:(id<WKURLSchemeTask>)urlSchemeTask
{
    CDVPlugin *plugin;
    @synchronized(self.handlerMap) {
        plugin = [self.handlerMap objectForKey:urlSchemeTask];
    }

    @synchronized(self.handlerMap) {
        [self.handlerMap removeObjectForKey:urlSchemeTask];
    }

    if ([plugin isEqual:[NSNull null]]) {
        [urlSchemeTask didFinish]; 
    } else if ([plugin respondsToSelector:@selector(stopSchemeTask:)]) {
        [plugin stopSchemeTask:urlSchemeTask];
    }
}
```

### Potential fix
It seems like there is a [fix ](https://github.com/ionic-team/capacitor/pull/7482/files)for a very similar issue in capacitor.
A potential safeguard could be to wrap task method calls with try/catch.

cordova-ios: master branch commit a706334
cordova: 13.0.0
iOS: 17+
Device: iPhone (physical devices and simulators)
Rolled back to cordova-ios 7.1.1 as a workaround

GitToTheHub:
### Platforms affected
iOS

### Motivation and Context

- This fixes #1605 by ensuring WKURLSchemeTask callbacks are dispatched on the main thread to prevent race conditions in `webView:startURLSchemeTask:` and also:
  - Resolve file errors and range validation before background execution
  - Send the URL scheme response before queuing background reads
  - Exit early when task is no longer active and close the file handle
- This PR reverts #1606 which was an attempt to fix this, but has created another exception `NSInternalInconsistencyException - This task has already been stopped`, because `[WKURLSchemeTask didFinish]` cannot be called in `webView:stopURLSchemeTask:`, which is documented by Apple: https://developer.apple.com/documentation/webkit/wkurlschemetask/didfinish()?language=objc
  - Fixes https://github.com/apache/cordova-ios/issues/1638
- Added tests
  - `testStartURLSchemeTaskSendsResponseBeforeBackgroundWork`
  - `testStopURLSchemeTaskDoesNotFinishStoppedTask`
  - `testMissingFileFailsBeforeBackgroundWork`
- Generated-By: GPT-5.3-Codex, GitHub Copilot Chat
- Generated-By: Claude Sonnet 4.6, GitHub Copilot Chat

### Testing

- I tested it on an iOS 18.5 and 26.4 Simulator on an production app by running rapidly multiple times the app on XCode by stopping and running. There were no load issues. This PR have to be tested, if it fixes #1605.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVURLSchemeHandler.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVURLSchemeHandler.m
@@ -150,33 +150,41 @@ static const NSUInteger FILE_BUFFER_SIZE = 1024 * 1024 * 4; // 4 MiB
         NSError *readError;
         NSUInteger responseSent = 0;
 
-        while ([self taskActive:urlSchemeTask] && responseSent < [responseSize unsignedIntegerValue]) {
+        // Read file data on the background thread and dispatch each chunk to the
+        // main thread. Dispatching to the main thread serialises all
+        // WKURLSchemeTask callbacks with webView:stopURLSchemeTask:, which is
+        // also called on the main thread. This eliminates the race condition
+        // between the taskActive check and the actual callback invocation.
+        while (responseSent < [responseSize unsignedIntegerValue]) {
             @autoreleasepool {
                 NSData *data = [self readFromFileHandle:fileHandle upTo:FILE_BUFFER_SIZE error:&readError];
-                if (!data || readError) {
-                    if ([self taskActive:urlSchemeTask]) {
-                        [urlSchemeTask didFailWithError:readError];
-                    }
+                if (!data || readError || data.length == 0) {
                     break;
                 }
-
-                if ([self taskActive:urlSchemeTask]) {
-                    [urlSchemeTask didReceiveData:data];
-                }
-
                 responseSent += data.length;
+
+                dispatch_async(dispatch_get_main_queue(), ^{
+                    if ([self taskActive:urlSchemeTask]) {
+                        [urlSchemeTask didReceiveData:data];
+                    }
+                });
             }
         }
 
         [fileHandle closeFile];
+        NSError *capturedError = readError;
 
-        if ([self taskActive:urlSchemeTask]) {
-            [urlSchemeTask didFinish];
-        }
+        dispatch_async(dispatch_get_main_queue(), ^{
+            if (capturedError && [self taskActive:urlSchemeTask]) {
+                [urlSchemeTask didFailWithError:capturedError];
+            } else if ([self taskActive:urlSchemeTask]) {
+                [urlSchemeTask didFinish];
+            }
 
-        @synchronized(self.handlerMap) {
-            [self.handlerMap removeObjectForKey:urlSchemeTask];
-        }
+            @synchronized(self.handlerMap) {
+                [self.handlerMap removeObjectForKey:urlSchemeTask];
+            }
+        });
     }];
 }
 
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVURLSchemeHandler.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVURLSchemeHandler.m
@@ -183,10 +183,9 @@ static const NSUInteger FILE_BUFFER_SIZE = 1024 * 1024 * 4; // 4 MiB
         [self.handlerMap removeObjectForKey:urlSchemeTask];
     }
 
-    if ([plugin isEqual:[NSNull null]]) {
-        // NSNull means we own this task, so we need to mark it as finished
-        [urlSchemeTask didFinish];
-    } else if ([plugin respondsToSelector:@selector(stopSchemeTask:)]) {
+    // After stopURLSchemeTask is invoked, WKURLSchemeTask must not receive
+    // additional callbacks from this handler.
+    if (![plugin isEqual:[NSNull null]] && [plugin respondsToSelector:@selector(stopSchemeTask:)]) {
         [plugin stopSchemeTask:urlSchemeTask];
     }
 }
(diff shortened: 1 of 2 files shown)


GitToTheHub:
@dpogue Would this be right approach? I read the documentation of [[WKURLSchemeTask didFinish]](https://developer.apple.com/documentation/webkit/wkurlschemetask/didfinish()?language=objc) where it says:
> It also raises an exception if you call it after WebKit calls the webView:stopURLSchemeTask: method of the corresponding handler object.

dpogue:
This is essentially reverting the change from https://github.com/apache/cordova-ios/pull/1606 which was an attempt to fix a different NSInternalInconsistencyException from https://github.com/apache/cordova-ios/issues/1605.

It's possible this is the right approach for a stopped task, but then we need to better understand the original issue and how to prevent that error.

GitToTheHub:
I tried to create a fix with AI based on #1605 where some things are not executed any more on the background thread, but am not sure if this brings new errors. I tested the changes on an iOS 26.4 simulator in a production app. I get one time a failure in my JavaScript that an Object could not be found and one time that an image could not be loaded. This happend after some runs in XCode. If this changes are here are not helpfull I will close the PR.

GitToTheHub:
Hi @dpogue, I dug deeper in this issue and the solution seems to be to call all WKURLSchemeTask callbacks on the main thread to prevent race conditions. I ran multiple times a production app on XCode on an iOS 18.5 and iOS 26.4 simulator and got no issues. Now it needs to verify if this fixes the original issue on #1605. Do you think these changes are good?

GitToTheHub:
Test it with a production app and had no issues.

NityaSantosh26:
@GitToTheHub
## Issue still reproducible after PR changes

I can confirm this crash still occurs even with the latest changes from this PR applied.

**Environment**
- `cordova-ios`: built from latest `git+https://github.com/apache/cordova-ios.git`
- Previous version: `cordova-ios@8.0.1`
- Platform: iOS (real device — iPhone 14, iOS 26.4.1)

**How to reproduce**
1. `cordova platform add git+https://github.com/apache/cordova-ios.git`
2. Build and install the app
3. Launch the app → logs in successfully ✅
4. Swipe-close the app
5. Re-open the app → crashes immediately, keeps flashing and cannot be opened ❌

**Exception**
```
NSInternalInconsistencyException:
"No response has been sent for this task"
```

**Crash stack (Thread — com.apple.root.default-qos)**
```
+[NSException raise:format:]
-[WKURLSchemeTaskImpl didFailWithError:]
__50-[CDVURLSchemeHandler webView:startURLSchemeTask:]_block_invoke  CDVURLSchemeHandler.m:83
_dispatch_call_block_and_release
_dispatch_queue_override_invoke
_dispatch_root_queue_drain
_pthread_wqthread
```

**Observation**
The crash only occurs on re-launch after swipe-close. A fresh install always succeeds on first launch. This points to the race condition between `stopURLSchemeTask:` and the background block still not being fully guarded — the task appears to be cancelled by WebKit on resume before the queued block completes.

GitToTheHub:
Hi, I tried it on a real device and simulator and couldn't reproduce your issue. I used XCode 26.4.1 to run the app. What do you mean with `Swipe-close the app`? Do you mean to put it in the background by swiping from bottom to top and open it again, or do you mean to swipe slowly from bottom to top to see a all open apps and swipe your app away so it's completely closed? I tried both and hadn't any issues.

Could you create a reproducible sample app repository?

PR #1643, opened 2026-03-31
fix: dont call nativeFetchMessages if not implemented

erisu:
### Motivation and Context

Don't call `nativeFetchMessages` if it's not implemented.

### Description

`execProxy.nativeFetchMessages` appears to call `cordovaExec().nativeFetchMessages` which is undefined as it was removed when UIWebView was dropped in PR: #773

It appears that some plugins might still rely on calling `execProxy.nativeFetchMessages` so this PR will wrap and check for the existence of `cordovaExec().nativeFetchMessages` before attempting to call it.

Those plugins would most likely need to figure out how to re-implement `cordovaExec().nativeFetchMessages` for it to work.

### Testing

- `npm t`
- `cordova platform add`
- `cordova build`
- Run default Cordova app in simulator

### Checklist

### DIFF
--- a/cordova-js-src/exec.js
+++ b/cordova-js-src/exec.js
@@ -134,17 +134,13 @@ iOSExec.nativeEvalAndFetch = function (func) {
 
 function cordovaExec () {
     var cexec = require('cordova/exec');
-    var cexec_valid = (typeof cexec.nativeFetchMessages === 'function') && (typeof cexec.nativeEvalAndFetch === 'function') && (typeof cexec.nativeCallback === 'function');
+    var cexec_valid = (typeof cexec.nativeEvalAndFetch === 'function') && (typeof cexec.nativeCallback === 'function');
     return (cexec_valid && execProxy !== cexec) ? cexec : iOSExec;
 }
 
 function execProxy () {
     cordovaExec().apply(null, arguments);
 }
-
-execProxy.nativeFetchMessages = function () {
-    return cordovaExec().nativeFetchMessages.apply(null, arguments);
-};
 
 execProxy.nativeEvalAndFetch = function () {
     return cordovaExec().nativeEvalAndFetch.apply(null, arguments);
--- a/cordova-js-src/exec.js
+++ b/cordova-js-src/exec.js
@@ -143,7 +143,14 @@ function execProxy () {
 }
 
 execProxy.nativeFetchMessages = function () {
-    return cordovaExec().nativeFetchMessages.apply(null, arguments);
+    // The built-in WKWebView-based WebView does not implement `nativeFetchMessages`
+    // since it uses `postMessage` instead.
+    // The `nativeFetchMessages` logic is kept guarded in case a custom pluggable
+    // WebView engine still implements it. This logic is apart of the old JS/native
+    // command queue dispatching/polling path.
+    if (cordovaExec().nativeFetchMessages) {
+        return cordovaExec().nativeFetchMessages.apply(null, arguments);
+    }
 };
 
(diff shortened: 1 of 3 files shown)


GitToTheHub:
Is this PR good to go?

dpogue:
Removing this (as the PR currently stands) requires a major version bump because it removes public API and potentially breaks third-party WebView plugins that call the ObjC method, so I think we want to be careful about what we do here.

GitToTheHub:
Ok understand, thanks for the clarification. But anyway this PR is ok for the next major, or not?

dpogue:
What is the actual reason for removing this? We thought it was unused, but it turns out to be used by plugins, so we shouldn't remove it without a compelling reason.

dpogue (approved):

cordova-js-src/exec.js
```diff
@@ -143,7 +143,9 @@ function execProxy () {
 }
 
 execProxy.nativeFetchMessages = function () {
-    return cordovaExec().nativeFetchMessages.apply(null, arguments);
+    if (cordovaExec().nativeFetchMessages) {
```
GitToTheHub, line 151: You write in your description the reason why checking for `cordovaExec().nativeFetchMessages` before calling it:

```markdown
execProxy.nativeFetchMessages appears to call cordovaExec().nativeFetchMessages which is undefined as it was removed when UIWebView was dropped in PR: https://github.com/apache/cordova-ios/pull/773

It appears that some plugins might still rely on calling execProxy.nativeFetchMessages so this PR will wrap and check for the existence of cordovaExec().nativeFetchMessages before attempting to call it.

Those plugins would most likely need to figure out how to re-implement cordovaExec().nativeFetchMessages for it to work.
```

For me it would not be clear why this line of code exists. Maybe you could add your description as documentation.

GitToTheHub (changes requested):

cordova-js-src/exec.js
dpogue: `execProxy.nativeFetchMessages` gets called from public API of CDVCommandQueue's `fetchCommandsFromJs`: https://github.com/apache/cordova-ios/blob/c20cdc93b77fc69720614122776d5ce27ea4f75e/CordovaLib/Classes/Public/CDVCommandQueue.m#L94-L109

That method isn't used by Cordova iOS itself, but was previously used by the UIWebView implementation, and is still available for plugins to call.

In particular, there are 3rd party plugins that provide UIWebView as a pluggable webview implementation (which is perfectly valid for enterprise apps that don't go through the AppStore), and they still rely on this method: https://github.com/bpresles/cordova-plugin-uiwebview-engine/blob/1fb24fd2569745c281c6f636bb9ae2bae769e746/src/ios/BPRUIWebViewNavigationDelegate.m#L111

They also provide the JavaScript implementation for the `nativeFetchMessages` method: https://github.com/bpresles/cordova-plugin-uiwebview-engine/blob/1fb24fd2569745c281c6f636bb9ae2bae769e746/src/www/ios/ios-uiwebview-exec.js#L201-L214

dpogue: So the implementation in this PR is the correct fix, to check if the WebView JS bridge implements the method before calling it, so that `[commandQueue fetchCommandsFromJs]` doesn't cause JS errors.

GitToTheHub: I just wanted to suggest to add documentation to this line of code. Also what you wrote would be good to add to `[CDVCommandQueue's fetchCommandsFromJs]` as documentation.

GitToTheHub (approved):

PR #1652, opened 2026-05-16
fix(actions): IPhone 16e not found on macOS 26 for latest OS

GitToTheHub:
### Platforms affected

- iOS

### Motivation and Context

- xcodebuild is called with destination: platform=iOS Simulator, name=iPhone 16e. Xcode interprets that as OS:latest, so it looks for iPhone 16e on the latest simulator runtime. On macos-26, latest runtime is 26.4.1, and there is no iPhone 16e for 26.4.1 (only iPhone 17e, iPhone 17, etc.). Therefore xcodebuild exits with: Unable to find a device matching the provided destination specifier npm exits with code 70 because npm run test:xcode failed.
- iPhone 17 is now used for macOS 26 and iPhone 16e otherwise
- Generated-By: GPT-5.3-Codex

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
             node-version: 24.x
 
     env:
-      CDV_IOS_SIM: iPhone 16e
+      CDV_IOS_SIM: ${{ matrix.os == 'macos-26' && 'iPhone 17' || 'iPhone 16e' }}
 
     steps:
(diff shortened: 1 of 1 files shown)


GitToTheHub:
All actions run now successfully, is this good to go?

dpogue (approved):
This works. We can probably drop the macos-15 runner at some point since macos-26 is stable now.

PR #1655, opened 2026-05-18
fix: forward NSUserActivity to plugins via CDVSceneDelegate

### LINKED ISSUE #1654: Regression: Universal Links from SFSafariViewController do not return to app on cordova-ios 8 due to SceneDelegate handling
### Issue Summary

Universal Links do not return to the Cordova app when tapped inside SFSafariViewController on `cordova-ios@8`.

### Environment

- Cordova CLI: 13
- cordova-ios: 8.x (using latest Git source due to blank screen / about:blank issue)
- Ionic CLI: 6.20.9
- Ionic Framework: v1 (AngularJS)
- Plugins:
  - cordova-plugin-safariviewcontroller@2.0.0
  - cordova-plugin-universallinks-fix@2.0.2 (based on CrowdhubApps implementation)
- iOS: 16.4.1
- Xcode: 26.5
- Xcode build: 17F42
- Device: real device

### Steps to Reproduce

1. Create a Cordova iOS project with `cordova-ios@8`
2. Install:
   - cordova-plugin-universallinks-fix@2.0.2
   - cordova-plugin-safariviewcontroller@2.0.0
3. Configure Universal Links (Associated Domains / AASA)
4. Open a page using SFSafariViewController
5. Tap a Universal Link pointing back to the same app

### Expected Behavior

The Universal Link should return to the app and be handled correctly.

### Actual Behavior

The Universal Link does not return to the app when tapped inside SFSafariViewController.

### Regression / Comparison

- On `cordova-ios@7`, Universal Links worked correctly (using a different plugin implementation)
- On `cordova-ios@8`, Universal Links are not delivered to the app in this scenario

### Notes

- The same Universal Link works correctly when opened from Slack
- I have not confirmed whether it works from Safari
- The issue occurs when the link is tapped inside SFSafariViewController
- This report is separate from the known `about:blank` / blank screen issue in cordova-ios@8

### Cause

The issue appears to be related to SceneDelegate handling in `cordova-ios@8`.

`NSUserActivity` (Universal Link) events received in SceneDelegate are not propagated to the Cordova AppDelegate layer, so plugins cannot receive them.

### Workaround / Fix

Forward `NSUserActivity` from SceneDelegate to AppDelegate:

```swift
import Cordova
import UIKit

class SceneDelegate: CDVDelegate {

    override func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
        if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
            _ = appDelegate.application(
                UIApplication.shared,
                continue: userActivity,
                restorationHandler: { _ in }
            )
        }
    }
}

GitToTheHub:
## Summary

`CDVSceneDelegate` was missing an implementation of `scene:continueUserActivity:`, which is the `UIWindowSceneDelegate` method iOS calls when an `NSUserActivity` event (i.e. a Universal Link with `NSUserActivityTypeBrowsingWeb`) is delivered to the active scene.

As a result, Universal Links were silently dropped when tapped inside `SFSafariViewController` on cordova-ios 8.x. On cordova-ios 7.x this worked because the activity was delivered to `AppDelegate.application(_:continue:restorationHandler:)` directly, before SceneDelegate was introduced.

## Changes

- **`CDVSceneDelegate.m`** — add `scene:continueUserActivity:` implementation that posts `CDVPluginContinueUserActivityNotification` via `NSNotificationCenter`, consistent with the existing URL-handling pattern.
- **`CDVPluginNotifications.h`** — declare the new `CDVPluginContinueUserActivityNotification` constant.
- **`CDVPlugin.m`** — define the new notification constant.

Plugins can now observe `CDVPluginContinueUserActivityNotification` to receive Universal Links and other `NSUserActivity` events.

Fixes #1654

Generated-By: Claude Sonnet 4.6

### DIFF
--- a/CordovaLib/Classes/Public/CDVPlugin.m
+++ b/CordovaLib/Classes/Public/CDVPlugin.m
@@ -44,6 +44,7 @@
 
 const NSNotificationName CDVPageDidLoadNotification = @"CDVPageDidLoadNotification";
 const NSNotificationName CDVPluginHandleOpenURLNotification = @"CDVPluginHandleOpenURLNotification";
+const NSNotificationName CDVPluginContinueUserActivityNotification = @"CDVPluginContinueUserActivityNotification";
 const NSNotificationName CDVPluginHandleOpenURLWithAppSourceAndAnnotationNotification = @"CDVPluginHandleOpenURLWithAppSourceAndAnnotationNotification";
 const NSNotificationName CDVPluginResetNotification = @"CDVPluginResetNotification";
 const NSNotificationName CDVViewWillAppearNotification = @"CDVViewWillAppearNotification";
--- a/CordovaLib/Classes/Public/CDVSceneDelegate.m
+++ b/CordovaLib/Classes/Public/CDVSceneDelegate.m
@@ -47,4 +47,9 @@
     }
 }
 
+- (void)scene:(UIScene *)scene continueUserActivity:(NSUserActivity *)userActivity
+{
+    [[NSNotificationCenter defaultCenter] postNotificationName:CDVPluginContinueUserActivityNotification object:userActivity];
+}
+
 @end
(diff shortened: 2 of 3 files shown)


GitToTheHub:
Works, when the plugin observes the new `CDVPluginContinueUserActivityNotification` notification.

PR #1661, opened 2026-06-03
fix(statusbar): Support setting alpha from the JS API

### LINKED ISSUE #1659: `window.statusbar.setBackgroundColor` don't support alpha
# Bug Report

## Problem

When setting a statusbar background color with alpha, like:

```js
window.statusbar.setBackgroundColor('#AAFFFFFF');
```

only the first 6 hex chars will be used. On JavaScript side everything is ok:

But the alpha is not used in `[CDVStatusBarInternal setBackgroundColor:]` but set to `1`:

https://github.com/apache/cordova-ios/blob/3d52d7b3b787292847dc3229b3a26e5ee3191696/CordovaLib/Classes/Private/Plugins/CDVStatusBarInternal/CDVStatusBarInternal.m#L35-L43

## Checklist

- [ ] I searched for existing GitHub issues
- [ ] I updated all Cordova tooling to most recent version
- [ ] I included all the necessary information above

dpogue:
### Platforms affected
iOS

### Motivation and Context

- Closes GH-1659
- The supported format is any valid [CSS color](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value) value, for e.g. `rebeccapurple`, `#RRGGBBAA`, `rgb(255 0 153)`

### Description

Use the alpha value from JS for the status bar background if it is provided, otherwise fallback to opaque (matching the current behaviour)

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVStatusBarInternal/CDVStatusBarInternal.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVStatusBarInternal/CDVStatusBarInternal.m
@@ -37,8 +37,9 @@
     NSInteger valueR = [[command argumentAtIndex:0 withDefault:@0] integerValue];
     NSInteger valueG = [[command argumentAtIndex:1 withDefault:@0] integerValue];
     NSInteger valueB = [[command argumentAtIndex:2 withDefault:@0] integerValue];
+    float valueA = [[command argumentAtIndex:3 withDefault:@1.f] floatValue];
 
-    UIColor *bgColor = [UIColor colorWithRed:valueR/255.f green:valueG/255.f blue:valueB/255.f alpha:1.f];
+    UIColor *bgColor = [UIColor colorWithRed:valueR/255.f green:valueG/255.f blue:valueB/255.f alpha:valueA];
     [self.viewController setStatusBarBackgroundColor:bgColor];
 }
 
--- a/cordova-js-src/plugin/ios/statusbar.js
+++ b/cordova-js-src/plugin/ios/statusbar.js
@@ -64,7 +64,7 @@ Object.defineProperty(statusBar, 'setBackgroundColor', {
             return;
         }
 
-        var rgbVals = rgbStr.match(/\d+/g).map(function (v) { return parseInt(v, 10); });
+        var rgbVals = rgbStr.match(/[\d.]+/g).map(function (v, i) { return (i < 3) ? parseInt(v, 10) : parseFloat(v); });
         if (rgbVals.length < 3) {
             return;
         }
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -26,7 +26,8 @@ module.exports = defineConfig([
     globalIgnores([
         '**/coverage/',
         'templates/project/',
-        'tests/spec/fixtures/'
+        'tests/spec/fixtures/',
+        'tests/CordovaLibTests/'
     ]),
     ...nodeConfig,
     ...nodeTestConfig.map(config => ({
--- a/tests/CordovaLibTests/CDVStatusBarTests.m
+++ b/tests/CordovaLibTests/CDVStatusBarTests.m
@@ -1,0 +1,184 @@
+/*
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+#import <UIKit/UIKit.h>
+#import <XCTest/XCTest.h>
+#import <Cordova/Cordova.h>
+#import "CDVWebViewEngine.h"
+#import "CordovaApp-Swift.h"
+#import "CDVTestHelpers.h"
+
+@interface CDVStatusBarTests : XCTestCase
+
+@property AppDelegate* appDelegate;
+@property (nonatomic, strong) CDVWebViewEngine* plugin;
+@property (nonatomic, strong) CDVViewController* viewController;
+
+@end
+
+@implementation CDVStatusBarTests
+
+- (void)setUp {
+    [super setUp];
+
+    self.appDelegate = (AppDelegate*)[[UIApplication sharedApplication] delegate];
+    [self.appDelegate createViewController];
+    self.viewController = self.appDelegate.testViewController;
+
+    self.plugin = (CDVWebViewEngine *)self.viewController.webViewEngine;
+}
+
+- (void)tearDown {
+    [self.appDelegate destroyViewController];
+    [super tearDown];
+}
+
+- (void) testStatusBarBackgroundColorNamedColor {
+    CGFloat rgba[4] = {0.f, 0.f, 0.f, 0.f};
+
+    [self.viewController loadStartPage];
+
+    XCTestExpectation *loadExpectation = [[XCTNSNotificationExpectation alloc] initWithName:CDVTestingDeviceReadyFired];
+    [self waitForExpectations:@[loadExpectation] timeout:5];
+
+    __block XCTestExpectation *redExpectation = [self expectationWithDescription:@"red"];
+    [self.plugin evaluateJavaScript:@"window.statusbar.setBackgroundColor('red');" completionHandler:^(id _, NSError* error) {
+        XCTAssertNil(error);
+        [redExpectation fulfill];
+    }];
+    [self waitForExpectations:@[redExpectation] timeout:5];
+
+    [self.viewController.statusBarBackgroundColor getRed:&rgba[0] green:&rgba[1] blue:&rgba[2] alpha:&rgba[3]];
+    XCTAssertEqual(rgba[0], 1.f);
+    XCTAssertEqual(rgba[1], 0.f);
+    XCTAssertEqual(rgba[2], 0.f);
+    XCTAssertEqual(rgba[3], 1.f);
+}
+
+- (void) testStatusBarBackgroundColor3HexColor {
+    CGFloat rgba[4] = {0.f, 0.f, 0.f, 0.f};
+
+    [self.viewController loadStartPage];
+
+    XCTestExpectation *loadExpectation = [[XCTNSNotificationExpectation alloc] initWithName:CDVTestingDeviceReadyFired];
+    [self waitForExpectations:@[loadExpectation] timeout:5];
+
+    __block XCTestExpectation *redExpectation = [self expectationWithDescription:@"red"];
+    [self.plugin evaluateJavaScript:@"window.statusbar.setBackgroundColor('#f00');" completionHandler:^(id _, NSError* error) {
+        XCTAssertNil(error);
+        [redExpectation fulfill];
+    }];
+    [self waitForExpectations:@[redExpectation] timeout:5];
+
+    [self.viewController.statusBarBackgroundColor getRed:&rgba[0] green:&rgba[1] blue:&rgba[2] alpha:&rgba[3]];
+    XCTAssertEqual(rgba[0], 1.f);
+    XCTAssertEqual(rgba[1], 0.f);
+    XCTAssertEqual(rgba[2], 0.f);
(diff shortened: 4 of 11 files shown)


CordovaLib/Classes/Private/Plugins/CDVStatusBarInternal/CDVStatusBarInternal.m
```diff
@@ -37,8 +37,9 @@ - (void)setBackgroundColor:(CDVInvokedUrlCommand *)command
     NSInteger valueR = [[command argumentAtIndex:0 withDefault:@0] integerValue];
     NSInteger valueG = [[command argumentAtIndex:1 withDefault:@0] integerValue];
     NSInteger valueB = [[command argumentAtIndex:2 withDefault:@0] integerValue];
+    NSInteger valueA = [[command argumentAtIndex:3 withDefault:@255] integerValue];
```
GitToTheHub: The alpha is at index 0 not 3

GitToTheHub (changes requested):

CordovaLib/Classes/Private/Plugins/CDVStatusBarInternal/CDVStatusBarInternal.m
dpogue: The alpha is at index 3 when returned from `getComputedStyle` in RGBA syntax, but it is not an integer, it's a float so this does need changing...

```js
document.body.style.background = '#ff000080';
getComputedStyle(document.body).backgroundColor
//=> 'rgba(255, 0, 0, 0.5)'
```

GitToTheHub: Wow this confuses things really up. The format for the config.xml preference `StatusBarBackgroundColor ` is `#AARRGGBB`. I used always that format. You can also see in the blog post for [cordova-android 15](https://cordova.apache.org/announcements/2026/03/06/cordova-android-15.0.0.html) that that format is used for the JS API:

```js
window.statusbar.setBackgroundColor('#AARRGGBB')
```

I tried setting the background color on Android by

```js
window.statusbar.setBackgroundColor('#aaff0000');
```

it it makes a light red status bar.

dpogue: Android and CSS do their hex notations differently: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/hex-color

This is iOS, and we're interpreting a colour as defined in CSS

GitToTheHub: But you are right, `window.getComputedStyle` creates a `rgba` value, where the alpha would be the last value. The JS side on Android reads the first value of `rgba` as alpha not as red:



So a hex like `#aaff0000` is interpreted as:
- A=170
- R=255
- G=0
- B=0

So I thing `window.getComputedStyle` was a little bit misused here to keep the format `#AARRGGBB`.

dpogue: The `#AARRGGBB` format is Android-specific, so that's not what we should be trying to do in a web context.

It looks like there are issues with the Android side of the colour parsing for the status bar, I'll try to get a PR up there to make it all consistent later this week.

GitToTheHub: I think the Android JS status bar API just orientated itself on the preference `StatusBarBackgroundColor` where the format is `#AARRGGBB`. Android correctly outputs `rgba` by `window.getComputedStyle`. Do you mean with Android-specific when you use color formats in the native Android development? `#AARRGGBB` is the documented color format for the plugin `cordova-plugin-statusbar` even for iOS.

dpogue: `StatusBarBackgroundColor` looks like it does use `#AARRGGBB` but it's handled somewhere else entirely: https://github.com/apache/cordova-ios/blob/7db38623f00e83fdb1e485e39c8b1580ee217dd0/lib/prepare.js#L637-L678

GitToTheHub: Yes, I wonder why it is on iOS handled differently and not on runtime like it is done on Android.

dpogue: Because iOS doesn't use Android's weird colour int stuff. Everything on iOS is UIColor.

We also need to convert it to a ColorSet resource so it can be applied at the storyboard level by Xcode at compile time with `ibtool`

GitToTheHub: But when setting the status bar backgorund color by the JS API `window.statusbar.setBackgroundColor` it's also happening on runtime and gets converted to `UIColor` by `[CDVStatusBarInternal setBackgroundColor:]`. Could the same not be done for the preference `StatusBarBackgroundColor`?

dpogue:
This should be good to go. I've added unit tests to confirm that it works correctly with a variety of valid CSS colour formats (including 8-digit hex) and that the opacity/alpha value is passed through correctly.

I've tried to make the unit tests reliable, but was having a lot of trouble where sometimes it seemed like cordova.js wasn't getting loaded and the `window.statusbar.setBackgroundColor` API would be undefined. I can't find a reason for that though, and they are passing most of the time on my machine now :\

GitToTheHub:
Alpha is lost, if the alpha is not `FF` like `#FF0000AA`. It gets resolved by `getComputedStyle` to `rgba(255, 0, 0, 0.667)` but wrongly parsed by the regular expression `/\d+/g` which resolves to an array with 5 elements: `[255, 0, 0, 0, 667]`. See https://regex101.com/r/p21APA/1

GitToTheHub:
> Alpha is lost, if the alpha is not `FF` like `#FF0000AA`. It gets resolved by `getComputedStyle` to `rgba(255, 0, 0, 0.667)` but wrongly parsed by the regular expression `/\d+/g` which resolves to an array with 5 elements: `[255, 0, 0, 0, 667]`. See https://regex101.com/r/p21APA/1

I tested old code. I forgot to do npm install on the checked out PR. Now everything is ok.

GitToTheHub (approved):
I don't have knowledge about the test stuff but otherwise it looks good to me. Tested with and without alpha and also uneven alpha and it worked.

GitToTheHub:
I added a note in the PR title and description, that the format must be `#RRGGBBAA`

GitToTheHub:
The related change for Android by using the css color format `#RRGGBBAA` is apache/cordova-android/pull/1955

dpogue:
> I added a note in the PR title and description, that the format must be `#RRGGBBAA`

No, the format is **any valid CSS [&lt;color&gt;](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value) value**.

Anything that's valid as a value of `meta theme-color` is valid here, it does not need to be in hexadecimal format

cordova-js-src/plugin/ios/statusbar.js
GitToTheHub, line 55: Maybe a documentation could be added here:

```jsdoc
// Sets the background color of the status bar, which will visually only have an impact,
// if the status bar is visible. The supported format is any valid CSS color value,
// like `rebeccapurple`, `#RRGGBBAA`, `rgb(255 0 153)`.
// See also: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value
// A CSS color value can only be used, if cordova-plugin-statusbar is not installed.
// If the external plugin is installed, the call will be forwarded to
// `window.StatusBar.backgroundColorByHexString`, which will only support the format `#RRGGBB`.
```

GitToTheHub (changes requested):

GitToTheHub:
> No, the format is any valid CSS [<color>](https://developer.mozilla.org/en->US/docs/Web/CSS/Reference/Values/color_value) value.
>
> Anything that's valid as a value of meta theme-color is valid here, it does not need to be in hexadecimal format

I reviewed the code and see now that alpha was never supported for iOS. So the user had never the chance to use the format `#AARRGGBB` for iOS. I just wanted to point out a format change, but this is not valid for iOS. I added a note in the description that any valid CSS Color is supported. People who were using the status bar plugin and only used the `#AARRGGBB` pattern, are maybe not aware that now any valid CSS color is supported and the hexadecimal format is different here.

GitToTheHub:
> No, the format is any valid CSS [<color>](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value) value.
>
> Anything that's valid as a value of meta theme-color is valid here, it does not need to be in hexadecimal format

This is only true, if the external plugin cordova-plugin-statusbar is not installed

cordova-js-src/plugin/ios/statusbar.js
dpogue: > ```
> // A CSS color value can only be used, if cordova-plugin-statusbar is not installed.
> // If the external plugin is installed, the call will be forwarded to
> // `window.StatusBar.backgroundColorByHexString`, which will only support the format `#RRGGBB`.
> ```

This is not true. If the status bar plugin is installed, it will use the specified CSS colour but with the alpha channel stripped out.

dpogue:
> This is only true, if the external plugin cordova-plugin-statusbar is not installed

No it is not. We take in any valid CSS colour, we ask the browser to convert it to RGB for us (by setting it as a CSS property and reading back the computed value with `getComputedStyle`), and then we pass the RGB values to either the built-in status bar or the status bar plugin.

cordova-js-src/plugin/ios/statusbar.js
GitToTheHub: Oh man sorry. But alpha is not supported, when the plugin is installed. Changed comment:

```jsdoc
// Sets the background color of the status bar, which will visually only have an impact,
// if the status bar is visible. The supported format is any valid CSS color value,
// like `rebeccapurple`, `#RRGGBBAA`, `rgb(255 0 153)`.
// See also: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value
// If cordova-plugin-statusbar is installed, the call will be forwarded to
// `window.StatusBar.backgroundColorByHexString`, which will not support an alpha value.
```

GitToTheHub:
> No it is not. We take in any valid CSS colour, we ask the browser to convert it to RGB for us (by setting it as a CSS property and reading back the computed value with getComputedStyle), and then we pass the RGB values to either the built-in status bar or the status bar plugin.

Yes I saw that [StatusBar.backgroundColorByHexString](https://github.com/apache/cordova-plugin-statusbar#statusbarbackgroundcolorbyhexstring) will be called when the plugin is installed and alpha is not given to it. When I read the documentation it says:

> On iOS, when you set StatusBar.overlaysWebView to false, you can set the background color of the statusbar by a hex string (#RRGGBB).

So would alpha be supported by this method when `StatusBar.overlaysWebView` is set to `true`?

dpogue:
> So would alpha be supported by this method when `StatusBar.overlaysWebView` is set to `true`?

No, it looks like the iOS side of the statusbar plugin only accepts 6 characters and always sets alpha to 1.0: https://github.com/apache/cordova-plugin-statusbar/blob/62d45cda82af84f5e75b02ef01e506d515c49579/src/ios/CDVStatusBar.m#L323-L346

GitToTheHub:
> No, it looks like the iOS side of the statusbar plugin only accepts 6 characters and always sets alpha to 1.0

Ok so the documentation is wrong on the plugin's readme. I will change this.

GitToTheHub:
Could the plugin documentation for iOS mean maybe, that the background color can only be set, when `StatusBar.overlaysWebView` is set to `false` and otherwise it would not be possible?:

> On iOS, when you set StatusBar.overlaysWebView to false, you can set the background color of the statusbar by a hex string (#RRGGBB).

dpogue:
IIRC if `overlaysWebView` is set to true, then there's no background shown, so the background colour doesn't matter.
You can still call `backgroundColorByHexString` and it will change the background color of the UIView, but the UIView is hidden.

In any case, that's plugin behaviour and not related to what's implemented here.

GitToTheHub:
> IIRC if overlaysWebView is set to true, then there's no background shown, so the background colour doesn't matter.
You can still call backgroundColorByHexString and it will change the background color of the UIView, but the UIView is hidden.

Ok thanks for clearifying this. I created a PR on the plugin to change the readme apache/cordova-plugin-statusbar/pull/287. I just asked you here because I was unsure if I changed the readme correctly. But made it now more generic.

erisu (approved):
LGTM

cordova-js-src/plugin/ios/statusbar.js
GitToTheHub: Want you add this comment to the function, or should this be done in a separate PR? Is it ok to add such a comment?

GitToTheHub (approved):
I will add documentation later

PR #1662, opened 2026-06-05
test(api): Add a test to ensure versions have availability macros

### LINKED ISSUE #1660: Cordova iOS 8.1.0 does not define __CORDOVA_8_1_0
# Bug Report

An availability macro should be defined for each Cordova iOS version so that plugins can check was version of Cordova they are being compiled against using the `CORDOVA_MIN_VERSION_REQUIRED` macro.

There is no macro defined for version 8.1.0 (`__CORDOVA_8_1_0`) despite new API being introduced that plugins might want to conditionally adopt.

### Command or Code

The following plugin code will fail to compile against Cordova iOS 8.1.0 because it cannot detect that the symbol is already defined, despite taking all the appropriate measures to only define it on older versions.

```objc
// Define this notification name if we're on an older Cordova version
#if !defined(__CORDOVA_8_1_0) || CORDOVA_VERSION_MIN_REQUIRED < __CORDOVA_8_1_0
const NSNotificationName CDVPluginContinueUserActivityNotification = @"CDVPluginContinueUserActivityNotification"
#endif
```

dpogue:
### Platforms affected
iOS

### Motivation and Context

The Cordova iOS 8.1.0 minor release is missing a `__CORDOVA_8_1_0` macro in CDVAvailability.h.

Closes #1660

### Description

Add a unit test to confirm the a macro is defined that matches the current version from package.json. This means the macro will need to be added when a dev version bump happens, and a missing version at tagging time will result in test failures.

### Testing

1. Ran tests with missing macro and ensured they failed.
2. Added the macros.
3. Ran tests and ensured that they succeeded.

### Checklist


erisu (approved):
LGTM.

This should help test and ensure that macros are added whenever the version is bumped and are ready for future releases. While this may create macros for versions that are never released, it's better to have extra macros than to miss one that is required.

GitToTheHub:
Nice :) We will see how this will feel with future development versions. I think this is a very good thing to check if the macro version is set.

NiklasMerz:
That's awesome. It's so easy to forget stuff like this and thanks for figuring out the automation to test this.

breautek (approved):

PR #1663, opened 2026-06-09
chore(versions): Vary minimum version based on compile SDK

### LINKED ISSUE #1723: Minimum deployment target in Xcode 27 is iOS 15+
# Bug Report

## Problem

Xcode 27 requires minimum deployment target iOS 15+. The iOS cordova project has minimum target set to 13 now. 

Success when building on XCode 27.

It fails because the target is set to 13. 

Adding `deployment-target` to the `config.xml` file helps only partially.
<preference name="deployment-target" value="15.0"/>

It fails on Pods.xcodeproj:

platforms/ios/Pods/Pods.xcodeproj: error: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 13.0, but the range of supported deployment target versions is 15.0 to 27.0.x. 

### Version information
macOS 27, XCode 27, Cordova iOS 8.1.1, Cordova CLI 13

## Checklist
- [ x ] I searched for existing GitHub issues
- [ x ] I updated all Cordova tooling to most recent version
- [ x ] I included all the necessary information above

dpogue:
### Platforms affected
iOS

### Motivation and Context

Xcode is getting more strict about minimum supported OS requirements, what used to be a warning about version numbers being out of range is being turned into an error. In Xcode 27 beta 1, our minimum of 13.0 is an error because the lowest supported target version is 15.0.

### Description

To avoid needing to raise our actual minimum OS version unnecessarily, we can use an Xcode macro that automatically results in the minimum supported OS version in the current Xcode version. These macros were introduced in Xcode 14, which is older than our minimum Xcode requirement.

The end result of this is that the realistic minimum supported OS version for a Cordova app will be determined by the Xcode version used to build it.

It doesn't seem to care about the minimum version declared in Package.swift and whether that's in the range of supported versions or not.

I would feel better if there were a way to use this conditionally and still set a hardcoded minimum, or a way to see clearly what the Xcode-supported minimum is, but I can't see a way to do that without delving into complex xcconfig chaining :(

Fixes #1723

### Testing

Opened the project and ran tests in Xcode 27 beta 1 with no issues.

### Checklist


GitToTheHub:
I would suggest to raise the minimum iOS vesion to 15. iOS 15 is still supported by Apple and get security updates: https://support.apple.com/de-de/100100. The latest iOS 15 version is iOS 15.8.8 released on 11th May 2026. Everything lower will not get security updates and is a security risk. The oldest device supported by iOS 15 is the iPhone 6s and all devices which support iOS 13 and 14 can be updated to 15. iPhone 6 for e.g. supports maximum iOS 12.5.8. I don't see why we should support unsupported versions by Apple which can be updated.

jcesarmobile:
We should bump the minimum iOS version to 15 in next major, but would still need this (or something else, haven't really evaluated the change), as a patch release, unless we plan to release the major very soon.

breautek (commented):
Not sure if the removal of `SUPPORTED_PLATFORMS` or `TARGETED_DEVICE_FAMILY` is intentional. Doesn't seem to be related to the deployment target change documented.

dpogue:
> Not sure if the removal of `SUPPORTED_PLATFORMS` or `TARGETED_DEVICE_FAMILY` is intentional. Doesn't seem to be related to the deployment target change documented.

That was intentional because they were getting set to one value at the project level and then (inconsistently) getting set to different values at the target level. We ideally want them to be consistent for the whole project so I updated the project values and removed the per-target overrides.

dpogue:
Bumping our default to iOS 15 would involve a major version bump, which I was hoping to hold off for a while. Our next major should also remove all our deprecated API and include some other fairly significant changes, and that's going to cause a bunch of older plugins to break and for people to avoid upgrading out of fear. There are a lot of people who hang around on old Cordova versions because upgrading causes everything to break, and we need to make the process as painless as possible for them.

I'm going to look at a slightly different approach here that would accomplish the same result (bumping the minimum to iOS 15 but only in Xcode 27) but with more guarantees around the actual minimum version number being used.

dpogue:
I pushed another commit here to explicitly vary based on SDK rather than the magic "recommended" minimum value.

It's clearer what's going on, but I still don't like it. Thanks Apple 😒

breautek (approved):
I think what we have here is good for a patch -- the minimum of 15 is coming from SDK/xcode 27 -- and not from Cordova as a framework.

GitToTheHub:
Ok I just noted it to have less work. But if you find a good solution than it's good also.

GitToTheHub:
Should we bump already to 8.2.0?

dpogue:
> Should we bump already to 8.2.0?

When we merge this we should definitely bump the minor

GitToTheHub:
Ok, i was just asking because 3 PRs with the 8.2.0 milestone were already merged https://github.com/apache/cordova-ios/milestone/17?closed=1

dpogue:
None of them introduce new features, so none of them technically require a minor version bump.

But I'm tagging the milestone with the assumption that we'll move right to a 8.2.0 release without doing an 8.1.1.

GitToTheHub:
Ok understand :)

erisu (approved):

GitToTheHub:
I think this could be merged now?

dpogue:
I feel like we should hold off on merging this until Xcode 27 is actually released and we can confirm it works with the final version.

GitToTheHub:
Yeah, I think that's a good idea. XCode 27 will be released soon.

dpogue:
I've tested this with the Xcode 27 RC and I think we're probably good to go here.

GitToTheHub:
I wonder why the config.xml preference `deployment-target` will not change the deployment target here. If it would do so, it would be more easy to change it according the Xcode version? When you take care about CocoaPods, must there be something taken into account about Swift Package Manager?

templates/project/App.xcodeproj/project.pbxproj
```diff
@@ -414,16 +417,19 @@
...
 				MTL_ENABLE_DEBUG_INFO = NO;
 				MTL_FAST_MATH = YES;
 				SDKROOT = iphoneos;
 				STRING_CATALOG_GENERATE_SYMBOLS = YES;
+				SUPPORTED_PLATFORMS = "iphonesimulator iphoneos xros xrsimulator";
 				SUPPORTS_MACCATALYST = YES;
 				SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = NO;
 				SWIFT_COMPILATION_MODE = wholemodule;
-				TARGETED_DEVICE_FAMILY = "1,2";
+				TARGETED_DEVICE_FAMILY = "1,2,6,7";
```
GitToTheHub: AI noted here:

> The generated Cordova app now inherits Mac Catalyst and visionOS support. In templates/project/App.xcodeproj/project.pbxproj, project-level settings change TARGETED_DEVICE_FAMILY from 1,2 (iPhone/iPod touch, iPad) to 1,2,6,7 (iPhone/iPod touch, iPad, Mac Catalyst, visionOS) and add `xros` (visionOS) platforms. This affects the consumer App target, not just the Cordova package target.

Is this change intended?

dpogue: As I mentioned in https://github.com/apache/cordova-ios/pull/1663#issuecomment-4673452172 that's intentional to de-duplicate some inconsistent per-target overrides

dpogue:
> I wonder why the config.xml preference `deployment-target` will not change the deployment target here. If it would do so, it would be more easy to change it according the Xcode version?

Whatever is specified in config.xml as `deployment-target` should override all of this. We can't know in the JS code what Xcode version the resulting project will be opened with though, so we have to do that in a way that Xcode itself can determine.



> When you take care about CocoaPods, must there be something taken into account about Swift Package Manager?

As far as I can tell, Swift package manager modules inherit the deployment target of the main project, so there's no issue there. (There's only an issue if the project deployment target is lower than the minimum supported version for a module.)

templates/project/App.xcodeproj/project.pbxproj
GitToTheHub: Ok sounds good :)

GitToTheHub:
So from my side it's fine to merge

GitToTheHub (approved):

dpogue:
I've tested this with a Cocoapod-using plugin now and the Pods project is correctly picking up the iOS 15 deployment target. However, the individual pod targets use whatever version is set in their podspec, which might be older than iOS 15 and that's not something that we can fix.

As a reminder, Cocoapods goes read-only as of December 2026.

PR #1664, opened 2026-06-10
fix(wkwebview): fix ViewController being the ScriptMessageHandler

dpogue:
### Platforms affected
iOS

### Motivation and Context

We've had a mechanism for a while where if the ViewController is a WKScriptMessageHandler then it will be registered as the handler for Cordova actions rather than the default plugin implementation.

Except there was a bug where it would try to register both of them with the same name and fail.

### Description

This checks if the ViewController is a candidate first, and only registers the plugin as the script handler if it is not.

Also add a unit test case to ensure that a ViewController that implements the WKScriptMessageHandler protocol correctly receives the script messages.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -40,9 +40,6 @@
 @interface CDVWebViewEngine ()
 
 @property (nonatomic, strong, readwrite) UIView* engineWebView;
-@property (nonatomic, strong, readwrite) id <WKUIDelegate> uiDelegate;
-@property (nonatomic, weak) id <WKScriptMessageHandler> weakScriptMessageHandler;
-@property (nonatomic, strong) CDVURLSchemeHandler * schemeHandler;
 @property (nonatomic, readwrite) NSString *CDV_ASSETS_URL;
 @property (nonatomic, readwrite) Boolean cdvIsFileScheme;
 @property (nullable, nonatomic, strong, readwrite) WKWebViewConfiguration *configuration;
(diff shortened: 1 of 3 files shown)


erisu (approved):
LGTM

CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
```diff
@@ -260,8 +248,21 @@ - (void)pluginInitialize
...
     }
 
+    CDVWebViewWeakScriptMessageHandler *weakScriptMessageHandler;
     if ([self.viewController conformsToProtocol:@protocol(WKScriptMessageHandler)]) {
-        [wkWebView.configuration.userContentController addScriptMessageHandler:(id < WKScriptMessageHandler >)self.viewController name:CDV_BRIDGE_NAME];
+        weakScriptMessageHandler = [[CDVWebViewWeakScriptMessageHandler alloc] initWithScriptMessageHandler:(id < WKScriptMessageHandler >)self.viewController];
+    } else {
+        weakScriptMessageHandler = [[CDVWebViewWeakScriptMessageHandler alloc] initWithScriptMessageHandler:self];
+    }
+    [userContentController addScriptMessageHandler:weakScriptMessageHandler name:CDV_BRIDGE_NAME];
```
GitToTheHub: Can you document what's happening here?

CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
```diff
@@ -260,8 +248,21 @@ - (void)pluginInitialize
...
+
+    if(self.CDV_ASSETS_URL) {
+        NSString *scriptCode = [NSString stringWithFormat:@"window.CDV_ASSETS_URL = '%@';", self.CDV_ASSETS_URL];
+        WKUserScript *wkScript = [[WKUserScript alloc] initWithSource:scriptCode injectionTime:WKUserScriptInjectionTimeAtDocumentStart forMainFrameOnly:YES];
+
+        if (wkScript) {
+            [userContentController addUserScript:wkScript];
+        }
```
GitToTheHub: Can you document this part?

GitToTheHub (changes requested):
I somebody is not familiar with the Cordova internals, it's hart to understand everything just from the code. Maybe some comments would help to understand it better.

dpogue:
I've updated this with comments and cleaned up some unnecessary instance properties of the CDVWebViewEngine

GitToTheHub (approved):
Looks good

PR #1667, opened 2026-06-13
fix: proper macro check for `TARGET_OS_VISION`

GitToTheHub:
### Platforms affected

- iOS

### Motivation and Context

- `TARGET_OS_VISION` is defined by Apple’s headers as a macro with a value, so `#ifndef TARGET_OS_VISION` is the wrong test. On iPhone it is typically defined as 0, which still makes `#ifndef` false. That means the guarded code gets excluded even though you are not on visionOS.
- Test the macro by `#if !TARGET_OS_VISION` instead of `#ifndef TARGET_OS_VISION`
- Generated-By: GPT-5.3-Codex

### Description

- During testing the status bar, I could no longer set the status bar background color, though `viewport-fit=contain` was set in `<meta name="viewport" content="...">`. After doing some analyzes with Copilot, I saw the added changes for checking `TARGET_OS_VISION` and commented them out. After that, the status bar was visible again with the set background color. GPT helped me to make a proper check of the macro.

### Testing

- Setting the status bar background color by preference `StatusBarBackgroundColor` and `<meta name="theme-color" content="#ffff00">` on a iOS 26.5 simulator.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -403,7 +403,7 @@ static UIColor *defaultBackgroundColor(void) {
 {
     [super viewDidAppear:animated];
 
-#if TARGET_OS_MACCATALYST
+#if defined(TARGET_OS_MACCATALYST) && TARGET_OS_MACCATALYST
     BOOL hideTitlebar = [self.settings cordovaBoolSettingForKey:@"HideDesktopTitlebar" defaultValue:NO];
     if (hideTitlebar) {
         UIWindowScene *scene = self.view.window.windowScene;
@@ -577,7 +577,7 @@ static UIColor *defaultBackgroundColor(void) {
 
 - (void)scrollViewDidChangeAdjustedContentInset:(UIScrollView *)scrollView
 {
-#if !TARGET_OS_VISION
+#if !defined(TARGET_OS_VISION) || !TARGET_OS_VISION
     if (self.webView.hidden) {
         self.statusBar.hidden = true;
         return;
@@ -587,7 +587,7 @@ static UIColor *defaultBackgroundColor(void) {
 #endif
 }
 
-#if !TARGET_OS_VISION
+#if !defined(TARGET_OS_VISION) || !TARGET_OS_VISION
 - (BOOL)prefersStatusBarHidden
 {
     // The CDVStatusBar plugin overrides this in a category extension, and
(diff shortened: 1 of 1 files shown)


CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -577,7 +577,7 @@ - (void)onWebViewPageDidLoad:(NSNotification*)notification
 
 - (void)scrollViewDidChangeAdjustedContentInset:(UIScrollView *)scrollView
 {
-#ifndef TARGET_OS_VISION
+#if !TARGET_OS_VISION
```
dpogue: This is _probably_ fine as-is, but the safer thing to do would be check `#if !defined(TARGET_OS_VISION) || !TARGET_OS_VISION` so that it doesn't caused undefined macro warnings if using an SDK that predates visionOS

dpogue (commented):
🤦🏼‍♂️ Good catch! When I added this, Xcode looked like it was doing the right thing by greying out the code for visionOS, but I didn't confirm that it re-enabled the code when targeting iOS again.

CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -577,7 +577,7 @@ - (void)onWebViewPageDidLoad:(NSNotification*)notification
 
 - (void)scrollViewDidChangeAdjustedContentInset:(UIScrollView *)scrollView
 {
-#ifndef TARGET_OS_VISION
+#if !defined(TARGET_OS_VISION) || !TARGET_OS_VISION
```
GitToTheHub, line 580: I was thinking of adding a global macro `CDV_TARGET_OS_VISION` which is only set when `TARGET_OS_VISION` is defined or is set to 1. What do you think?

GitToTheHub: Also a comment could be added that this is only available since XCode 26 and will be set to 1 if Vision OS is targeted.

dpogue: I don't know if there's much benefit to defining our own macro in addition to the ones Apple already defines, which are well established convention across their platforms

GitToTheHub: Maybe this pattern could be used somewhere else. In this case the condition is more easy:

```objc
#ifdef CDV_TARGET_OS_VISION
```

or

```objc
#ifndef CDV_TARGET_OS_VISION
```

erisu (approved):

PR #1670, opened 2026-06-23
doc(statusbar.js): document `visible` property and `setBackgroundColor` method

GitToTheHub:
### Motivation and Context

- Document `visible` property
  - Make clear that this hides the system status bar completely when `false` is set.
- Document `setBackgroundColor`
  - Document that the supported format for `setBackgroundColor` is any valid CSS color and adding some examples.
  - Add minimum cordova-ios version for full CSS color formats support which is 8.1.1
  - Mention that `setBackgroundColor` is intended for special cases only, and not a recommended API. The recommended way to set the background colour of the status bar is with the `<meta name="theme-color">` tag.
- The comments don't make references to the status bar plugin, because that is likely to add to developer confusion about whether the status bar plugin is required or not. If people are using the status bar plugin, they should be calling its APIs. We added the fallback for compatibility reasons, not to encourage mixing the two.

### Checklist

### DIFF
--- a/cordova-js-src/plugin/ios/statusbar.js
+++ b/cordova-js-src/plugin/ios/statusbar.js
@@ -52,6 +52,13 @@ Object.defineProperty(statusBar, 'visible', {
     }
 });
 
+/**
+ * Sets the background color of the status bar, which will visually only have an impact,
+ * if the status bar is visible. The supported format is any valid CSS color value,
+ * like `rebeccapurple`, `#RRGGBBAA`, `rgb(255 0 153)`.
+ * If cordova-plugin-statusbar is installed, the call will be forwarded to
+ * `window.StatusBar.backgroundColorByHexString`, which will not support an alpha value.
+ */
 Object.defineProperty(statusBar, 'setBackgroundColor', {
     configurable: false,
     enumerable: false,
--- a/cordova-js-src/plugin/ios/statusbar.js
+++ b/cordova-js-src/plugin/ios/statusbar.js
@@ -26,6 +26,10 @@ var statusBar = {};
 const statusBarScript = document.createElement('script');
 document.head.appendChild(statusBarScript);
 
+/**
+ * Sets the visibility of the status bar. If cordova-plugin-statusbar is used
+ * the call will be forwarded to `window.StatusBar.show` or `window.StatusBar.hide` of the plugin.
+ */
 Object.defineProperty(statusBar, 'visible', {
     configurable: false,
     enumerable: true,
--- a/cordova-js-src/plugin/ios/statusbar.js
+++ b/cordova-js-src/plugin/ios/statusbar.js
@@ -58,9 +58,9 @@ Object.defineProperty(statusBar, 'visible', {
 /**
  * Sets the background color of the visible status bar.
  * Supports valid CSS color values, e.g. `rebeccapurple`, `#RRGGBBAA`, `rgb(255 0 153)`.
- * 
+ *
  * Note: Runtime support for all valid CSS color formats is fully functional since cordova-ios 8.1.1.
- * 
+ *
  * Note: This is intended for special cases only, and not a recommended API. The recommended way
  * to set the background colour of the status bar is with the `<meta name="theme-color">` tag.
  */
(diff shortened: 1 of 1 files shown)


cordova-js-src/plugin/ios/statusbar.js
```diff
@@ -52,6 +52,13 @@ Object.defineProperty(statusBar, 'visible', {
     }
 });
 
+/**
+ * Sets the background color of the status bar, which will visually only have an impact,
+ * if the status bar is visible. The supported format is any valid CSS color value,
+ * like `rebeccapurple`, `#RRGGBBAA`, `rgb(255 0 153)`.
+ * If cordova-plugin-statusbar is installed, the call will be forwarded to
+ * `window.StatusBar.backgroundColorByHexString`, which will not support an alpha value.
+ */
```
erisu, line 64:
```suggestion
/**
 * Sets the background color of the visible status bar.
 * Supports valid CSS color values, e.g. `rebeccapurple`, `#RRGGBBAA`, `rgb(255 0 153)`.
 * 
 * If cordova-plugin-statusbar is installed, calls are forwarded to the plugin API:
 * `window.StatusBar.backgroundColorByHexString`
 * See {@link https://s.apache.org/cdv-plugin-statusbar} for cordova-plugin-statusbar details.
 */
```

What you think of this?

It simplifies the wording and removing the alpha limitation of the plugin.

IMO, the alpha limitation comment is not necessary here. While it may be unlikely, if the implementation detail of the plugin changes, it could make the comment obsolete. Instead, I thought including the plugin doc link would be nice way to avoid this.

GitToTheHub: > MO, the alpha limitation comment is not necessary here. While it may be unlikely, if the implementation detail of the plugin changes, it could make the comment obsolete. Instead, I thought including the plugin doc link would be nice way to avoid this.

I think also that implementation detail of the plugin will not change. I wanted to make things easy to understand, without needing to read the implementation details on the doc. I don't know what is best. Maybe @dpogue has also an opinion. I'm not sad if the link is added instead of the alpha detail. Otherwise you simplifications look good.

dpogue (commented):
My preference would be that these comments didn't make reference to the status bar plugin, because that is likely to add to developer confusion about whether the status bar plugin is required or not. If people are using the status bar plugin, they should be calling its APIs. We added the fallback for compatibility reasons, not to encourage mixing the two.

We should also mention that `setBackgroundColor` is intended for special cases only, and not a recommended API. The recommended way to set the background colour of the status bar is with the `meta theme-color` tag.

GitToTheHub:
> My preference would be that these comments didn't make reference to the status bar plugin, because that is likely to add to developer confusion about whether the status bar plugin is required or not. If people are using the status bar plugin, they should be calling its APIs. We added the fallback for compatibility reasons, not to encourage mixing the two.
> 
> We should also mention that setBackgroundColor is intended for special cases only, and not a recommended API. The recommended way to set the background colour of the status bar is with the meta theme-color tag.

I changed it like you suggested.

cordova-js-src/plugin/ios/statusbar.js
```diff
@@ -52,6 +55,15 @@ Object.defineProperty(statusBar, 'visible', {
     }
 });
 
+/**
+ * Sets the background color of the visible status bar.
+ * Supports valid CSS color values, e.g. `rebeccapurple`, `#RRGGBBAA`, `rgb(255 0 153)`.
+ * 
+ * Note: Runtime support for all valid CSS color formats is fully functional since cordova-ios 8.1.1.
```
dpogue: Is this note needed? The intention was always to support all valid CSS colour formats, the fact it didn't handle opacity properly was a bug that was fixed, not a new feature that needs to be called out.

dpogue (commented):
Hitting some eslint failures around trailing whitespace

cordova-js-src/plugin/ios/statusbar.js
GitToTheHub: This was when I synced the cordova-android comment with cordova-ios. But it's true, it was just about alpha. I corrected that.

GitToTheHub:
> Hitting some eslint failures around trailing whitespace

I fixed this.

dpogue (approved):
Still not sure that we need the bit about 8.1.1 (since this comment only exists in versions after 8.1.1), but it's probably fine. We can remove it later if it causes confusion.

GitToTheHub:
> Still not sure that we need the bit about 8.1.1 (since this comment only exists in versions after 8.1.1), but it's probably fine. We can remove it later if it causes confusion.

This would for example the case if someone looking in the master code on GitHub but still uses an older cordova-ios version

PR #1673, opened 2026-06-30
chore(ci): update & clean workflow

erisu:
### Motivation and Context

Update CI Workflow

### Description

* Use latest actions
* Move version number comment inline with usage
* Add node-26 to test matrix
* Set checkout persist-credentials to false
* Remove `.ratignore` since ci can use `.rat-excludes`

### Testing

`npm cit`

### Checklist


dpogue (commented):
We're currently running macos-15 for 3 NodeJS versions, and had added macos-26 with only NodeJS 24 as a test for Xcode 26 (since macos-26 runners were in beta at the time).

We should, at the very least, ensure that we're testing NodeJS 26 with macOS, and decide if we want everything to run on macOS 26 or if there's value in splitting and testing multiple Xcode versions.

Do we also want to pull the CodeQL stuff out to a step under linting so it only runs once instead of a dozen times?

dpogue (approved):

PR #1680, opened 2026-07-10
fix(CDVStatusBarTests): wait for native exec

GitToTheHub:
### Motivation and Context

- Sometimes actions will fail due to race conditions on the `testStatusBarBackgroundColor*`-methods. `evaluateJavaScript` completes once JavaScript has executed, but `cordova.exec(...)` queues the native `StatusBarInternal.setBackgroundColor` command asynchronously. The test then immediately reads `statusBarBackgroundColor`, potentially before the native command runs.
- Add a reusable test helper that evaluates JavaScript and waits until Cordova exec calls queued by that JavaScript have reached native code.
- StatusBar tests now use this helper before asserting native status bar state, avoiding races where evaluateJavaScript completes before the native StatusBarInternal command has run. Also register the deviceready expectation before loading the start page.
- Generated-By: ChatGPT 5.5

### Checklist

### DIFF
--- a/tests/CordovaLibTests/CDVStatusBarTests.m
+++ b/tests/CordovaLibTests/CDVStatusBarTests.m
@@ -34,20 +34,6 @@
 
 @implementation CDVStatusBarTests
 
-/**
- * Wait until statusBarBackgroundColor is set natively, because calling the JS part would return immediately
- * and not wait for the native part to finish.
- */
-- (void)waitForStatusBarBackgroundColor {
-    NSPredicate *predicate = [NSPredicate predicateWithBlock:^BOOL(id _, NSDictionary *__) {
-        return self.viewController.statusBarBackgroundColor != nil;
-    }];
-    XCTestExpectation *colorExpectation = [[XCTNSPredicateExpectation alloc]
-        initWithPredicate:predicate
-        object:self.viewController];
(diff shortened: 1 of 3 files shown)


dpogue (commented):
oooh, thanks for tracking down the root cause of these random failures!

This is probably fine in the short term for this specific test case, but I wonder if it might be good to have a generic helper that waits until the command queue is empty (i.e., all pending plugin exec calls have been processed)

GitToTheHub:
> This is probably fine in the short term for this specific test case, but I wonder if it might be good to have a generic helper that waits until the command queue is empty (i.e., all pending plugin exec calls have been processed)

Would make sense. But I don't know how this could be done :)

GitToTheHub:
I created a generic solution with the help of ChatGPT:

> I replaced the StatusBar-specific wait with a generic test helper that waits for Cordova exec calls to reach native code.
> 
> The helper registers a test-only no-op plugin, appends a barrier exec call in the same JavaScript evaluation, and waits until that barrier is received natively. This avoids relying on evaluateJavaScript completion, which only confirms the JS finished running, not that the native plugin command has been processed.
> 
> I also moved the deviceready expectation before loadStartPage to avoid missing the notification if it fires quickly.

GitToTheHub:
If no one has any objections, I would merge this PR

PR #1681, opened 2026-07-24
fix(CDVSceneDelegate): deliver cold-launch deep link URLs after page load

### LINKED ISSUE #1671: iOS: custom-scheme deep link dropped on cold launch under UIScene (notification posted before CDVHandleOpenURL registers its observer)
# Bug Report

On a **cold (killed-state)** launch from a custom-scheme URL (e.g. `myscheme://path?foo=bar`), the global `window.handleOpenURL` should be called with the launch URL — the same way it is called when the app is already running.

On a cold launch under the UIScene lifecycle, `window.handleOpenURL` is **never called**; the URL is dropped in native code before it reaches JavaScript. The same link works correctly when the app is already running (warm start). Android is unaffected.

## Information

On a cold launch the open-URL notification is posted **before** `CDVHandleOpenURL` registers its observer, so `NSNotificationCenter` has no observer to deliver it to and the URL is lost.

Sequence:

1. `CDVSceneDelegate scene:willConnectToSession:options:` forwards `connectionOptions.URLContexts` to `scene:openURLContexts:`, which posts `CDVPluginHandleOpenURLNotification` synchronously during scene connection.
2. `CDVHandleOpenURL` adds its observer only in `pluginInitialize`, which runs later — after the web view and plugins bootstrap.
3. `NSNotificationCenter` delivers a notification only to observers present at post time. The observer is not registered yet, so `applicationLaunchedWithUrl:` never runs, `self.url` stays `nil`, and `applicationPageDidLoad:` later finds nothing to process.

Instrumented NSLog timeline on the iOS 18.3 simulator (stock generated `SceneDelegate`, `class SceneDelegate: CDVSceneDelegate {}`):

```
T+0.000  scene willConnectTo   urls=[myscheme://path?...]   // posts CDVPluginHandleOpenURLNotification
T+0.000  scene openURLContexts urls=[myscheme://path?...]   // forwarded by CDVSceneDelegate
T+0.139  CDVHandleOpenURL.pluginInitialize                  // observer registered — ~139 ms too late
T+1.171  CDVHandleOpenURL.applicationPageDidLoad hasUrl=0
```

This worked under the pre-scene `AppDelegate` model, because `application:openURL:options:` was called after `didFinishLaunchingWithOptions:` (the plugin observer already existed). Moving URL delivery into `scene:willConnectToSession:` placed it before plugin initialization.

**Suggested fix (framework):** buffer the launch URL so it survives until the observer exists. Either retain the cold-launch URL in `CDVSceneDelegate` and re-post `CDVPluginHandleOpenURLNotification` on `CDVPageDidLoadNotification`, or have `CDVHandleOpenURL` / CordovaLib hold the most recent launch URL and replay it when the plugin initializes.

**App-side workaround that resolves it today** — a custom `SceneDelegate` that buffers and re-posts on `CDVPageDidLoad`:

```swift
import Cordova

class SceneDelegate: CDVSceneDelegate {
    private var coldLaunchUrl: URL?

    override func scene(_ scene: UIScene, willConnectTo session: UISceneSession, options connectionOptions: UIScene.ConnectionOptions) {
        super.scene(scene, willConnectTo: session, options: connectionOptions)
        guard let url = connectionOptions.urlContexts.first?.url else { return }
        coldLaunchUrl = url
        NotificationCenter.default.addObserver(self, selector: #selector(redeliver),
            name: NSNotification.Name("CDVPageDidLoadNotification"), object: nil)
    }

    @objc private func redeliver() {
        NotificationCenter.default.removeObserver(self, name: NSNotification.Name("CDVPageDidLoadNotification"), object: nil)
        guard let url = coldLaunchUrl else { return }
        coldLaunchUrl = nil
        DispatchQueue.main.async {
            NotificationCenter.default.post(name: NSNotification.Name("CDVPluginHandleOpenURLNotification"), object: url)
        }
    }
}
```

A companion early JS buffer (define `window.handleOpenURL` before the app framework boots and drain it when ready) is also needed, because the re-delivered URL can still arrive before the page's `deviceready` handler assigns the real handler.

### Command or Code

```sh
xcrun simctl uninstall <udid> com.example.app   # ensure cold / not running
xcrun simctl install   <udid> "App.app"
xcrun simctl openurl   <udid> "myscheme://path?foo=bar"
# window.handleOpenURL is NOT called. With the app already running, it IS called.
```

The app must register the custom scheme in `Info.plist` (`CFBundleURLSchemes`) and define a global `window.handleOpenURL`.

- iOS 18.3 simulator (iPhone). Reproduces on scene-based apps (iOS 13+).
- WKWebView engine (`CDVWKWebViewEngine`).
- Stock generated `SceneDelegate` (no app changes).

### Version information

- Cordova platform: `cordova-ios` 8.1.0 (latest release `rel/8.1.0`)
- Engine: `cordova-plugin-wkwebview` / `CDVWKWebViewEngine`
- Build tools: Xcode 26.x

## Checklist

- [x] I updated all Cordova tooling to most recent version (`cordova-ios` 8.1.0 is the latest release)

---

<sub>Disclosure: this report was prepared with AI assistance (Claude / Claude Code). The root-cause analysis, NSLog instrumentation, timeline, and workaround were generated with that tooling and verified by the reporter on the iOS 18.3 simulator. No third-party copyrighted content is included; the code samples are original.
Generated-by: Claude Opus 4.8 (Claude Code)</sub>

rohitsux:
Closes #1671

## Summary

On a cold (killed-state) launch from a custom-scheme deep link under the UIScene lifecycle, `window.handleOpenURL` was never called — the launch URL was dropped in native code before reaching JavaScript. Warm launches were unaffected.

Root cause: `CDVSceneDelegate scene:willConnectToSession:` posts `CDVPluginHandleOpenURLNotification` synchronously during scene connection, but `CDVHandleOpenURL` only registers its observer later, in `-pluginInitialize`. `NSNotificationCenter` delivers a notification only to observers present at post time, so the notification — and the URL — was lost.

## Changes

- `CDVSceneDelegate.m` — on cold launch, buffer the launch URL contexts and register a one-shot observer for `CDVPageDidLoadNotification`; replay the open-URL notification once the page has loaded and the plugin observer exists. A `sceneDidDisconnect:` teardown removes the pending observer if the scene disconnects before the page loads. The runtime (warm) `openURLContexts:` path is unchanged.
- `CDVSceneDelegateTests.m` — new unit tests covering: cold-launch URL is buffered until page load, replayed after page load, no notification when there is no launch URL, disconnect-before-load cancels the pending URL, and the warm path still posts immediately.
- `CordovaTests.xcodeproj/project.pbxproj` — register the new test file in the CordovaTests target.

## Testing

All tests pass on the iOS 26.5 simulator (iPhone 17). The five new tests exercise both the cold-launch buffering/replay contract and the unchanged warm path; the existing suite continues to pass.

---

Generated-by: Opus 4.8

### DIFF
--- a/CordovaLib/Classes/Public/CDVSceneDelegate.m
+++ b/CordovaLib/Classes/Public/CDVSceneDelegate.m
@@ -20,12 +20,60 @@
 #import <Cordova/CDVSceneDelegate.h>
 #import <Cordova/CDVPluginNotifications.h>
 
+@interface CDVSceneDelegate ()
+
+// URL contexts captured on a cold launch, held until the page loads so that
+// CDVHandleOpenURL has registered its observer before we post the notification.
+@property (nonatomic, strong, nullable) NSSet<UIOpenURLContext *> *launchURLContexts;
+@property (nonatomic, weak, nullable) UIScene *launchScene;
+
+@end
+
 @implementation CDVSceneDelegate
 
 - (void)scene:(UIScene *)scene willConnectToSession:(UISceneSession *)session options:(UISceneConnectionOptions *)connectionOptions
 {
-    // If the app was launched from a URL, that should also fire the CDVPluginOpenURLNotification
-    [self scene:scene openURLContexts:connectionOptions.URLContexts];
+    // If the app was launched from a URL, that should also fire the CDVPluginHandleOpenURLNotification.
+    //
+    // On a cold launch this runs during scene connection, which is *before*
+    // CDVHandleOpenURL registers its observer in -pluginInitialize. Posting the
+    // notification now would deliver it to no observer and the URL would be lost
+    // (see https://github.com/apache/cordova-ios/issues/1671). Buffer the URL
+    // contexts and replay them once the page has loaded, by which time the
+    // plugin observer exists.
+    if (connectionOptions.URLContexts.count > 0) {
+        self.launchScene = scene;
+        self.launchURLContexts = connectionOptions.URLContexts;
+        [[NSNotificationCenter defaultCenter] addObserver:self
+                                                 selector:@selector(pageDidLoad:)
+                                                     name:CDVPageDidLoadNotification
+                                                   object:nil];
+    }
+}
+
+- (void)pageDidLoad:(NSNotification *)notification
+{
+    [[NSNotificationCenter defaultCenter] removeObserver:self name:CDVPageDidLoadNotification object:nil];
+
+    UIScene *scene = self.launchScene;
+    NSSet<UIOpenURLContext *> *contexts = self.launchURLContexts;
+    self.launchScene = nil;
+    self.launchURLContexts = nil;
+
+    if (scene != nil && contexts != nil) {
+        [self scene:scene openURLContexts:contexts];
+    }
+}
+
+- (void)sceneDidDisconnect:(UIScene *)scene
+{
+    // If the scene disconnects before the page loaded, tear down the pending
+    // launch-URL observer so it cannot fire for a defunct scene.
+    if (self.launchURLContexts != nil) {
+        [[NSNotificationCenter defaultCenter] removeObserver:self name:CDVPageDidLoadNotification object:nil];
+        self.launchScene = nil;
+        self.launchURLContexts = nil;
+    }
 }
 
 - (void)scene:(UIScene *)scene openURLContexts:(NSSet<UIOpenURLContext *> *)URLContexts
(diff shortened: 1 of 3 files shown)


GitToTheHub:
Is this AI generated? If so, you have to add a `Generated-By: [Modelname]` to your commit message, which is a requirement by the ASF (Apache Software Foundation)

GitToTheHub (approved):

PR #1686, opened 2026-08-06
fix(wkwebview): retain UIDelegate to restore alert/confirm/prompt

### LINKED ISSUE #1684: Regression: alert() no longer works in 8.1.1
# Bug Report

## Problem
Using `alert()` from JavaScript inside of a Cordova app using Cordova iOS 8.1.1 does nothing.

An alert message should appear.

Nothing appears.

## Information
It appears that cordova-ios 8.1.1 inadvertently broke alert handling, by changing some code so that the object responsible for displaying JavaScript alerts can be removed. I admittedly used AI to diagnose this, and it reports this:

> Cordova iOS 8.1.1 creates a `CDVWebViewUIDelegate` as a local variable and assigns it to `WKWebView.UIDelegate`. However, `WKWebView.UIDelegate` is a weak property, so it does not retain the delegate. Cordova no longer stores the delegate in its own strong `uiDelegate` property. Consequently, ARC deallocates the delegate after initialization, and JavaScript functions such as `alert()`, `confirm()`, and `prompt()` stop working.
> 
> This appears to have been introduced by commit [5669c2c9](https://github.com/apache/cordova-ios/commit/5669c2c9cb209f7ba6ecf70e1e2ea86b369fd459), which removed the assignment to `self.uiDelegate`.

### Command or Code
```
document.addEventListener("deviceready", function () {
    alert("Plain JavaScript alert");
}, false);
```

- iPhone 15 running iOS 26.6
- App created with Cordova 13.0, Cordova-iOS 8.1.1

### Version information
Cordova: Cordova CLI 13.0.0, Cordova iOS platform 8.1.1
No other frameworks
App built on macOS 15.7.7, Xcode 26.3

## Checklist

GitToTheHub:
### Motivation and Context

- Fixes a regression introduced by #1664 where the property `uiDelegate` was removed from `CDVWebViewEngine`.
- `WKWebView.UIDelegate` is weak, so the default `CDVWebViewUIDelegate` could be deallocated after assignment, which breaks JavaScript alert/confirm/prompt handling.
- Retain the delegate in `CDVWebViewEngine` by a strong property `uiDelegate` and clear it during dispose.
- The property is not only for retaining the UIDelegate, it's also part of the exposed CDVWebViewEngine API as readonly, so callers can inspect what delegate is currently installed. It can also be set through `updateWithInfo:` by a plugin. This is documented in [guides/Setting Delegates, Preferences and Script Message Handlers in the WebView.md](https://github.com/apache/cordova-ios/blob/master/guides/Setting%20Delegates%2C%20Preferences%20and%20Script%20Message%20Handlers%20in%20the%20WebView.md)
- Add a regression test that verifies the web view engine keeps a non-nil uiDelegate.
- Fixes #1684

Generated-By: GPT-5.3-Codex

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.h
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.h
@@ -22,9 +22,10 @@
 
 @interface CDVWebViewEngine : CDVPlugin <CDVWebViewEngineProtocol, WKScriptMessageHandler, WKNavigationDelegate>
 
-// The canonical WKUIDelegate used by this engine.
-// Exposed as readonly for inspection, while the engine retains it internally
-// because WKWebView.UIDelegate is weak.
+// WKUIDelegate reference for this engine, which can also be set via `updateWithInfo:`
+// by a plugin. Additionally it is a helper property to retain the UIDelegate, since
+// WKWebView.UIDelegate is weak and would otherwise be deallocated after assignment
+// which would cause JavaScript alert/confirm/prompt handling to stop working.
 @property (nonatomic, strong, readonly) id <WKUIDelegate> uiDelegate;
 
 - (void)allowsBackForwardNavigationGestures:(CDVInvokedUrlCommand*)command;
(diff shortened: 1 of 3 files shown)


dpogue (approved):

dwengs:
@dpogue I am still having this issue on

- Cordova 13.0.0
- Cordova iOS 8.1.1
- iOS 27 (iPhone 16)
- Xcode v27.0 (27A266a)

Do I have to update something or wait for a new cordova-ios release?

GitToTheHub:
This PR is not released yet and can only be found on the master.
You could add the development version of cordova-ios by:

```bash
cordova platform add https://github.com/apache/cordova-ios
```

PR #1689, opened 2026-08-10
feat(statusbar): change dark/light appearance according to background color

GitToTheHub:
### Platforms affected

iOS

### Motivation and Context

Since iOS 18.5 the system handles automatically the status bar appearance according to the underlying color, but before we would have to handle it by ourselves.

### Description

- Adds a luminance check to the set background color and sets the status bar appearance accordingly
- This is only done when older than 18.5. Since iOS 18.5 the system automatically handles the appearance.
- Added test `testStatusBarAppearanceFollowsBackgroundColor`

Generated-By: GPT-5.6 Terra

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -209,6 +209,7 @@ static UIColor *defaultBackgroundColor(void) {
     }
 
     [self.statusBarBackground setBackgroundColor:self.statusBarBackgroundColor];
+    [self setNeedsStatusBarAppearanceUpdate];
 }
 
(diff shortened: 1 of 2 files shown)


dpogue (approved):
This seems okay to me. I feel like earlier iOS versions would sometimes update the text colour but sometimes not, and it was wildly inconsistent and unreproducible. Maybe just calling `setNeedsStatusBarAppearanceUpdated` is enough to fix it? But in any case, we're essentially doing the same thing as the system here so it should be fine.

@GitToTheHub Could you rebase this on top of the test fixes so we can try to get a green CI run before merging?

GitToTheHub:
I rebased it

dpogue (approved):

PR #1694, opened 2026-08-24
fix(tests): Fix test warnings about nil UISceneSession

dpogue:
### Platforms affected
iOS

### Motivation and Context

The `+alloc` method on UISceneSession isn't publicly exposed, so we need to do it indirectly by calling it on the class instance. We don't actually do anything with the session, we just need it not to be nil so it doesn't cause warnings.

### Description

Create a UISceneSession and pass it where required in test cases.

### Testing

Tests run with no warnings.

### Checklist

### DIFF
--- a/tests/CordovaLibTests/CDVSceneDelegateTests.m
+++ b/tests/CordovaLibTests/CDVSceneDelegateTests.m
@@ -63,6 +63,7 @@
 @interface CDVSceneDelegateTests : XCTestCase
 @property (nonatomic, strong) CDVSceneDelegate *sceneDelegate;
 @property (nonatomic, strong) UIScene *placeholderScene;
+@property (nonatomic, strong) UISceneSession *placeholderSession;
 @end
 
 @implementation CDVSceneDelegateTests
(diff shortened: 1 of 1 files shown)


tests/CordovaLibTests/CDVSceneDelegateTests.m
```diff
@@ -97,13 +97,14 @@ - (CDVFakeSceneConnectionOptions *)connectionOptionsForURL:(NSURL *)url
...
+    UISceneSession *session = [[[UISceneSession class] alloc] init];
 
     XCTNSNotificationExpectation *notFiredYet = [[XCTNSNotificationExpectation alloc]
         initWithName:CDVPluginHandleOpenURLNotification];
     notFiredYet.inverted = YES;
 
     [self.sceneDelegate scene:self.placeholderScene
-          willConnectToSession:nil
+          willConnectToSession:session
                        options:[self connectionOptionsForURL:launchURL]];
```
GitToTheHub, line 114: Since this is repeating 4 times this could be sort out in a method like `setupSceneDelgateWithOptions:`? Also `[[[UISceneSession class] alloc] init]` could be set directly to `willConnectToSession:` since the variable `session` is not used elsewhere?

GitToTheHub: Also a documentation could be added for `[[[UISceneSession class] alloc] init]` what you wrote in the PR description:

```
The +alloc method on UISceneSession isn't publicly exposed, so we need to do it indirectly by calling it on the class instance. We don't actually do anything with the session, we just need it not to be nil so it doesn't cause warnings.
```

If the variable should be kept, it could be named `placeholderSession` like there already exists a `self.placeholderScene`.

GitToTheHub (approved):

Issue #1692, opened 2026-08-21
cordova-ios 8.1.1 status bar built in feature not working properly


xavierserrai:
# Bug Report

### What is expected to happen?

The status bar should be able to be configured from config.xml and the color should be able to be changed from the new js api using:

``````
window.statusbar.setBackgroundColor('#HEX_COLOR');
``````

### What does actually happen?

Configuration in config.xml:

```````
<preference name="StatusBarOverLaysWebView" value="false" />
<preference name="StatusBarBackgroundColor" value="#FF0000" />
```````

Tested in several devices with the Apache Cordova Hello World App:

- iPhone SE - iOS 26.1

When the app starts, it shows the statusbar with a red background, but immediatelly disappears and is set to transparent mode.

When calling:
``````
window.statusbar.setBackgroundColor('#HEX_COLOR');
``````
Anything happens.

- iPhone 12 Pro - iOS 27.0 public beta 5

The app starts and the status bar is allways in transparent mode.

When calling:
``````
window.statusbar.setBackgroundColor('#HEX_COLOR');
``````
Anything happens.

## Information

I have not been able to find documentation to use that new embedded statusbar feature.

### Command or Code

Just build and install

### Environment, Platform, Device

- iPhone SE - iOS 26.1
- iPhone 12 Pro - iOS 27.0 public beta 5

### Version information

cordova cli 13.0.0
cordova-ios 8.1.1
Version 27.0 beta 5
MacOS Tahoe 26.5 (25F71)

## Checklist

- [ X] I searched for existing GitHub issues
- [ X] I updated all Cordova tooling to most recent version
- [ X] I included all the necessary information above

**Comment by GitToTheHub** [comment]
Hi you have to change 

```html
<meta name="viewport" content="... viewport-fit=cover">
```

to 

```html
<meta name="viewport" content="... viewport-fit=contain">
```

to make it work. This is described in the blog article for cordova-ios 8: https://cordova.apache.org/announcements/2025/11/23/cordova-ios-8.0.0.html

Setting `<preference name="StatusBarOverLaysWebView" value="false" />` has no effect on cordova-ios 8.x and is handled by `viewport-fit` or the status bar JS API `window.statusbar.visible`.

PR #1695, opened 2026-08-24
fix(statusbar): keep built-in status bar view visibility in sync with the JS API

### LINKED ISSUE #1693: cordova-plugin-statusbar not working with cordova-ios 8.1.1 and iOS 27
# Bug Report

Status bar should work as always, should be configurable from config.xml and from the StatusBar js API.

Nothing happens, the configuration in config.xml has no effect and the StatusBar js API doesn't change anything from statusbar.

## Information

Made a fresh apache cordova-ios 8.1.1 app with the Apache Cordova Hello World app and installed the cordova-plugin-statusbar v.4.0.0 and tested in the following setups:

Configuration in config.xml:

```````
<preference name="StatusBarOverLaysWebView" value="false" />
<preference name="StatusBarBackgroundColor" value="#FF0000" />
```````

- iPhone SE - iOS 26.1

Working as usual, status bar appears in red and is accessible from js api.

- iPhone 12 Pro - iOS 27.0 public beta 5

Not working at all, no background color from config.xml and unable to manipulate it from StatusBar js API. Seems that plugin is somehow overriden by the built-in cordova-ios 8.1.1 statusbar, that doesn't also work as expected.

### Command or Code

Build and run.

iPhone SE - iOS 26.1
iPhone 12 Pro - iOS 27.0 public beta 5

### Version information

cordova cli 13.0.0
cordova-ios 8.1.1
cordova-plugin-statusbar 4.0.0 "StatusBar"
Version 27.0 beta 5
MacOS Tahoe 26.5 (25F71)

As this seems to be a problem shared with cordova-plugin-statusbar, I will replicate the issue there.

## Checklist

- [ X] I searched for existing GitHub issues
- [ X] I updated all Cordova tooling to most recent version
- [ X] I included all the necessary information above

### PR DESCRIPTION
**Pull request by MatrixNeoKozak**
## What
The built-in status bar view (which paints StatusBarBackgroundColor from config.xml) was created with `hidden = YES`, and its `hidden` flag was only ever updated by the `scrollViewDidChangeAdjustedContentInset:` delegate callback. `showStatusBar:` (the `window.statusbar.visible` / `StatusBarInternal.setVisible` path) only toggled the view's `alpha`, so when the scroll-view delegate callback does not fire the status bar view stays hidden regardless of the JS API or config.xml, making the status bar appear unresponsive (apache/cordova-ios#1693, broken on iOS 27). This change creates the status bar view visible and makes `showStatusBar:` drive `hidden` together with `alpha`, keeping the two visibility controls coherent, and adds a regression test asserting the view's `hidden`/`alpha` and `prefersStatusBarHidden` state after both show and hide calls.

## Why
This change resolves the target issue or improvement.

## How to test
Verify that the project builds/runs correctly and the specific bug/improvement is addressed.

Fixes #1693

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -746,8 +746,6 @@ static UIColor *defaultBackgroundColor(void) {
     [self.statusBar.trailingAnchor constraintEqualToAnchor:self.view.trailingAnchor].active = YES;
     [self.statusBar.topAnchor constraintEqualToAnchor:self.view.topAnchor].active = YES;
     [self.statusBar.bottomAnchor constraintEqualToAnchor:self.view.safeAreaLayoutGuide.topAnchor].active = YES;
-
-    self.statusBar.hidden = YES;
 #endif
 }
 
@@ -890,6 +888,7 @@ static UIColor *defaultBackgroundColor(void) {
 - (void)showStatusBar:(BOOL)visible
 {
 #if !defined(TARGET_OS_VISION) || !TARGET_OS_VISION
+    self.statusBar.hidden = !visible;
     [self.statusBar setAlpha:(visible ? 1 : 0)];
     [self setNeedsStatusBarAppearanceUpdate];
 #endif
--- a/tests/CordovaLibTests/CDVStatusBarTests.m
+++ b/tests/CordovaLibTests/CDVStatusBarTests.m
@@ -21,6 +21,7 @@
 #import <XCTest/XCTest.h>
 #import <Cordova/Cordova.h>
 #import "CDVWebViewEngine.h"
+#import "CDVViewController+Private.h"
 #import "CordovaApp-Swift.h"
 #import "CDVTestHelpers.h"
 
@@ -160,6 +161,21 @@ andWaitForCordovaCommandQueueWithWebViewEngine:self.plugin
     XCTAssertEqual(rgba[1], 0.f);
     XCTAssertEqual(rgba[2], 0.f);
     XCTAssertEqual(rgba[3], 0.25f);
+}
+
+- (void) testStatusBarVisibility {
+    UIView *statusBar = [self.viewController valueForKey:@"statusBar"];
+    XCTAssertNotNil(statusBar);
+
+    [self.viewController showStatusBar:NO];
+    XCTAssertTrue(self.viewController.prefersStatusBarHidden);
+    XCTAssertTrue(statusBar.hidden);
+    XCTAssertEqual(statusBar.alpha, 0.f);
+
+    [self.viewController showStatusBar:YES];
+    XCTAssertFalse(self.viewController.prefersStatusBarHidden);
+    XCTAssertFalse(statusBar.hidden);
+    XCTAssertEqual(statusBar.alpha, 1.f);
 }
 
 @end

### CODE REVIEW

**Comment by xavierserrai** [comment]
Dear @MatrixNeoKozak ,

I have tested the cordova-ios 8.1.1 platform with your fix and js api works properly:

```````
window.statusbar.setBackgroundColor('#FF0000');
```````

Sets the status bar to red color.

However, the configuration in config.xml:

```````
<preference name="StatusBarOverLaysWebView" value="false" />
<preference name="StatusBarBackgroundColor" value="#FF0000" />
```````

Is not working properly. At app startup, for one moment it appears a red background statusbar, but it disappears suddently leaving a statusbar with a gradient background.

Also, the statusbar in both cases overlays the webview.

For the overlay statusbar issue it was solved in [cordova-plugin-statusbar](https://github.com/apache/cordova-plugin-statusbar) using this code:

```````
CGRect statusBarFrame = CGRectZero;
if (@available(iOS 13.0, *)) {
    UIWindowScene *windowScene = self.viewController.view.window.windowScene;
    statusBarFrame = windowScene.statusBarManager.statusBarFrame;
}
````````

Thanks,

**Review by GitToTheHub — APPROVED** [review]
This change looks good. I think it's a little bit wrong to hide/show the status bar in `scrollViewDidChangeAdjustedContentInset:` depending on `scrollView.contentInsetAdjustmentBehavior == UIScrollViewContentInsetAdjustmentNever`. I also found out, that this is only the case on Mac Catalyst if `HideDesktopTitlebar` is `false`. Maybe this could be made better.

**Comment by GitToTheHub** [comment]
I documented the Mac Catalyst behaviour for the status bar on PR #1696

**Comment by MatrixNeoKozak** [comment]
Thanks for testing! The config.xml background color flash and overlay behavior sound like a separate issue from what #1693 covers (the JS API visibility sync) — want me to open a follow-up issue for those, referencing the statusBarManager.statusBarFrame approach from cordova-plugin-statusbar you mentioned?

**Comment by xavierserrai** [comment]
Dear @MatrixNeoKozak ,

As commented here:

https://github.com/apache/cordova-ios/issues/1692#issuecomment-5398204841

Those old tags from config.xml seem to be outdated.

Thanks,

File: CordovaLib/Classes/Public/CDVViewController.m [hunk 1]
```diff
@@ -746,8 +746,6 @@ - (void)createStatusBarView
     [self.statusBar.trailingAnchor constraintEqualToAnchor:self.view.trailingAnchor].active = YES;
     [self.statusBar.topAnchor constraintEqualToAnchor:self.view.topAnchor].active = YES;
     [self.statusBar.bottomAnchor constraintEqualToAnchor:self.view.safeAreaLayoutGuide.topAnchor].active = YES;
-
-    self.statusBar.hidden = YES;
```
GitToTheHub, line 750: I see you removed the line which initially hides the status bar. Did you test, if this has no regressions? For e.g. the status bar should be not visible when `viewport-fit=cover` is set.

GitToTheHub:
@xavierserrai 

> As commented here:
>
> https://github.com/apache/cordova-ios/issues/1692#issuecomment-5398204841
> 
> Those old tags from config.xml seem to be outdated.

The preference `StatusBarBackgroundColor` is still supported by cordova-ios 8.x.
It's only that `StatusBarOverLaysWebView` is not supported.

xavierserrai:
Dear @GitToTheHub,

Thanks for the clarification. However, I have not been able to find that in the documentation, only in changelogs.

Thanks,

GitToTheHub:
I know, the documentation is currently not the best, but I try to improve it.

dpogue:
> I think it's a little bit wrong to hide/show the status bar in `scrollViewDidChangeAdjustedContentInset:` depending on `scrollView.contentInsetAdjustmentBehavior == UIScrollViewContentInsetAdjustmentNever`.

This is how we detect changes to `viewport-fit`:
`viewport-fit=cover` causes the `contentInsetAdjustmentBehavior` to be `UIScrollViewContentInsetAdjustmentNever`

<small>(because WebKit still doesn't expose `_webView:didChangeSafeAreaShouldAffectObscuredInsets:` as public API, so we need to react based on the scrollView delegate)</small>

dpogue:
> when the scroll-view delegate callback does not fire the status bar view stays hidden

Under what conditions does `scrollViewDidChangeAdjustedContentInset:` not fire?


WebKit internally sets the content inset on the scroll view: https://github.com/WebKit/WebKit/blob/4087ab818e4949f1d63a8c4c54fd08369b48f420/Source/WebKit/UIProcess/API/ios/WKWebViewIOS.mm#L3735

That method gets called unconditionally as part of setting up the scroll and content views: https://github.com/WebKit/WebKit/blob/4087ab818e4949f1d63a8c4c54fd08369b48f420/Source/WebKit/UIProcess/API/ios/WKWebViewIOS.mm#L244

It's also called again any time the `viewport-fit` meta directive is changed: 
https://github.com/WebKit/WebKit/blob/4087ab818e4949f1d63a8c4c54fd08369b48f420/Source/WebKit/UIProcess/API/ios/WKWebViewIOS.mm#L3757

GitToTheHub:
@dpogue Thanks for the explanation. The author wrote about Status Bar JS API `window.statusbar.visible`. Maybe this will not change any insets?

dpogue:
`window.statusbar.visible` makes the status bar content and background visible or invisible, it does not impact the safe area insets at all

GitToTheHub:
> window.statusbar.visible makes the status bar content and background visible or invisible, it does not impact the safe area insets at all

Correct but the status bar is set initially to `self.statusBar.hidden = YES`:

https://github.com/apache/cordova-ios/blob/f525ec9ec8b434f79928c66dff2bca06e178066d/CordovaLib/Classes/Public/CDVViewController.m#L738-L758

and when `showStatusBar:` is called, it will only change the alpha, not the hidden state:

https://github.com/apache/cordova-ios/blob/f525ec9ec8b434f79928c66dff2bca06e178066d/CordovaLib/Classes/Public/CDVViewController.m#L896-L902

So the hidden state only depends on when `scrollViewDidChangeAdjustedContentInset:` is called.

Anyway why is the alpha changed on the status bar? Would it be not enough to just hide it?

GitToTheHub:
> WebKit internally sets the content inset on the scroll view: https://github.com/WebKit/WebKit/blob/4087ab818e4949f1d63a8c4c54fd08369b48f420/Source/WebKit/UIProcess/API/ios/WKWebViewIOS.mm#L3735
>
> That method gets called unconditionally as part of setting up the scroll and content views: https://github.com/WebKit/WebKit/blob/4087ab818e4949f1d63a8c4c54fd08369b48f420/Source/WebKit/UIProcess/API/ios/WKWebViewIOS.mm#L244

So you mean when the WebView gets initialized it should always call scrollViewDidChangeAdjustedContentInset:?

dpogue:
> Anyway why is the alpha changed on the status bar? Would it be not enough to just hide it?

We detect if the alpha is 0 and use that as the flag to hide the status bar content: https://github.com/apache/cordova-ios/blob/f525ec9ec8b434f79928c66dff2bca06e178066d/CordovaLib/Classes/Public/CDVViewController.m#L597-L602

We can't rely on `visible` because `visible` is false when `viewport-fit=cover` but the status bar **content** should still be shown (just not the background)

> So you mean when the WebView gets initialized it should always call scrollViewDidChangeAdjustedContentInset:?

According to the WebKit source code, yes.

If someone can put together a test case where that's not happening, I'd love to investigate.

GitToTheHub:
Ok I will make some improvements to the code, this is very confusing. I added more documentation to `scrollViewDidChangeAdjustedContentInset:` on PR #1699 and will rename `statusBar` to `statusBarBackground` on `CDVViewController`, since this is just the background and not the real status bar.

GitToTheHub:
I think the code should be improved for `prefersStatusBarHidden` and `showStatusBar:`.

`statusBar.alpha` will only be changed if `showStatusBar:` will be called:

https://github.com/apache/cordova-ios/blob/f525ec9ec8b434f79928c66dff2bca06e178066d/CordovaLib/Classes/Public/CDVViewController.m#L896-L902

First improvement: `statusBar.hidden` should be set here accordingly.

Since calling `showStatusBar:` is the only place where `statusBar.alpha` could be set to `1` or `0`, which is only to make the real status bar appear or disappear by `prefersStatusBarHidden`, it would be better to add a property named `statusBarHidden` which is for hidding/showing the real status bar. I renamed `statusBar` to `statusBarBackground` in PR #1700 to make a better distinction between the real status bar and our own background. Using `statusBar.alpha` by hidding/showing the real status bar seems for me misleading as for e.g. the real status bar would be shown when `self.statusBar.alpha` is `1.000000` but `self.statusBar.hidden` is `YES`.

dpogue:
I was just thinking that a `statusBarContentHidden` bool would probably make more sense. I can make a PR to implement that.

GitToTheHub:
Since `statusBar` is called now `statusBarBackground` I would say that `statusBarHidden` is sufficient now. The method for hidding/showing the real status bar is called `prefersStatusBarHidden` and not `prefersStatusBarContentHidden`. Why you would prefer `statusBarContentHidden`?

dpogue:
yeah, `statusBarHidden` is probably fine

GitToTheHub:
Ok than you can make a PR for it :)

PR #1697, opened 2026-08-25
fix(codesigning): Use an ASCII EXECUTABLE_NAME

### LINKED ISSUE #1590: Upload to App Store failing after update to to version 8.0.0
# Bug Report

## Problem
App with special characters in name gives an error when submiting to the App Store after updating to cordova ios 8.

The same App with name without special charaters is able to be sbmited with no errors.

I should be able to publish with no errors

I'm able to archive the app correctly but when publishing to the App store I get error:
```
Validation failed (409)
Missing or invalid signature. The bundle '......' at bundle path 'Payload/<name with special characters>.app' is not signed using an Apple submission certificate. 
```

## Information
I control the App name by changing the `CFBundleDisplayName` in info.plist with:
```
<config-file target="*-Info.plist" parent="CFBundleDisplayName">
            <string>My App</string>
</config-file>
```
The `name` node in `config.xml` falls back to a safe name. In previous `cordova-ios` version this worked fine. In the new version the App get's the name which is set in `name` node of config.xml.

Due to this, I now have set the `name` to the name with special characters and stopped editing the `CFBundleDisplayName` but the upload fails with the error above. 

MacOS Latest

### Version information
cordova ios 8
XCode 26

## Checklist

dpogue:
### Platforms affected
iOS

### Motivation and Context

Hopefully closes GH-1590 based on the investigation work from @deguchi

### Description

This should allow the app bundle and the display name to have whatever Unicode characters they want, but the actual executable within the bundle will always be named App and hopefully avoid any issues Xcode has with finding and properly code-signing the executable.

### Testing

Built locally and confirmed that the app bundle had the name provided in config.xml and the executable file within the bundle was name "App".

I am not able to test making an App Store build.

### Checklist


GitToTheHub:
Is this ready to merge or do you want to wait for user feedback of @deguchi?

dpogue:
I was hoping someone would be able to test this before merging, to see if it actually fixes the problem or not

GitToTheHub:
Hi @seamlink-aalves, you created originally the issue #1590. Would it maybe be possible to test this PR if this fixes the issue you had when you created the issue? This would be a nice help.

seamlink-aalves:
I confirm that this PR solves the issue I originally reported.

GitToTheHub:
Great 😀

GitToTheHub (approved):

PR #1698, opened 2026-08-26
feat(geolocation): Implement WKUIDelegate geolocation permission method

dpogue:
I expect this to be potentially controversial given the state of APIs here, but I believe it is safe and will remove the need for the geolocation plugin for apps targeting iOS 15 and newer.

It would be nice if there were a way to test that this passes Apple's automatic API usage scan before merging 😕 

### Platforms affected
iOS (15+)

### Motivation and Context

This method allows controlling whether access to the geolocation API is available based on the origin requesting access. This resolves the primary reason that the Cordova Geolocation plugin exists.

### Description

However, the API is only available as public API in iOS 27 and newer, but has been available as private API with the same signature since iOS 15. Since we're allowed to declare the public API now, we can do that and then dynamically alias it to the private API for older iOS versions.

A new `GeolocationPermissionGrantType` preference is supported in config.xml, following the same rules as the media permissions. If not specified, the app scheme will have access to geolocation APIs, but no other origins.

Any geolocation access requires the `NSLocationWhenInUseUsageDescription` option to be set in Info.plist, otherwise the app will crash when initializing Location Services.

### Testing

Confirmed that this allows geolocation to work with just a single prompt in a Cordova app, without the need for the geolocation plugin.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
@@ -278,6 +278,7 @@
 
         uiDelegate.title = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleDisplayName"];
         uiDelegate.mediaPermissionGrantType = [self parsePermissionGrantType:[settings cordovaSettingForKey:@"MediaPermissionGrantType"]];
+        uiDelegate.geolocationPermissionGrantType = [self parsePermissionGrantType:[settings cordovaSettingForKey:@"GeolocationPermissionGrantType"]];
         uiDelegate.allowNewWindows = [settings cordovaBoolSettingForKey:@"AllowNewWindows" defaultValue:NO];
 
         // Assigning to self.uiDelegate keeps a strong reference to the delegate,
--- a/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewUIDelegate.m
+++ b/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewUIDelegate.m
@@ -20,6 +20,10 @@
 #import "CDVWebViewUIDelegate.h"
 #import <Cordova/CDVViewController.h>
 
+#if __IPHONE_OS_VERSION_MIN_REQUIRED < 270000
+#import <objc/runtime.h>
+#endif
+
 @interface CDVWebViewUIDelegate ()
 
 @property (nonatomic, weak) CDVViewController *viewController;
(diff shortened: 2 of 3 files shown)


GitToTheHub:
Is this read to review?

dpogue:
This is ready for review, but should not be merged until after iOS 27 is released

CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewUIDelegate.m
```diff
@@ -31,6 +35,42 @@ @implementation CDVWebViewUIDelegate
...
+    } else if (@available(iOS 15.0, *)) {
+        /* Alias the public API delegate method to the private API */
+        Class class = [self class];
+
+        SEL publicSelector = @selector(webView:requestGeolocationPermissionForOrigin:initiatedByFrame:decisionHandler:);
+        SEL privateSelector = NSSelectorFromString([NSString stringWithFormat:@"_%@", NSStringFromSelector(publicSelector)]);
+
+        Method publicMethod = class_getInstanceMethod(class, publicSelector);
+
+        class_addMethod(class, privateSelector, method_getImplementation(publicMethod), method_getTypeEncoding(publicMethod));
```
GitToTheHub: AI noted here:

> High priority: remove the private API fallback. [CDVWebViewUIDelegate.m, lines 60–69](https://github.com/apache/cordova-ios/blob/2fcedd29798d3fae11e326d22eb69c849c4a6363/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewUIDelegate.m#L60) dynamically registers the private selector on iOS 15–26. Constructing its name at runtime does not make it public API, and publishing an equivalent in iOS 27 does not retroactively authorize the older API. This exposes consuming apps to rejection under [Apple’s guideline 2.5.1](https://developer.apple.com/app-store/review/guidelines/#software-requirements). Passing an automated scan would not resolve that concern. Keep the public iOS 27 implementation and retain the plugin approach for older versions.

CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m
```diff
@@ -278,6 +278,7 @@ - (void)pluginInitialize
 
         uiDelegate.title = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleDisplayName"];
         uiDelegate.mediaPermissionGrantType = [self parsePermissionGrantType:[settings cordovaSettingForKey:@"MediaPermissionGrantType"]];
+        uiDelegate.geolocationPermissionGrantType = [self parsePermissionGrantType:[settings cordovaSettingForKey:@"GeolocationPermissionGrantType"]];
```
GitToTheHub: AI noted here:

> Medium priority: reconcile the default permission behavior with the stated intent.
> 
> The new preference assignment uses a parser that defaults to `grantIfSameHostElsePrompt`. Consequently, other hosts can obtain permission through a prompt, whereas the PR description says only the app scheme gets access by default. The comparison also checks only the hostname, ignoring scheme and port. Either implement the intended restriction or explicitly document the broader policy.

GitToTheHub (commented):
I reviewed this with GPT-6:

> The reported CI checks passed, but this PR adds no automated tests or documentation. Before merging, I’d want permission-policy tests and documentation for GeolocationPermissionGrantType and the required NSLocationWhenInUseUsageDescription.

dpogue:
I'm really not interested in what AI thinks of the code, when I've already explained in detail why we're falling back to implementing a private API and why that should be safe in this case, and the implementation matches the behaviour of the camera/microphone media access policy preference. If someone has actual objections to that, please say what those objections are instead of just saying something that's already known information.

If people can't be bothered to use their brains to look at code and read what's been written, then I'm not particularly interested in contributing code.

GitToTheHub:
Hey @dpogue,
using AI for reviewing was just a helper to make a review and push things forward. It was not meant to bother you. AI is a tool like we use search engines which for e.g. points us to StackOverflow. I just wanted to make a fast review.

I was just concerned about, that it mentioned something about

> This exposes consuming apps to rejection under [Apple’s guideline 2.5.1](https://developer.apple.com/app-store/review/guidelines/#software-requirements)

when it was about the private API fallback. When this is wrong than everything is good to go? But anyway, should this feature be documented somewhere in cordova-docs? Which place would be best for it?

breautek:
Anything that breaches the apple guidelines I think is a valid concern. Apple is really stingy in their app reviews, and even if this was used today and passed app reviews, it doesn't mean they won't randomly reject an app later on citing Section 2.5.1.

I've had rejections From Apple because of particular components that the App had that had existed in the app for years before they decided to take issue with it. Point here being that Apple reviews are not always consistent and is subjected to the reviewer's discretion. We can get reports from 100 different people stating they successfully submitted an app with this update and the 101th submission apple could reject due to this change. So I'd be worried that using a private API could cause problems later.

I understand that the API itself became public in the iOS 27 SDK, so this is kind of a gray area -- but technically speaking the API isn't public on iOS 15-26, so it should be considered unsafe to use directly on those versions and we are hacking our way around that so that we can use it directly.

All of this is kind of theoretical, but should this become a real issue -- reverting this kind of update after people rely on it might cause an even bigger headache. And once it's flagged for one app submission, the only option that user would have to get their app approved is to downgrade to a cordova-ios framework to a version that doesn't include this change, which might not be practically possible.

Taking 2.5.1 as literal as possible, you'd have to assume it won't consistently pass app review, because it states that Apps may only use public APIs, and the consequences of a rejection would have severe implications; so unfortunately, I think this is something that we shouldn't accept as is. We either need a graceful fallback for devices older than iOS 27, a fallback that doesn't involve accessing a private API on those device versions, or we must wait until we can set our deployment target to iOS 27, which obviously would not be anytime soon.

breautek:
> All of this is kind of theoretical

There is older precedent based on: https://ajnaware.wordpress.com/2009/09/24/yet-another-dubious-app-rejection-story/

And while the blog author denies actually using any private APIs, their posted Apple response states that their app was using APIs marked as private on iOS 3.0 and those APIs were only available as public APIs on iOS 3.1.

The author went on to say that Apple basically requested the author to change the min deployment to 3.1+ if they wanted to use APIs introduced publicly in 3.1.

> However, the bad news was that Apple was requesting me to resubmit the app for deployment only on OS3.1.

PR #1699, opened 2026-08-26
doc(CDVViewController): document `scrollViewDidChangeAdjustedContentInset:`

GitToTheHub:
### Motivation and Context

- Make clear that `scrollViewDidChangeAdjustedContentInset:` is called when the `WKWebView` is initialized and also when `viewport-fit` is changed.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -579,9 +579,8 @@ static UIColor *defaultBackgroundColor(void) {
 }
 
 /**
- * This is called initially when the WKWebView is initialized and also when viewport-fit is changed.
- * WebKit still doesn't expose _webView:didChangeSafeAreaShouldAffectObscuredInsets: as public API,
- * so we need to react based on the scrollView delegate.
+ * We use this delegate method on the web view's scrollView to detect when the safe area insets are changed
+ * by the viewport-fit value.
  */
 - (void)scrollViewDidChangeAdjustedContentInset:(UIScrollView *)scrollView
 {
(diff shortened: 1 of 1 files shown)


CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -578,6 +578,11 @@ - (void)onWebViewPageDidLoad:(NSNotification*)notification
     }
 }
 
+/**
+ * This is called initially when the WKWebView is initialized and also when viewport-fit is changed.
+ * WebKit still doesn't expose _webView:didChangeSafeAreaShouldAffectObscuredInsets: as public API,
+ * so we need to react based on the scrollView delegate.
+ */
 - (void)scrollViewDidChangeAdjustedContentInset:(UIScrollView *)scrollView
```
dpogue, line 585: This is a standard UIScrollViewDelegate method: https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/scrollviewdidchangeadjustedcontentinset(_:)?language=objc

So it's probably best not to make reference to WebKit internals here. Something like "We use this delegate method on the web view's scrollView to detect when the safe area insets are changed by the viewport-fit value" would probably be better.

The problem with putting lots of specific details in comments is that they can quickly get outdated and then nobody knows if they're still relevant of not. I've found that sort of thing in several places in Cordova with comments dating back to Cordova 3, and I still can't make sense of some of them.

GitToTheHub: I used your suggestion.

dpogue (approved):

PR #1701, opened 2026-08-26
refactor(statusbar): Refactor the status bar hiding code

dpogue:
### Platforms affected
iOS

### Motivation and Context

The use of background colour alpha to determine whether status bar content was visible was confusing, so make it its own boolean.

### Description

Added an internal `_statusBarHidden` boolean to CDVViewController to control whether the status bar content is visible or not.

### Testing

Added test cases to confirm the behaviour.

### Checklist

### DIFF
--- a/CordovaLib/Classes/Public/CDVViewController.m
+++ b/CordovaLib/Classes/Public/CDVViewController.m
@@ -45,6 +45,7 @@ static UIColor *defaultBackgroundColor(void) {
     UIColor *_statusBarBackgroundColor;
     UIColor *_statusBarWebViewColor;
     UIColor *_statusBarDefaultColor;
+    BOOL _statusBarHidden;
     CDVSettingsDictionary* _settings;
 }
 
@@ -598,7 +599,11 @@ static UIColor *defaultBackgroundColor(void) {
     // On Mac Catalyst, Cordova may manually constrain the web view below the title bar.
     // In that case automatic scroll-view inset adjustment is disabled, so the synthetic
     // status bar view should be hidden.
-    self.statusBarBackground.hidden = (scrollView.contentInsetAdjustmentBehavior == UIScrollViewContentInsetAdjustmentNever);
+    if (_statusBarHidden) {
+        self.statusBarBackground.hidden = true;
+    } else {
+        self.statusBarBackground.hidden = (scrollView.contentInsetAdjustmentBehavior == UIScrollViewContentInsetAdjustmentNever);
+    }
 #endif
 }
 
@@ -607,7 +612,7 @@ static UIColor *defaultBackgroundColor(void) {
 {
     // The CDVStatusBar plugin overrides this in a category extension, and
     // should bypass this implementation entirely
-    return self.statusBarBackground.alpha < 0.0001f;
+    return _statusBarHidden;
 }
 #endif
 
@@ -905,7 +910,11 @@ static UIColor *defaultBackgroundColor(void) {
 - (void)showStatusBar:(BOOL)visible
 {
 #if !defined(TARGET_OS_VISION) || !TARGET_OS_VISION
-    [self.statusBarBackground setAlpha:(visible ? 1 : 0)];
+    _statusBarHidden = !visible;
+
+    UIScrollView *scrollView = [self.webView performSelector:@selector(scrollView)];
+    [self scrollViewDidChangeAdjustedContentInset:scrollView];
+
     [self setNeedsStatusBarAppearanceUpdate];
 #endif
 }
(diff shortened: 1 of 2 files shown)


CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -45,6 +45,7 @@ @interface CDVViewController () <CDVWebViewEngineConfigurationDelegate, UIScroll
     UIColor *_statusBarBackgroundColor;
     UIColor *_statusBarWebViewColor;
     UIColor *_statusBarDefaultColor;
+    BOOL _statusBarHidden;
```
GitToTheHub, line 48: What about renaming the variable to `_systemStatusBarHidden`? Than it's clear it's not about our own custom status bar background

CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -598,7 +599,11 @@ - (void)scrollViewDidChangeAdjustedContentInset:(UIScrollView *)scrollView
     // On Mac Catalyst, Cordova may manually constrain the web view below the title bar.
     // In that case automatic scroll-view inset adjustment is disabled, so the synthetic
     // status bar view should be hidden.
-    self.statusBarBackground.hidden = (scrollView.contentInsetAdjustmentBehavior == UIScrollViewContentInsetAdjustmentNever);
+    if (_statusBarHidden) {
+        self.statusBarBackground.hidden = true;
```
GitToTheHub, line 603: Rather than updating `self.statusBarBackground.hidden` to `true` when `_statusBarHidden` changes, I would sync it directly in `showStatusBar:`. It's more clean if it's just:

```objc
if (!_statusBarHidden) {
    self.statusBarBackground.hidden = (scrollView.contentInsetAdjustmentBehavior == UIScrollViewContentInsetAdjustmentNever);
}
```

with the comment above. The comment above also is only relevant for this piece of code.

CordovaLib/Classes/Public/CDVViewController.m
```diff
@@ -905,7 +910,11 @@ - (void)showSplashScreen:(BOOL)visible
 - (void)showStatusBar:(BOOL)visible
 {
 #if !defined(TARGET_OS_VISION) || !TARGET_OS_VISION
-    [self.statusBarBackground setAlpha:(visible ? 1 : 0)];
+    _statusBarHidden = !visible;
+
+    UIScrollView *scrollView = [self.webView performSelector:@selector(scrollView)];
+    [self scrollViewDidChangeAdjustedContentInset:scrollView];
+
```
GitToTheHub, line 917: I would sync here ` self.statusBarBackground.hidden` directly. After that `scrollViewDidChangeAdjustedContentInset` could additionally be called.

dpogue: The correct value of `hidden` here depends on the scrollView state, so calling `scrollViewDidChangeAdjustedContentInset:` allows us to avoid duplicating that contentInsetAdjustmentBehavior  check in multiple places

GitToTheHub: I just want to remove this code in `scrollViewDidChangeAdjustedContentInset:`:

```objc
if (_statusBarHidden) {
        self.statusBarBackground.hidden = true;
```

because this will be only the case when `showStatusBar:` is called with `NO` and somebody might wonder why `self.statusBarBackground.hidden` is set here to `true`, when `_statusBarHidden` is `YES`.

In `showStatusBar:` everything would stay what you added. Just `self.statusBarBackground.hidden= _statusBarHidden` would be added for clarity.

GitToTheHub: I know `self.statusBarBackground.hidden` would be assigned twice when `showStatusBar:YES` is called, but I think this is not a big problem.

GitToTheHub: Another thing could be to source out the code which sets `self.statusBarBackground.hidden` in `scrollViewDidChangeAdjustedContentInset:`, like:

```objc
- (BOOL)shouldStatusBarBeHidden
{
    // Detect changes on viewport-fit:
    // viewport-fit=cover causes the contentInsetAdjustmentBehavior to be UIScrollViewContentInsetAdjustmentNever
    //
    // On Mac Catalyst, Cordova may manually constrain the web view below the title bar.
    // In that case automatic scroll-view inset adjustment is disabled, so the synthetic
    // status bar view should be hidden.
    UIScrollView *scrollView = [self.webView performSelector:@selector(scrollView)];
    return scrollView.contentInsetAdjustmentBehavior == UIScrollViewContentInsetAdjustmentNever
}
```

This could be called than in `showStatusBar:` and `scrollViewDidChangeAdjustedContentInset:`

dpogue: I feel like pulling all the logic out into a separate `shouldStatusBarBeHidden` would make it less obvious why we need the `scrollViewDidChangeAdjustedContentInset:`. That is where the actual logic for showing or hiding the status bar background makes sense to live.

GitToTheHub (approved):

PR #1717, opened 2026-09-08
fix(CDVSceneDelegate): deliver a universal link that arrives at launch

MaximBelov:
### Platforms affected

iOS

### Motivation and Context

A universal link that launches the app is dropped: the app opens on its start page and the link is
never handled.

`-scene:continueUserActivity:` is only called for an app that is already running. On a cold launch
the activity arrives in `connectionOptions.userActivities`, and `-scene:willConnectToSession:options:`
does not look at it — `userActivities` appears nowhere in CordovaLib. So no plugin ever hears about
it, and there is nothing an app or plugin can do about that from outside CordovaLib.

This is the user-activity half of #1671, which was fixed for `URLContexts` in the same method.
Related: #1654 covered the already-running case, which works.

### Description

Buffers the launch user activity exactly as the launch URL contexts are buffered, and replays it on
`CDVPageDidLoadNotification` so the observer registered in a plugin's `-pluginInitialize` exists by
then. Posting at scene-connect time is not enough — plugins are created with the view controller,
after the scene connects.

Cleared in `-sceneDidDisconnect:` alongside the URL contexts, same as before.

### Testing

Tested on a device (iPhone 16, iOS 26.6) with an app using Branch for universal links, tapping the
link in Notes so it is delivered as a universal link rather than as a URL:

- app not running: without this change the app opened on its start page; with it, it opens the
  linked screen
- app in the background, and app suspended after a dwell: unaffected either way, both already worked
  through `-scene:continueUserActivity:`

Unit tests added in `tests/CordovaLibTests/CDVSceneDelegateTests.m`, following the existing
URL-launch cases: buffered at scene-connect time, replayed after `CDVPageDidLoadNotification`,
nothing posted without an activity, teardown on `-sceneDidDisconnect:`, the warm path still posting
immediately, and a URL and an activity together. `npm run test:xcode` passes. Reverting
`CDVSceneDelegate.m` makes exactly the two tests that assert this change fail, so they do bite.

(An earlier version of this description said there was no test file for `CDVSceneDelegate`. That was
wrong — see @dpogue's comment below.)

### Checklist

- [ ] Commit is prefixed with `(platform)` if this change only applies to one platform — n/a, iOS-only repo
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above
- [ ] I've updated the documentation if necessary — n/a, no documented behaviour changes


dpogue:
Thanks for the PR!

There are some test cases for CDVSceneDelegate here that do cover the case for buffering the URL launch: https://github.com/apache/cordova-ios/blob/master/tests/CordovaLibTests/CDVSceneDelegateTests.m

It might be possible to add userActivity cases to that following the same sort of pattern.

MaximBelov:
You're right, and my apologies — I said there was no test file for `CDVSceneDelegate`, which was
simply wrong. I went looking, truncated the listing, and reported the truncation as a fact.

Added six cases mirroring the URL ones: buffered at scene-connect time, replayed after
`CDVPageDidLoadNotification`, nothing posted when there is no activity, teardown on
`-sceneDidDisconnect:`, the warm `-scene:continueUserActivity:` path still posting immediately, and
one case with a URL and an activity together — the URL tests had no "both" case, but the
implementation has that path.

`npm run test:xcode` passes (193s, no failures). I also checked the new tests actually bite: with
`CDVSceneDelegate.m` reverted to its pre-change state, exactly the two that assert the fix fail and
the other nine pass.

Fixing the PR description too, since it still carries that claim about coverage.

dpogue (approved):
This looks good to me. I see it's still marked as a draft, is there more work to do here or is this ready to be reviewed and merged?

MaximBelov:
Nothing outstanding — it stayed a draft until I had verified it on a device against a real universal link rather than a simulator. Done: iPhone 16 / iOS 26.6, cold launch reaches the linked screen, warm and background unchanged. Marking it ready.

One note on `userActivities.anyObject`: the header calls it "a set of available user activities to handle on connection", so more than one is contemplated, but I could not construct a launch that produced two — handoff is explicitly not in this set (it arrives via `handoffUserActivityType` and the `willContinue`/`continue` pair), and the notification response, shortcut item and CloudKit share metadata are each their own property. Iterating and posting per activity is a couple of lines if you would rather have it; say which and I will push it.

erisu:
@MaximBelov 

I apologies if I'm wrong, but was any AI tooling used to help generate or assistant in part of the commits, in this PR?

If any tooling was used, would it be possible to update the commit messages to include the `Generated-By:` tag as per the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html).

For example, if "Claude Opus 5" was used, the commit message would contain: `Generated-by: Claude Opus 5`

MaximBelov:
Force-pushed a message-only rewrite: both commits now carry `Generated-by: Claude Opus 5`, per the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html). No code change, the tree is identical.

PR #1724, opened 2026-09-22
Replace plugin variables with their actual values

chemerisuk:
### Platforms affected
ios

### Motivation and Context
In cordova-ios v8 framework added ability to use Swift Package Manager to manage plugin dependencies. After trying to switch to this new capability I noticed that there is no way to use plugin variables specified from CLI or default preference values from `plugin.xml` in `Package.swift`. At the same time there is logic that replaces plugin variables in `plugin.xml`, and this is really helpful for plugin authors.

### Description
I was able to introduce similar logic for `Package.swift` using hooks (example https://github.com/chemerisuk/cordova-plugin-firebase-analytics/blob/master/hooks/ios/update_package_swift.js) but it requires adding extra file to every plugin. Better to add such capability into the platform.

### Testing
New spec is added to verify proposed changes.

### Checklist


dpogue (approved):
This looks good to me!

My one caution (although this is more for plugin authors): One of the advantages of using Package.swift is that plugin repos can be valid Swift packages themselves and include things like test cases. If your Package.swift file contains variables for version placeholders, it probably won't work as a standalone package.

GitToTheHub:
I added documentation about this in apache/cordova-docs#1529, which can be merged after this is merged.

GitToTheHub:
> My one caution (although this is more for plugin authors): One of the advantages of using Package.swift is that plugin repos can be valid Swift packages themselves and include things like test cases. If your Package.swift file contains variables for version placeholders, it probably won't work as a standalone package.

I noted this in the documentation of apache/cordova-docs#1529 also.

